Issue #2, opened 2019-01-02
Publish messages when bot started/stopped


ssudake21:
We should notify on communication channel for following scenarios

1. Bot starts initially
2. Bot goes under graceful shutdown, or bot is aware of its shutdown

Any kubernetes actions which result in starting or stopping of bot should be considered.

PR #3, opened 2019-01-02
Send bot start/stop messages to slack channel

PrasadG193:
Fixes: #2 
- Send bot start/stop messages to slack channel
- Fix timestamps in delete notifications

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -1,9 +1,9 @@
 package controller
 
 import (
-	//"fmt"
 	"os"
 	"os/signal"
+	"strconv"
 	"strings"
 	"syscall"
 	"time"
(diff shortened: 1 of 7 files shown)


pkg/execute/executor.go
```diff
@@ -132,11 +132,13 @@ func runNotifierCommand(args []string) string {
 		}
 		if args[1] == "start" {
 			config.Notify = true
-			return "Notifier started!"
+			log.Logger.Info("Notifier enabled")
+			return "Brace yourselves, notifications are coming."
```
ssudake21: Define constants for these strings at package level. We can easily change text later if needed.
Same for all text message being sent to slackc

pkg/controller/controller.go
```diff
@@ -168,5 +190,14 @@ func logEvent(obj interface{}, kind, eventType string, err error) {
 
 	// Send notification to communication chennel
 	notifier := notify.NewSlack()
-	notifier.Send(event)
+	notifier.SendEvent(event)
+}
+
+func sendMessage(msg string) {
```
ssudake21: Keep this function in notifier

PrasadG193: @ssudake21 we already have `notifier.SendMessage(msg)` function in notifier package

mszostok (approved):

PR #13, opened 2019-01-15
Support notifications from multiple clusters

PrasadG193:
Fixes https://github.com/infracloudio/botkube/issues/8 and https://github.com/infracloudio/botkube/issues/7


config.yaml
```diff
@@ -118,3 +118,10 @@ communications:
   slack:
     channel: 'SLACK_CHANNEL'
     token: 'SLACK_API_TOKEN'
+
+# Setting to support multiple clusters
+settings:
+  # Cluster name to differentiate incoming messages
+  clustername: dev-cluster
```
ssudake21: Should this be part of helm installation? In the way we accept slack token.

Default can be `not-configured` or `unknown`

PR #25, opened 2019-01-18
Serve BotKube commands without '@' mention in DMs 

PrasadG193:


pkg/slack/slack.go
```diff
@@ -56,12 +56,42 @@ func (b *Bot) Start() {
 			logging.Logger.Debug("Connection Info: ", ev.Info)
 
 		case *slack.MessageEvent:
-			// Serve only if mentioned
-			if !strings.HasPrefix(ev.Text, "<@"+botID+"> ") {
+			// Check if message posted in a channel
+			chanInfo, _ := api.GetChannelInfo(ev.Channel)
```
ssudake21: what is value of ev.Channel in case of DM?

pkg/slack/slack.go
```diff
@@ -56,12 +56,42 @@ func (b *Bot) Start() {
...
+			groupInfo, _ := api.GetGroupInfo(ev.Channel)
+
+			if chanInfo != nil || groupInfo != nil {
+				// Message posted in a channel/group
+				// Serve only if starts with mention
+				if !strings.HasPrefix(ev.Text, "<@"+botID+"> ") {
+					continue
+				}
+				logging.Logger.Debugf("Slack incoming message: %+v", ev)
+				msg := strings.TrimPrefix(ev.Text, "<@"+botID+"> ")
```
ssudake21: Isnt this code same as line 92 to 101. Let try to merge both. In case of channel without prefix we return or continue

pkg/slack/slack.go
```diff
@@ -56,12 +56,42 @@ func (b *Bot) Start() {
...
+					InMessage: msg,
+					RTM:       rtm,
+				}
+				sm.HandleMessage(b.AllowKubectl)
+				continue
+			}
+
+			// Message posted as a DM
+			// Skip if message posted by BotKube
+			if ev.User == botID {
```
ssudake21: move this to top of function

pkg/slack/slack.go
```diff
@@ -56,12 +56,42 @@ func (b *Bot) Start() {
...
 				continue
 			}
 			logging.Logger.Debugf("Slack incoming message: %+v", ev)
-			msg := strings.TrimPrefix(ev.Text, "<@"+botID+"> ")
+			msg := ev.Text
+
+			// Trim @BotKube prefix if exists
+			if strings.HasPrefix(ev.Text, "<@"+botID+"> ") {
```
ssudake21: this seems to be correct place to trim message. Rest are redundant and logic can be arranged so that code arrives here.

pkg/slack/slack.go
```diff
@@ -56,16 +56,44 @@ func (b *Bot) Start() {
...
 
 		case *slack.MessageEvent:
-			// Serve only if mentioned
-			if !strings.HasPrefix(ev.Text, "<@"+botID+"> ") {
+			// Skip if message posted by BotKube
+			if ev.User == botID {
 				continue
 			}
+
+			inMessage := ""
```
ssudake21: remove this def. redundant

PR #39, opened 2019-02-20
Added check slack channel in config, when get message 

gimmetm:
For cluster-channel pair use

### DIFF
--- a/pkg/slack/slack.go
+++ b/pkg/slack/slack.go
@@ -14,6 +14,8 @@ import (
 type Bot struct {
 	Token        string
 	AllowKubectl bool
+	ChannelName  string
+	CheckChannel bool
 }
 
 // slackMessage contains message details to execute command and send back the result
@@ -35,6 +37,8 @@ func NewSlackBot() *Bot {
 	return &Bot{
 		Token:        c.Communications.Slack.Token,
 		AllowKubectl: c.Settings.AllowKubectl,
+		ChannelName:  c.Communications.Slack.Channel,
+		CheckChannel: c.Settings.CheckChannel,
 	}
 }
 
@@ -61,21 +65,17 @@ func (b *Bot) Start() {
 				continue
 			}
 
-			// Check if message posted in a channel
-			_, err := api.GetChannelInfo(ev.Channel)
+			info, err := api.GetConversationInfo(ev.Channel, true)
 			if err == nil {
-				// Message posted in a channel
-				// Serve only if starts with mention
-				if !strings.HasPrefix(ev.Text, "<@"+botID+"> ") {
-					continue
-				}
-			} else {
-				// Check if message posted in a group
-				_, err := api.GetGroupInfo(ev.Channel)
-				if err == nil {
-					// Message posted in a group
+				if info.IsChannel || info.IsPrivate {
+					// Message posted in a channel
 					// Serve only if starts with mention
 					if !strings.HasPrefix(ev.Text, "<@"+botID+"> ") {
+						continue
+					}
+					// if config.settings.checkChannel is true
+					// Serve only if current channel is in config
+					if b.CheckChannel && (b.ChannelName != info.Name) {
 						continue
 					}
 				}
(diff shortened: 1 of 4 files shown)


PrasadG193:
Hey @gimmetm,
Thanks for the PR :slightly_smiling_face: 

Looks like `hack/verify-gofmt.sh` is failing.

```
$ hack/verify-gofmt.sh
diff -u ./pkg/slack/slack.go.orig ./pkg/slack/slack.go
--- ./pkg/slack/slack.go.orig	2019-02-20 05:50:34.363532422 +0000
+++ ./pkg/slack/slack.go	2019-02-20 05:50:34.363532422 +0000
@@ -75,7 +75,7 @@
 				}
 				// if checkChannel is true
 				// Serve only if current channel is in config
-				if b.CheckChannel && (b.ChannelName != channel.Name){
+				if b.CheckChannel && (b.ChannelName != channel.Name) {
 					continue
 				}
 
@@ -90,7 +90,7 @@
 					}
 					// if checkChannel is true
 					// Serve only if current group is in config
-					if b.CheckChannel && (b.ChannelName != group.Name){
+					if b.CheckChannel && (b.ChannelName != group.Name) {
 						continue
 					}
 				}
Run ./hack/update-gofmt.sh
The command "hack/verify-gofmt.sh" failed and exited with 1 during .
```
Running `hack/update-gofmt.sh` should fix that.

PrasadG193:
```
  # Channels configuration
  communications:
    slack:
      channel: 'SLACK_CHANNEL'
```
@gimmetm, the communication.slack.channel in the configuration is only for restricting push notifications from BotKube. So as per the design, we should be able to execute `kubectl` commands from any channel, group (shared chat) or even from Direct Message to BotKube.
I am trying to understand what use case you are trying to solve?

PrasadG193:
BTW we are disabling `notifier` commands execution outside the channel as a part of separate issue

gimmetm:
disabling notifier commands excution outside the channel sounds like solve
my issue.
in case of multi cluster and each channel for each cluster, excute command
message shows other channel(cluster) info too. I don’t want show other
cluster command result on the channel.

thank you.


2019년 2월 20일 (수) 19:48, Prasad Ghangal <notifications@github.com>님이 작성:

pkg/slack/slack.go
```diff
@@ -62,22 +66,33 @@ func (b *Bot) Start() {
...
+				group, err := api.GetGroupInfo(ev.Channel)
 				if err == nil {
 					// Message posted in a group
 					// Serve only if starts with mention
 					if !strings.HasPrefix(ev.Text, "<@"+botID+"> ") {
 						continue
 					}
+					// if checkChannel is true
+					// Serve only if current group is in config
+					if b.CheckChannel && (b.ChannelName != group.Name) {
```
PrasadG193: @gimmetm can you please remove this check? We should be able to execute BotKube commands from group chats.

PrasadG193:
Also, please squash your commits into one

pkg/slack/slack.go
gimmetm: @PrasadG193  This check is for private channel. (https://api.slack.com/methods/groups.info)
group chat info might be in conversation.info or else.

PrasadG193:
@gimmetm actually, client.GroupInfo() (https://godoc.org/github.com/nlopes/slack#Client.GetGroupInfo) returns true for a private channel as well as group chat. We need to find a better way to differentiate them.

gimmetm:
@PrasadG193 I replaced GetChannelInfo&GetGroupInfo -> GetConversationInfo, 
and checked (IsChannel, IsPrivate)
Please review it.

PrasadG193:
Nice! But still, group BotKube is not responding in group chat. Can you please test it on your setup?

PrasadG193:
> @gimmetm actually, client.GroupInfo() (https://godoc.org/github.com/nlopes/slack#Client.GetGroupInfo) returns true for a private channel as well as group chat. We need to find a better way to differentiate between them.

Let's track this as a separate issue

Issue #41, opened 2019-02-28
Reduce docker image size


PrasadG193:
Currently, docker image size is around 572 MB. The major reason behind that is, we are coping source code into the docker base image and building the binary there (https://github.com/infracloudio/botkube/blob/develop/build/Dockerfile). Instead we should be generating binary on local env and copy that binary into the docker image.
So we should have architecture specific build scripts (atleast linux/amd64 for now) which will generate binary and we should modify Dockerfile to just copy generated binary to docker image.
Something like this: https://github.com/fission/fission-workflows/tree/master/build

PR #44, opened 2019-03-01
Issue #41: Updated Dockerfile for multi-stage build

mugdha-adhav:


build/Dockerfile
```diff
@@ -1,9 +1,5 @@
-FROM golang:1.11-alpine
```
ssudake21: Anyway we can pass GOOS and GOARCH  while build docker image ?

PR #52, opened 2019-03-11
Modified cluster flag to support quotes

mugdha-adhav:


pkg/execute/executor.go
```diff
@@ -146,8 +144,21 @@ func runKubectlCommand(args []string, clusterName string, isAuthChannel bool) st
 		}
 		if strings.HasPrefix(arg, ClusterFlag.String()) {
 			if arg == ClusterFlag.String() {
+				if index == len(args)-1 {
+					return ""
+				}
+				arg = args[index+1]
```
PrasadG193: @mugdha-adhav can we use `continue` here?

mugdha-adhav: As discussed, will add comments to make code readable.

pkg/execute/executor.go
```diff
@@ -144,10 +142,27 @@ func runKubectlCommand(args []string, clusterName string, isAuthChannel bool) st
...
 			continue
 		}
+		// Check --cluster-name flag
 		if strings.HasPrefix(arg, ClusterFlag.String()) {
+			// Check if flag value is in current or next argument
 			if arg == ClusterFlag.String() {
+				if index == len(args)-1 {
+					return ""
+				}
+				arg = args[index+1]
```
PrasadG193: Can we avoid reassigning arg? The above `for loop` already takes care of it. Also rename `checkFlag` to something more meaningful like `isClusterFlagPresent`? `continue` here and process the value of `--cluster-name` in the next iteration.

mugdha-adhav: Made required changes and commited.

PR #55, opened 2019-03-14
Add support for Mattermost

mugdha-adhav:


cmd/botkube/main.go
```diff
@@ -17,8 +18,15 @@ func main() {
...
+		log.Logger.Info("Starting slack bot")
+		sb := slack.NewSlackBot()
+		go sb.Start()
 
-	controller.RegisterInformers(Config)
+		controller.RegisterInformers(Config)
+	} else if Config.Communications.Mattermost.Enabled {
+		log.Logger.Info("Starting mattermost bot")
+		mb := mattermost.NewMattermostBot()
+		mb.Start()
```
ssudake21: This is not a go routine ?

mugdha-adhav: Yes it will be. Will update it in the final commit.

ssudake21: No go routine yet ?

helm/botkube/values.yaml
```diff
@@ -7,7 +7,7 @@ replicaCount: 1
 image:
   repository: infracloud/botkube
   tag: "0.6"
-  pullPolicy: Always
+  pullPolicy: Always
```
ssudake21: fix ws

cmd/botkube/main.go
mugdha-adhav: We need to perform mandatory checks for connectivity/user/team/channel in Mattermost. So go routine will start in pkg/mattermost/mattermost.go

pkg/mattermost/mattermost.go
```diff
@@ -0,0 +1,236 @@
...
+var webSocketClient *model.WebSocketClient
+var botUser *model.User
+var botTeam *model.Team
+var botChannel *model.Channel
+
+const (
+	// BotName stores Botkube details
+	BotName = "botkube"
+	// BotPassword stores password for Botkube
+	BotPassword = "BotPassword@123"
```
ssudake21: hardcoding ?

pkg/controller/controller.go
```diff
@@ -200,8 +200,14 @@ func sendEvent(obj interface{}, kind, eventType string, err error) {
 	}
 
 	// Send notification to communication chennel
-	notifier := notify.NewSlack()
-	notifier.SendEvent(event)
+	notifier, status := notify.NewSlack()
```
ssudake21: no checks if slack/mattermost enabled? Also, can we have some mechanism where for notifying on slack we are not waiting for mattermost and vice versa. Basically notifying can be asynchronous

pkg/mattermost/mattermost.go
```diff
@@ -0,0 +1,236 @@
...
+	logging.Logger.Info("Channel created successfully in Mattermost")
+
+	// Adding Botkube user to channel
+	_, resp = client.AddChannelMember(botChannel.Id, botUserID)
+	logging.Logger.Info("Botkube User added to the Channel")
+	return botChannel, nil
+}
+
+// Check incomming message and take action
+func handleMMessage(event *model.WebSocketEvent, allowkubectl bool, clusterName, channelName string) {
```
ssudake21: can be handleMessage only as its in mattermost package and private function. Same for rest of private functions

Gopkg.lock
```diff
@@ -1,6 +1,14 @@
 # This file is autogenerated, do not edit; changes may be undone by the next 'dep ensure'.
 
 
+[[projects]]
+  branch = "master"
+  digest = "1:e95c8ee2ac122e3b55a7bef845c4c65bbd240685da1dbb48aee2d6854e16c758"
+  name = "github.com/alecthomas/log4go"
```
ssudake21: are we explicitly adding log4go package  or added by mattermost server ?

pkg/mattermost/mattermost.go
mugdha-adhav: We will mention in the Readme that this is the default password for Botkube user. It is recommended to change the password in settings.

helm/botkube/values.yaml
mugdha-adhav: sure.

pkg/mattermost/mattermost.go
mugdha-adhav: Yes. All the functions in mattermost package will be private.

pkg/controller/controller.go
mugdha-adhav: The status flag is added to check if Slack/Mattermost is enabled. Will figure out some mechanism to make asynchronous calls for notifier.

Gopkg.lock
mugdha-adhav: Will check on that and also push the pending mattermost-server dependencies in vendor.

pkg/controller/controller.go
PrasadG193: @mugdha-adhav let's call SendEvent and SendMessage as a go routine.
ref: https://github.com/infracloudio/botkube/pull/59/files#diff-3c862eb54a8e0e161b534e0c67e5379eR163-R213

pkg/mattermost/mattermost.go
mugdha-adhav: User creation will be done by the admin itself. So no need of hardcoding password.

pkg/controller/controller.go
mugdha-adhav: sure @PrasadG193 changes made respectively and will update in next commit.

Gopkg.lock
mugdha-adhav: Checked on the dependencies. All the added dependencies are required for Mattermost.

ssudake21:
@mugdha-adhav  Lets rebase this once with PR  #59 . Seeing lot of duplicate changes.  Will check all changes in one go

pkg/notify/mattermost.go
```diff
@@ -0,0 +1,142 @@
...
+		Channel:     botChannel.Id,
+		ClusterName: c.Settings.ClusterName,
+	}
+}
+
+// SendEvent sends event notification to Mattermost
+func (m *Mattermost) SendEvent(event events.Event) error {
+	log.Logger.Info(fmt.Sprintf(">> Sending to Mattermost: %+v", event))
+
+	fields := []*model.SlackAttachmentField{
```
ssudake21: SlackAttachmentField should be renamed ?

mugdha-adhav: That's a structure in mattermost-server/model.

PrasadG193:
@mugdha-adhav please update https://github.com/infracloudio/botkube/blob/infracloudio/mattermost-support/deploy-all-in-one.yaml with the configuration changes

pkg/mattermost/mattermost.go
```diff
@@ -0,0 +1,208 @@
...
+
+const (
+	// BotName stores Botkube details
+	BotName = "botkube"
+	// ChannelPurpose describes the purpose of Botkube channel
+	ChannelPurpose = "Botkube alerts"
+	// WebSocketProtocol stores protocol initials for web socket
+	WebSocketProtocol = "ws:"
+	// Logs file name
+	Logs = "logs"
```
PrasadG193: @mugdha-adhav why do we need this?

pkg/mattermost/mattermost.go
```diff
@@ -0,0 +1,208 @@
...
+}
+
+// Send messages to Mattermost
+func sendMessage(msg, postID, channelID string) {
+	post := &model.Post{}
+	post.ChannelId = channelID
+	post.RootId = postID
+	// Create file if message is too large
+	if len(msg) >= 3990 {
+		res, resp := client.UploadFileAsRequestBody([]byte(msg), channelID, Logs)
```
PrasadG193: @mugdha-adhav it is not necessary that only logs messages would have size greater than 3990. Some responses like, `get pods --all-namespaces` can also be too long. We should set title of the file same as the command we are executing (just like we do it for slack)
https://github.com/infracloudio/botkube/blob/master/pkg/slack/slack.go#L136

pkg/mattermost/mattermost.go
PrasadG193: I am assuming this is the name of file if we are returning response as a file?

mugdha-adhav: Updated the logs filename.

PrasadG193:
@mugdha-adhav not a part of this PR, but could you remove `func formatAndSendLogs(rtm *slack.RTM, channelID, logs string, filename string)` function, I don't think we need that anymore.
https://github.com/infracloudio/botkube/blob/master/pkg/slack/slack.go#L119

pkg/mattermost/mattermost.go
```diff
@@ -0,0 +1,206 @@
...
+func getBotUser(teamID string) (*model.User, error) {
+	users, resp := client.AutocompleteUsersInTeam(teamID, BotName, "")
+	if resp.Error != nil {
+		return nil, resp.Error
+	}
+	return users.Users[0], nil
+}
+
+// Create channel if not present and add botkube user in channel
+func getBotChannel(channelName, botTeamID, botUserID string) (*model.Channel, error) {
```
PrasadG193: Can we create MattermostMessage{} struct with all required variables and implement methods on that instead of passing parameters in each function?

mugdha-adhav: Sure. Will change and commit.

README.md
```diff
@@ -31,7 +31,9 @@ $ git clone https://github.com/infracloudio/botkube.git
 If you are not interested in events about particular resource, just remove it's entry from the config file.
 - Deploy BotKube controller using **helm install** in your cluster.
 ```bash
-$ helm install --name botkube --namespace botkube --set config.communications.slack.channel={SLACK_CHANNEL_NAME},config.communications.slack.token={SLACK_API_TOKEN_FOR_THE_BOT},config.settings.clustername={CLUSTER_NAME},config.settings.allowkubectl={ALLOW_KUBECTL} helm/botkube/
+$ helm install --name botkube --namespace botkube --set
```
ssudake21: Lets make this more readable as in botkube docs.
 Ideally we should remove installation steps from README and just rely on botkube docs

README.md
```diff
@@ -72,6 +74,34 @@ $ kubectl create ns botkube && kubectl create -f deploy-all-in-one.yaml -n botku
...
+
+**MATTERMOST_SERVER_URL** is the URL where Mattermost is running<br>
+**MATTERMOST_TOKEN** is the Token you received after installing BotKube user and creating Personal Access Token<br>
+**MATTERMOST_TEAM** is the team name where BotKube will be added<br>
+**MATTERMOST_CHANNEL** is the channel name where BotKube will be added<br>
+
+#### Using helm
+
+```bash
+$ helm install --name botkube --namespace botkube --set
```
ssudake21: Same as above

README.md
mugdha-adhav: Updated the helm command in README.

PR #65, opened 2019-03-26
Hide sensitive info while displaying configuration

PrasadG193:


PrasadG193:
image

pkg/execute/executor.go
```diff
@@ -234,19 +235,34 @@ func runPingCommand(args []string, clusterName string) string {
...
 	b, err := ioutil.ReadAll(file)
 	if err != nil {
-		return "", err
+		return configYaml, err
 	}
 
-	return string(b), nil
+	c := &config.Config{}
+	if len(b) != 0 {
+		yaml.Unmarshal(b, c)
```
ssudake21: do we expect any errors in unmarshal ?

pkg/execute/executor.go
```diff
@@ -234,19 +235,34 @@ func runPingCommand(args []string, clusterName string) string {
...
+	}
+
+	// hide sensitive info
+	c.Communications.Slack.Token = ""
+	c.Communications.ElasticSearch.Password = ""
```
ssudake21: if we have omitempty above why do we need to set empty strings again ?

PrasadG193: `omitempty` drops fields if they are empty while Marshaling. When we Unmarshal yaml to struct, we get Token and Password set in the struct. So in order to drop the fields, we need to make their value empty

pkg/execute/executor.go
PrasadG193: In the worst case, we will get empty values of keys don't match. Will handle error though

PR #67, opened 2019-03-27
Add Skip var in Event struct to skip an event

PrasadG193:
Fixes: https://github.com/infracloudio/botkube/issues/63

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -186,13 +186,17 @@ func sendEvent(obj interface{}, c *config.Config, kind, eventType string, err er
 
 	// Check if Notify disabled
 	if !config.Notify {
-		log.Logger.Info("Skipping notification")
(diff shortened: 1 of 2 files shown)


pkg/controller/controller.go
```diff
@@ -186,13 +186,17 @@ func sendEvent(obj interface{}, c *config.Config, kind, eventType string, err er
...
-		log.Logger.Info("Skipping notification")
+		log.Logger.Debug("Skipping notification")
 		return
 	}
 
 	// Create new event object
 	event := events.New(obj, eventType, kind)
 	event = filterengine.DefaultFilterEngine.Run(obj, event)
+	if event.Skip {
+		log.Logger.Debug("Skipping event")
```
ssudake21: Add event data as well

PrasadG193: Done

PR #73, opened 2019-04-01
Return BotKube version with response to ping

PrasadG193:
Fixes: https://github.com/infracloudio/botkube/issues/62

### DIFF
--- a/helm/botkube/values.yaml
+++ b/helm/botkube/values.yaml
@@ -6,7 +6,6 @@ replicaCount: 1
 
 image:
   repository: infracloud/botkube
-  tag: "0.6"
   pullPolicy: Always
 
 nameOverride: ""
--- a/pkg/execute/executor.go
+++ b/pkg/execute/executor.go
@@ -24,7 +24,6 @@ var validKubectlCommands = map[string]bool{
 	"get":           true,
 	"logs":          true,
 	"top":           true,
-	"version":       true,
 	"auth":          true,
 }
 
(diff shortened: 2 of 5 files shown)


PrasadG193:
image

helm/botkube/values.yaml
```diff
@@ -6,7 +6,6 @@ replicaCount: 1
 
 image:
   repository: infracloud/botkube
-  tag: "0.6"
```
ssudake21: How we are deciding image version after removal of tag ?

pkg/execute/executor.go
```diff
@@ -213,7 +222,16 @@ func runNotifierCommand(args []string, clusterName string, isAuthChannel bool) s
 	return printDefaultMsg()
 }
 
-func runPingCommand(args []string, clusterName string) string {
+func findBotKubeVersion() (versions string) {
+	args := []string{"-c", fmt.Sprintf("%s version --short=true | awk '/Server/ {print $3}'", kubectlBinary)}
+	cmd := exec.Command("sh", args...)
+	k8sVersion, _ := cmd.CombinedOutput()
```
ssudake21: please handle error here

ssudake21: LOG warning/error based on severity of error

pkg/execute/executor.go
```diff
@@ -213,7 +222,16 @@ func runNotifierCommand(args []string, clusterName string, isAuthChannel bool) s
...
+
+	botkubeVersion := os.Getenv("BOTKUBE_VERSION")
```
ssudake21: If BotKube version is empty, we should show Unknown or nil

helm/botkube/values.yaml
PrasadG193: We'll use chart version as an image tag
```
          image: "{{ .Values.image.repository }}:{{ .Chart.Version }}"
```

PR #74, opened 2019-04-03
Add logic to restart BotKube pod if config file is updated

PrasadG193:
Fixes: https://github.com/infracloudio/botkube/issues/72

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -4,11 +4,13 @@ import (
 	"fmt"
 	"os"
 	"os/signal"
+	"path/filepath"
 	"strconv"
 	"strings"
 	"syscall"
 	"time"
 
+	"github.com/fsnotify/fsnotify"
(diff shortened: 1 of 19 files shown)


pkg/controller/controller.go
```diff
@@ -235,3 +241,46 @@ func sendMessage(c *config.Config, msg string) {
...
+	if err != nil {
+		log.Logger.Fatal("Failed to create file watcher ", err)
+	}
+	defer watcher.Close()
+
+	done := make(chan bool)
+	go func() {
+		for {
+			select {
+			case _, ok := <-watcher.Events:
```
ssudake21: Do we need to check type of events?
What kind of events we can get on channel ?

PrasadG193: These are file change events (not k8s events). Whenever file gets updated, we get event and we restart the pod.

PR #82, opened 2019-04-16
feature set command

nnao45:
Hi

I add "set" command botkube.
because for using ChatOps Pipeline with botkube.

スクリーンショット 2019-04-16 18 32 42

this PR, fix parse slack url encording in pkg/execute/executor.go 
because, container image url, example `set image deployment nginx-svc nginx=gcr.io/hoge/hello-world-container`, 
slack print `set image deployment nginx-svc nginx=<https://gcr.io/hoge/hello-world-container|gcr.io/hoge/hello-world-container>` and sending to botkube.

### DIFF
--- a/helm/botkube/templates/clusterrole.yaml
+++ b/helm/botkube/templates/clusterrole.yaml
@@ -10,4 +10,4 @@ metadata:
 rules:
   - apiGroups: ["*"]
     resources: ["*"]
-    verbs: ["get", "watch", "list"]
+    verbs: ["get", "watch", "list", "patch"]
--- a/pkg/execute/executor.go
+++ b/pkg/execute/executor.go
@@ -25,6 +25,7 @@ var validKubectlCommands = map[string]bool{
 	"logs":          true,
 	"top":           true,
 	"auth":          true,
+	"set":           true,
 }
 
 var validNotifierCommand = map[string]bool{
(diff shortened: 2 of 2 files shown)


ssudake21:
Hey @nnao45  This would require WRITE access to Kubernetes cluster. Currently, we consider only commands which get access through a READONLY service account.

nnao45:
@ssudake21 
> we consider only commands which get access through a READONLY service account.
umm....Okay. But Now I use this under production env, is very useful...
you don't think botkube WRITABLE mode?

ssudake21:
We avoid updates to cluster for security reasons &  we believe if you need to make changes in K8s cluster it should be done in a more sane manner through helm charts or your versioned Kubernetes yamls.

Having said that, we are currently thinking of adding some configuration options which can give user ability to enable specific commands with a write-enabled service account, which can also satisfy the above use case.
This would be on the user though to enable specific commands he needs. Botkube will just check if command is allowed or not.

nnao45:
thanks!! I understand.
security considering is very important..!

> Having said that, we are currently thinking of adding some configuration options which can give user ability to enable specific commands with a write-enabled service account, which can also satisfy the above use case.

it is very cool!! I will help this that I can so if you need help, please tell me.

PR #89, opened 2019-05-09
[helm chart] make image tag configurable

### LINKED ISSUE #88: Image tag in Helm chart should be configurable
If we build BotKube image for some bug fix or want to use image with a tag like `latest` then one should be able to set that image tag while installing BotKube with Helm chart. 

Changing the version from `Chart.yaml` is not a clean solution, also it won't allow to set it to tag values which are not according to sem ver 2. We should have `.Values.image.tag`

https://helm.sh/docs/developing_charts/#charts-and-versioning

bhavin192:
- If image.tag is set in values.yaml then that tag is used, otherwise
  falls back to Chart version
- This makes easy to deploy images tagged with different names which
  are not according to sem ver 2 i.e. 'latest' or other valid
  container image tags

fixes #88

### DIFF
--- a/helm/botkube/values.yaml
+++ b/helm/botkube/values.yaml
@@ -7,6 +7,9 @@ replicaCount: 1
 image:
   repository: infracloud/botkube
   pullPolicy: Always
+  ## default tag is appVersion from Chart.yaml. If you want to use
+  ## some other tag then it can be specified here
+  # tag: latest
 
 nameOverride: ""
 fullnameOverride: ""
--- a/helm/botkube/Chart.yaml
+++ b/helm/botkube/Chart.yaml
@@ -2,6 +2,7 @@ apiVersion: v1
 name: botkube
 home: https://www.botkube.io
 version: 0.7.0
+appVersion: 0.7.0
 icon: https://www.botkube.io/images/botkube.ico
 description: Controller for the BotKube Slack app which helps you monitor your Kubernetes cluster,
  debug deployments and run specific checks on resources in the cluster.
--- a/helm/botkube/templates/deployment.yaml
+++ b/helm/botkube/templates/deployment.yaml
@@ -26,7 +26,7 @@ spec:
       serviceAccountName: {{ include "botkube.fullname" . }}-sa
       containers:
         - name: {{ .Chart.Name }}
-          image: "{{ .Values.image.repository }}:{{ .Chart.Version }}"
+          image: "{{ .Values.image.repository }}:{{ default .Chart.AppVersion .Values.image.tag }}"
           imagePullPolicy: {{ .Values.image.pullPolicy }}
           volumeMounts:
             - name: config-volume


helm/botkube/values.yaml
```diff
@@ -7,6 +7,7 @@ replicaCount: 1
 image:
   repository: infracloud/botkube
   pullPolicy: Always
+  # tag: latest
```
ssudake21: We should keep this line uncommented. What is the default value we are considering currently?

bhavin192: Currently the default value is `{{ .Chart.Version }}` in deployment.yaml

bhavin192: @PrasadG193 ^

PrasadG193: @bhavin192 if we keep it blank, it should work. We should be able to set tag with `--set image.tag=latest` without making any change in the chart. Can you please check if the variable needs to be present in values.yaml if we want to set it with `--set` command?

bhavin192: >  if we keep it blank

Can you please elaborate this? Blank as in commented or like `tag: ""`?
> without making any change in the chart.

Do you mean without making changes to deployment.yaml? 

And yes, it works even if you don't have value in values.yaml but keeping it commented in `values.yam` would make it easy for people to figure out that it's possible to set tags as well.

PrasadG193: > Can you please elaborate this? Blank as in commented or like tag: ""?
Yes, or something like `tag:  `. I think, commenting field will create confusion

bhavin192: I feel blank field will create confusion :smile:, how about adding some explanation like: 
```yaml
# default tag is Chart version which is picked from Charts.yaml if you
# want to use some other tag it can be specified here
```

PrasadG193: yes, that sounds good

rimusz: Helm best practices recommend to use in `Chart.yaml` `appVersion:` to set application version.
Then it can be used in the deployment as: 
```
image: "{{ .Values.image.repository }}::{{ default .Chart.AppVersion .Values.image.tag }}"
```

bhavin192: And the chart version should be updated only when there is a change in the chart itself not the application, that's what I have seen with helm stable repo.

rimusz: no, if you update appversion, you need to update chat version as well!
or if appversion stays the same, chart gets changes, you update chat version.
Any changes to the chart you bump chart version :)

bhavin192: Yep, that's what I meant to say but missed the fact that changing Charts.yaml means changing chart version as well.

rimusz: 👍

bhavin192:
@ssudake21 @PrasadG193 should I address the changes suggested by rimusz in this PR (having appVersion in `Chart.yaml`) or should that be addressed in separate PR?

PrasadG193:
We can address those in the same PR

helm/botkube/values.yaml
bhavin192: @rimusz can I use this https://github.com/infracloudio/botkube/pull/89#discussion_r282802760 in my commit?

rimusz: sure

Issue #123, opened 2019-07-08
[BUG] BotKube sends older update events after resync


PrasadG193:
**Describe the bug**
Default resync period of BotKube is 30mins. If BotKube is configured to inform about update events, it sends update event notification after resync.

**To Reproduce**
1. Configure BotKube to receive update events of a resource (e.g job)
2. Wait for 30mins i.e default resync period
3. After resync, you will see older update event notifications

**Expected behavior**
BotKube should skip older update events

PR #124, opened 2019-07-08
Fix timestamp for update events

PrasadG193:
##### ISSUE TYPE

 - Bug fix Pull Request

##### SUMMARY
Fix timestamp for update events
Ignore older update events after resync

Fixes #123

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -47,7 +47,7 @@ func findNamespace(ns string) string {
 // RegisterInformers creates new informer controllers to watch k8s resources
 func RegisterInformers(c *config.Config) {
 	sendMessage(c, fmt.Sprintf(controllerStartMsg, c.Settings.ClusterName))
-	startTime = time.Now().Local()
+	startTime = time.Now()
 
 	// Start config file watcher
 	go configWatcher(c)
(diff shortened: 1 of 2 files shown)


pkg/events/events.go
```diff
@@ -103,16 +103,40 @@ func New(object interface{}, eventType config.EventType, kind string) Event {
...
 		event.Kind = "Pod"
+		if eventType == config.UpdateEvent {
+			condLen := len(obj.Status.Conditions)
+			if condLen != 0 {
+				event.TimeStamp = obj.Status.Conditions[condLen-1].LastTransitionTime.Time
+			}
+		}
 	case *apiV1.Node:
 		event.Kind = "Node"
+		if eventType == config.UpdateEvent {
```
ssudake21: See this snippet being repeated across all conditions? Can we merge these somehow? See a lot of code duplication.

pkg/controller/controller.go
```diff
@@ -47,7 +47,7 @@ func findNamespace(ns string) string {
 // RegisterInformers creates new informer controllers to watch k8s resources
 func RegisterInformers(c *config.Config) {
 	sendMessage(c, fmt.Sprintf(controllerStartMsg, c.Settings.ClusterName))
-	startTime = time.Now().Local()
+	startTime = time.Now()
```
ssudake21: Whats difference between `time.Now()` and `time.Now().Local()` ?

PrasadG193: Local() sets location with Time

pkg/events/events.go
PrasadG193: To access resource obj, we need to typecast `runtime.Object` interface into resource object. That's what we do with switch cases. We can't access obj.Status from interface or without typecasting it

PR #138, opened 2019-08-01
[feature] Add Object Annotation filter, Fixes #132, #133

### LINKED ISSUE #132: Annotations Based Multi-Channel support
Presently, Botkube sends all the notifs to a particular channel. An organisation with mulyiple teams governing individual services would require seperate notif channel to moniter their deployed services.

Kubenetes resources when annotated with, for instance `botkube.io/channel: frontend`
Botkube reads them and sends all the notifs for the particular resource to the `frontend` channel. 

This provides a way to classify notifications based on internal teams or workgroup by using the annotation data available in the resources.

### LINKED ISSUE #133: Annotations based enable/disable notifications for a particular object.
Disable notification for a particular noisy resource ( least important or consistantly crashing pod) based on 
Kubernetes annotations at run time.

Adding an annotation to a kubernetes resource at run time, for instance `botkube.io/disable-notif: true`
will skip the notifications from being sent to the channels.

This provides us control to enable or disable notifications of a particular object at run time.

This can be implemented by adding an annotations fliter.

aananthraj:
##### ISSUE TYPE

 - Feature Pull Request

##### SUMMARY

This commit, 
- enables filtering of events based on annotations present in objects at run time.
- annotation `botkube.io/disable: true` disables event notifications for the annotated object
- annotation `botkube.io/channel: <channel_name>` sends events notifications of the annotated object to the mentioned channel.
- adds func `utils.ExtractAnnotations()`. It extracts annotations from Event.InvolvedObject.  Extracted Annotations are then added to the event.Metadata.Annotations
- implements individual actions using internal functions.
- adds unit tests for internal functions.
- replaces `utils.Init()` with `InitialiseKubeClient()` to decouple `config.yaml` and `KubeClinet` dependencies from unit testing

Fixes #132

Adding `botkube.io/channel: <channel_name>` annotation to a kubenetes resources sends all the notifications of the annotated object to specified channel in the workspace.
This provides a way to classify notifications based on internal teams or workgroup by using the annotation data available in the resources.
  
Fixes #133 

Adding `botkube.io/disable: true` annotation to a kubernetes resource, skips the notifications about the annotated object from being sent to the channels.
This provides us the control to enable or disable notifications of a particular object at run time.
Up

### DIFF
--- a/pkg/filterengine/filters/object-annotation-filter/object_annotation_checker.go
+++ b/pkg/filterengine/filters/object-annotation-filter/object_annotation_checker.go
@@ -1,0 +1,76 @@
+package filters
+
+import (
+	"github.com/infracloudio/botkube/pkg/events"
+	"github.com/infracloudio/botkube/pkg/filterengine"
+	log "github.com/infracloudio/botkube/pkg/logging"
+	"github.com/infracloudio/botkube/pkg/utils"
+	metaV1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	// DisableAnnotation is the object disable annotation
+	DisableAnnotation string = "botkube.io/disable"
+	// ChannelAnnotation is the multichannel support annotation
+	ChannelAnnotation string = "botkube.io/channel"
+)
+
+// ObjectAnnotationChecker add recommendations to the event object if pod created without any labels
+type ObjectAnnotationChecker struct {
+	Description string
+}
+
+// Register filter
+func init() {
+	filterengine.DefaultFilterEngine.Register(ObjectAnnotationChecker{
+		Description: "Checks if annotations botkube.io/* present in object specs and filters them.",
+	})
+}
+
+// Run filters and modifies event struct
+func (f ObjectAnnotationChecker) Run(object interface{}, event *events.Event) {
+
+	// get objects metadata
+	obj := utils.GetObjectMetaData(object)
+
+	// Check annotations in object
+	if isObjectNotifDisabled(obj) {
+		event.Skip = true
+		log.Logger.Debug("Object Notification Disable through annotations")
+	}
+
+	if channel, ok := reconfigureChannel(obj); ok {
+		event.Channel = channel
+		log.Logger.Debugf("Redirecting Event Notifications to channel: %s", channel)
+	}
+
+	log.Logger.Debug("Object annotations filter successful!")
+}
+
+// Describe filter
+func (f ObjectAnnotationChecker) Describe() string {
+	return f.Description
+}
+
+// isObjectNotifDisabled checks annotation botkube.io/disable
+// annotation botkube.io/disable disables the event notifications from objects
+func isObjectNotifDisabled(obj metaV1.ObjectMeta) bool {
+
+	if obj.Annotations[DisableAnnotation] == "true" {
+		log.Logger.Debug("Skipping Disabled Event Notifications!")
+		return true
+	}
+	return false
+}
+
+// reconfigureChannel checks annotation botkube.io/channel
+// annotation botkube.io/channel directs event notifications to channels
+// based on the channel names present in them
+// Note: Add botkube app into the desired channel to receive notifications
+func reconfigureChannel(obj metaV1.ObjectMeta) (string, bool) {
+	// redirect messages to channels based on annotations
+	if channel, ok := obj.Annotations[ChannelAnnotation]; ok {
+		return channel, true
+	}
+	return "", false
+}
--- a/pkg/filterengine/filters/object-annotation-filter/object_annotation_checker_test.go
+++ b/pkg/filterengine/filters/object-annotation-filter/object_annotation_checker_test.go
@@ -1,0 +1,50 @@
+package filters
+
+import (
+	"testing"
+
+	metaV1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestIsObjectNotifDisabled(t *testing.T) {
+	tests := map[string]struct {
+		annotaion metaV1.ObjectMeta
+		expected  bool
+	}{
+		`Empty ObjectMeta`:                 {metaV1.ObjectMeta{}, false},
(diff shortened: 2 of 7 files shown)


aananthraj:
rebased to develop

PrasadG193:
Thanks for the PR @aananthraj! Please rebase your branch with develop.

aananthraj:
@PrasadG193, Rebased to develop. thanks :+1:

pkg/controller/controller.go
```diff
@@ -228,7 +228,29 @@ func sendEvent(obj interface{}, c *config.Config, kind string, eventType config.
...
 		notifier = notify.NewSlack(c)
-		go notifier.SendEvent(event)
+
+		// empty value in event.channel sends notifications to default channel.
+		if event.Channel == "" {
+			go notifier.SendEvent(event)
+		} else {
+			// non empty value in event.channel demands redirection of events to a different channel
+			// clone config and update Slack.Channel
+			newC := *c
```
ssudake21: I believe this should be handled in `notifier.SendEvent`, where `event.Channel` if set we send to channel from event.
Logic should be more cleaner in this case. notifier will give preference to `event.Channel` over configured channel.
This also gives flexibility for notifier to use Channel as it wants , across different type of notifiers.

pkg/filterengine/filters/object_annotation_checker.go
```diff
@@ -0,0 +1,54 @@
...
+func (f ObjectAnnotationChecker) Run(object interface{}, event *events.Event) {
+
+	// get objects metadata
+	obj := utils.GetObjectMetaData(object)
+
+	// Check annotations in object
+	if len(obj.Annotations) != 0 {
+		// check annotation botkube.io/disable
+		// annotation botkube.io/disable disables the event notifications from objects
+		if obj.Annotations["botkube.io/disable"] == "true" {
```
ssudake21: This filter looks good. Please see if you can make annotation specific actions as internal methods of this filter.
It should be easier to add actions for any future annotations we may desire.

aananthraj: resolved it

pkg/controller/controller.go
aananthraj: resolved it.

aananthraj:
Resolved requested changes.

aananthraj:
Rebased to develop

pkg/filterengine/filters/object-annotation-filter/object_annotation_checker.go
```diff
@@ -0,0 +1,83 @@
...
+	})
+}
+
+// Run filters and modifies event struct
+func (f ObjectAnnotationChecker) Run(object interface{}, event *events.Event) {
+
+	// get objects metadata
+	obj := utils.GetObjectMetaData(object)
+
+	// if eventObj, ok := object.(*apiV1.Event); ok {
```
ssudake21: Please removed dead code if not needed.

pkg/utils/utils.go
```diff
@@ -257,3 +268,121 @@ func DeleteDoubleWhiteSpace(slice []string) []string {
 	}
 	return result
 }
+
+// ExtractAnnotaions returns annotations of InvolvedObject for the given event
+func ExtractAnnotaions(obj *apiV1.Event) map[string]string {
```
ssudake21: This works for object which are deleted ?
We may try to extract embedded object which is delete ?

pkg/filterengine/filters/object-annotation-filter/object_annotation_checker.go
aananthraj: sure

pkg/utils/utils.go
aananthraj: > We may try to extract embedded object which is deleted?

I would like to discuss the workflows here

In general, 
- we use ExtractAnnotaions only for object kind: Events
- Kubernetes events of type Normal/Warning are associated with an existing object
- In most cases, we can get those objects associated with events with no trouble.
- In case of error during the getting these objects, notifs will be sent to the default channel.

In this rare case, 
- ExtractAnnotaions() may try to get the particular deleted object which results in `Error from server (NotFound): <object> "foo" not found`
- ExtractAnnotaions() then informs the error and returns empty map
- No annotations from actual objects get updated in the current event object
- botkube sends notif to the default channel or We receive event notifications even after disabling it through annotations.

sanketsudake: Explanation looks good to me. Lets make sure we have unit tests around this scenario.

pkg/filterengine/filters/object-annotation-filter/object_annotation_checker_test.go
```diff
@@ -0,0 +1,50 @@
...
+	}
+}
+
+func TestReconfigureChannel(t *testing.T) {
+	tests := map[string]struct {
+		objectMeta      metaV1.ObjectMeta
+		expectedChannel string
+		expectedBool    bool
+	}{
+		`Empty ObjectMeta`:                    {metaV1.ObjectMeta{}, "", false},
```
aananthraj: I guess, this covers the emty object use case as well as the object not available (object deleted) use case.
Do correct me, if am missing something.

pkg/utils/utils.go
```diff
@@ -257,3 +268,121 @@ func DeleteDoubleWhiteSpace(slice []string) []string {
...
+		log.Logger.Error(err)
+	case "ClusterRoleBinding":
+		object, err := KubeClient.RbacV1().ClusterRoleBindings().Get(obj.InvolvedObject.Name, metaV1.GetOptions{})
+		if err == nil {
+			return object.ObjectMeta.Annotations
+		}
+		log.Logger.Error(err)
+	}
+
+	return map[string]string{}
```
aananthraj: This adds up with the discussion regarding the unit test case for object deleted

sanketsudake:
@aananthraj  Please rebase PR.

aananthraj:
Sure, thanks

PR #146, opened 2019-08-08
Use SharedInformerFactory instead of cache.Informer

### LINKED ISSUE #143: [Refactoring] Use SharedInformerFactory instead of cache.NewInformer to watch K8s resources
We should use SharedInformerFactory which has the following advantages over cache.NewInformer:
- Allows to have multiple instances of BotKube running without having duplicate events
- Works with Fake client (https://godoc.org/k8s.io/client-go/kubernetes/fake)

PrasadG193:
##### ISSUE TYPE

 - Refactoring

##### SUMMARY

This PR includes following fixes:
- Use SharedInformerFactory instead of cache.Informer
- Improve event filtering logic
- Fix timestamp for deployment/statefulset delete event
- Fix job_status_checker filter to allow create events

Fixes #143


PrasadG193:
cc: @aananthraj

ssudake21:
Please see if you can followup on below items.
1. Add update notes regarding configuration changes 
2. Identify if any performance benefits - CPU & Memory with this approach, just simple kubectl top
3. This should not break any existing functionalities, no regressions. We should add related tests. I know you have plan to add tests in another PR

PrasadG193:
> Add update notes regarding configuration changes

Sure

> Identify if any performance benefits - CPU & Memory with this approach, just simple kubectl top

@ssudake21 CPU usage has been reduced from 2m to 0m in idle condition as well as little workload. Memory usage is somewhat similar to older changes i.e around 15Mi.

> This should not break any existing functionalities, no regressions. We should add related tests. I know you have plan to add tests in another PR

We need these changes to start adding integration tests. I'll fix if I find any issue as a part of that PR

PR #154, opened 2019-08-13
Add E2E Integration tests

### LINKED ISSUE #136: Improve test coverage for BotKube
Currently we lack any kind of integration test or unit tests for BotKube. We plan to increase test coverage and add basic suite of tests for BotKube verification. 

1. Improved test coverage for BotKube
2. Run code coverage tools if possible to cover as much as current code.

**Additional context**

PrasadG193:
##### ISSUE TYPE
 - Feature Pull Request

##### SUMMARY

This PR:
- Adds E2E integration test framework
- Refactors Slack notifier to use latest nlopes/slack
- Refactor Bot implementations to work with mocked Slack
- Add CommandRunner interface and fake_executor to mock kubectl command execution
- Add warnings field to the Event struct
- Update filter recommendation message format
- Rename hack/unit-test.sh to hack/runtests.sh
- Modify make target unit-test to test

To Run: 
go test -tags=test ./test/e2e/ -v **OR** ./hack/runtests.sh **OR** make test

Fixes #136


PrasadG193:
Negative tests are yet to be added, will raise separate PR for that

test/e2e/command/botkube.go
```diff
@@ -0,0 +1,151 @@
...
+			}
+		})
+	}
+}
+
+func compareFilters(expected, actual []string) bool {
+	for _, a := range actual {
+		found := false
+		for _, e := range expected {
+			if a == e {
```
sanketsudake: This checks first line match only and then breaks immediately ?

PrasadG193: this function is for comparing if two string slices are the same or not. If a string in actual slice match with string in expected slice, we should move to the next string, that's why the break statement.
We should check length of the slices as well here

PR #156, opened 2019-08-15
add proxyURL, deployment env variable if-loop

### LINKED ISSUE #155: Add basic proxy support for helm deployment
No current way to add proxy info via values file

Here is a minor change based on my understanding of gorilla websocket https://github.com/opensourcebapu/botkube/commit/406b2736a6775fcd51f2422dba431ea37ba948c9

rajinator:
##### ISSUE TYPE

 - Feature Pull Request

##### SUMMARY
Adding Support for Basic Proxy

Need a proxy configured to use the proxy value. If statement in deployment.yaml will make sure deployment doesn't fail, if proxy is not being used

Fixes #155

### DIFF
--- a/helm/botkube/values.yaml
+++ b/helm/botkube/values.yaml
@@ -17,6 +17,9 @@ fullnameOverride: ""
 # set one of the log levels- info, warn, debug, error, fatal, panic
 logLevel: info
 
+# Proxy URL to use
+proxyURL: None
+
 config:
   ## Resources you want to watch
   resources:
--- a/helm/botkube/values.yaml
+++ b/helm/botkube/values.yaml
@@ -17,9 +17,6 @@ fullnameOverride: ""
 # set one of the log levels- info, warn, debug, error, fatal, panic
 logLevel: info
 
-# Proxy URL to use
-proxyURL: None
-
 config:
   ## Resources you want to watch
   resources:
@@ -203,6 +200,9 @@ resources: {}
   #  cpu: 100m
   #  memory: 128Mi
 
+extraEnv:   # Extra environment variables to pass to the botkube container, example HTTP_PROXY
+  #HTTP_PROXY: <proxyURL>:<port>
+  
 nodeSelector: {}
 
 tolerations: []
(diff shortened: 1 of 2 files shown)


bhavin192:
How about adding a section like `extraEnv`, so that we can add any number of environment variables to the container spec. @PrasadG193 what do you think?

Reference: 

https://github.com/helm/charts/blob/d8c4f7996f843f976b7e9e8213a5235fa950c13a/stable/kiam/templates/server-daemonset.yaml#L97-L103

https://github.com/helm/charts/blob/d8c4f7996f843f976b7e9e8213a5235fa950c13a/stable/kiam/values.yaml#L95

PrasadG193:
@bhavin192 sounds good to me. @rajinator could you please make those changes?

rajinator:
@PrasadG193 @bhavin192 @sanketsudake changes added, also added new issues #160 and #161

helm/botkube/values.yaml
```diff
@@ -200,6 +200,9 @@ resources: {}
   #  cpu: 100m
   #  memory: 128Mi
 
+extraEnv:   # Extra environment variables to pass to the botkube container, example HTTP_PROXY
+  #HTTP_PROXY: <proxyURL>:<port>
+
```
PrasadG193: @rajinator please move all env vars under `extraEnv` section - `CONFIG_PATH`, `LOG_LEVEL`

rajinator: Are these two considered optional? Or mandatory? If mandatory, makes sense to add another section called `env`, so they don't get ignored

bhavin192: I think we should not move other variables to values.yaml. 
1. `LOG_LEVEL` is already configurable via different variable. 
2. If we want to make `CONFIG_PATH` configurable through values.yaml then we have to take care of volume mount as well. (Does that make sense to have this variable configurable by the user as BotKube binary already runs inside a Pod (container), so the path should not be an issue to someone)

It would be better to keep these variables unexposed to users.

rajinator: Seems it's ready to merge then? @bhavin192

PrasadG193: Ok. For now, we have only `LOG_LEVEL` configurable variable. As suggested, in future, if we have multiple vars, let's keep them in `env` section

PrasadG193:
PR merged, thanks @rajinator !

PR #158, opened 2019-08-17
Enhance Update Events with resource spec diff and Change event message formats

### LINKED ISSUE #131: Send more info with update resource notification
BotKube, when configured to send update events on resources, sends only `Resource updated` message with the notification which doesn't make much sense. Instead, it should also provide information about what has changed with that update

BotKube should provide resource spec diff or about the fields that are updated with resource update notification

### LINKED ISSUE #135: Show Docker Image Tag on Deployments
Hi, this bot looks awesome, it has a lot of great features, good work.

I', trying to find a bot that is able to notify about the upgrades of docker images. 

For example, let's suppose that we are running a k8s Deployment that uses using a Docker image tagged as "myapp:v0.1.0" Then, we upgrade that deployment to a newer image ("myapp:v0.2.0"). I would like to see a bot message on Slack showing that the deployment is now running the new Docker image version ("myapp:v.0.2.0")

I took a look at the screenshots at https://www.botkube.io/examples/ but I don't see that kind of information there. Is it supported right now? Is it in the roadmap?

Thank you very much!

aananthraj:
##### ISSUE TYPE

 - Feature Pull Request

##### SUMMARY

This PR,
- enhances update event notifications with details of what Spec is getting changed
- adds custom `SpecDiffReporter` and `utils.Diff()` method.
- adds update events in all config.yaml files
- excludes namespace field from namespace independent resource events

Fixes #131
#### Preview

   

Fixes #135 
#### Preview

Fixes **excludes namespace field from namespace independent resource events**
#### Preview
- secret being namespaced gets namespace field

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -19,6 +19,7 @@ resources:
       - all
     events:
       - create
+      - update
       - delete
       - error
   - name: statefulset
@@ -26,6 +27,7 @@ resources:
       - all
     events:
       - create
+      - update
       - delete
       - error
   - name: ingress
@@ -89,9 +91,9 @@ resources:
       - all
     events:
       - create
+      - update
       - delete
       - error
-      - update
   - name: role
     namespaces:
       - all
--- a/deploy-all-in-one-tls.yaml
+++ b/deploy-all-in-one-tls.yaml
@@ -29,6 +29,7 @@ data:
           - all
         events:
           - create
+          - update
           - delete
           - error
       - name: statefulset
@@ -36,6 +37,7 @@ data:
           - all
         events:
           - create
+          - update
           - delete
           - error
       - name: ingress
--- a/deploy-all-in-one.yaml
+++ b/deploy-all-in-one.yaml
@@ -29,6 +29,7 @@ data:
           - all
         events:
           - create
+          - update
           - delete
           - error
       - name: statefulset
@@ -36,6 +37,7 @@ data:
           - all
         events:
           - create
+          - update
           - delete
           - error
       - name: ingress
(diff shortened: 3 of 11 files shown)


config.yaml
```diff
@@ -5,119 +5,135 @@ resources:
       - all
     events:                  # List of lifecycle events you want to receive, e.g create, update, delete, error OR all
       - create
+      - update
```
PrasadG193: @aananthraj lets keep `update` events only for `pod` and `job` resources in the default configurations. I don't think people will be interested in update events of all the resources

PrasadG193:
@aananthraj could you please also verify if `JobStatusChecker` filter runs and we get `Job Succeeded` notification on the creation of Job resource (when configured with update event)

config.yaml
aananthraj: Sure lets make them available.
I guess, we need that for deployment as well.

PrasadG193: I would say either deployment or pod. Not both. Otherwise, the user will get redundant notifications

aananthraj: sure, then let's make them default for pods and jobs

aananthraj: resolved it

PrasadG193: Thanks. Just a thought, would it make sense to configure update for `deployment` and `statefulset` instead of `pod`. Because if deployment/statefulset has multiple replica of pod, it will send redundant notifications about each pod updates

aananthraj: In general, `Deployments` and `Statefulset` are deployed in clusters to manage replicas. hence informed it as
> I guess, we need that for deployment as well.

added update event to `Deployments`, `Statefulset` and `Job` by default.

aananthraj:
@PrasadG193, Resolved those issues :+1:

PrasadG193:
Awesome! Great work @aananthraj :tada:

aananthraj:
Thanks for the excellent support @PrasadG193 :+1:

PR #159, opened 2019-08-18
Node Level Critical Events filter

### LINKED ISSUE #134: Node Level Critical Events using filters
Notifications for Node level critical events like NodeNotReady, NodeRebooted and NodeReady can be added.

Provide Red Alert Notifications on NodeRebooted, NodeNotReady, etc. using filters.

aananthraj:
##### ISSUE TYPE

 - Feature Pull Request

##### SUMMARY

This PR,
- Allows  kubernetes events of `Warning` Type as `ErrorEvents`
- Allows kubernetes events of `Normal` Type as Insignificant `InfoEvents`
- Adds `NodeEventsChecker` filter to check for Node InfoEvents with critical reasons like `NodeNotReady` ,`NodeReady`.
- `InfoEvents` with critical reasons are promoted as significant Node `ErrorEvents`
- Post events filtering, unpromoted Insignificant `InfoEvents` are skipped. 

Fixes #134

**Preview**
node-event-filter

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -55,7 +55,8 @@ func RegisterInformers(c *config.Config) {
 	}
 
 	// Register informers for k8s events
-	log.Logger.Infof("Registering kubernetes events informer for types: %+v", config.AllowedEventType.String())
+	log.Logger.Infof("Registering kubernetes events informer for types: %+v", config.WarningEvent.String())
+	log.Logger.Infof("Registering kubernetes events informer for types: %+v", config.NormalEvent.String())
 
 	utils.KubeInformerFactory.Core().V1().Events().Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
 		AddFunc: func(obj interface{}) {
@@ -72,9 +73,13 @@ func RegisterInformers(c *config.Config) {
 			// Kind of involved object
 			kind := strings.ToLower(eventObj.InvolvedObject.Kind)
 
-			// If event type is AllowedEventType and configured for the resource
-			if strings.ToLower(eventObj.Type) == config.AllowedEventType.String() {
+			switch strings.ToLower(eventObj.Type) {
+			case config.WarningEvent.String():
+				// Send WarningEvent as ErrorEvents
 				sendEvent(obj, nil, c, kind, config.ErrorEvent)
+			case config.NormalEvent.String():
+				// Send NormalEvent as Insignificant InfoEvent
+				sendEvent(obj, nil, c, kind, config.InfoEvent)
 			}
 		},
 	})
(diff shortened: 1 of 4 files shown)


aananthraj:
rebased onto develop

pkg/controller/controller.go
```diff
@@ -77,6 +77,25 @@ func RegisterInformers(c *config.Config) {
...
+			eventObj, ok := new.(*apiV1.Event)
+			if !ok {
+				return
+			}
+
+			// Kind of involved object
+			kind := strings.ToLower(eventObj.InvolvedObject.Kind)
+
+			// Send events update notifs if InvolvedObject is node
+			if kind == "node" {
```
PrasadG193: @aananthraj 
1. What type of event.Type we get? `warnings/info`
2. If it's `warnings`, does it make sense to send events for all the resources configured with `error` events instead of hardcoding kind here?

aananthraj: > If it's warnings, does it make sense to send events for all the resources configured with error events instead of hardcoding kind here?

I guess update events on Kubernetes `Events objects` will mostly be repetitve and hence can become noisy

PrasadG193: We get two types of events - `normal` and `warnings`. Out of which we are allowing only `warnings` events (https://github.com/infracloudio/botkube/blob/ff21d65c300b008b895fb3e6bf6fda9cb08935c6/pkg/controller/controller.go#L76). Please check if we get lots of redundant `warning` events, if not then we should send events for those

aananthraj: resolved this issue. @PrasadG193 kindly check.

aananthraj:
rebased onto develop.
I guess it is ready, Let's get this merged.

pkg/filterengine/filters/node_event_checker.go
```diff
@@ -0,0 +1,73 @@
...
+	// Check for Event object
+	eventObj, ok := object.(*apiV1.Event)
+	if !ok {
+		return
+	}
+
+	// Run filter only on Node events
+	if event.Kind != "Node" {
+		// Allow Warning events for all resources
+		if strings.ToLower(eventObj.Type) != config.AllowedEventType.String() {
```
PrasadG193: I don't think this is the right place to have this check. Let's handle it in the following way:
- Add new event type `InfoEvent EventType = "info"` in config.go. Also change `AllowedEventType EventType = ErrorEvent`
- While calling sendEvent() from Event informer handlers (https://github.com/infracloudio/botkube/blob/develop/pkg/controller/controller.go#L60), set `eventType=config.ErrorEvent` for `Warning` events and `eventType=config.InfoEvent` for `Normal` events
- In event.New(), treat InfoEvent same as ErrorEvent
- In this filter, we should just change event type from InfoEvent -> ErrorEvent in the switch case writen below.
- Finally, in controller, after running all filters,  we should add a check to send only ErrorEvent types.
```
if strings.ToLower(eventObj.Type) != config.AllowedEventType.String() {
    return
}
```
This approach will allow us to create filters for sending particular events as Critical events for other resources also

pkg/events/events.go
```diff
@@ -93,7 +93,12 @@ func New(object interface{}, eventType config.EventType, kind string) Event {
 	}
 
 	if kind != "events" {
-		event.Messages = []string{fmt.Sprintf("Resource %sd\n", eventType.String())}
+		switch eventType {
+		case "error":
```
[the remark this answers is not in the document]
aananthraj: resolved it.

pkg/filterengine/filters/node_event_checker.go
aananthraj: resolved requested changes

pkg/events/events.go
```diff
@@ -93,7 +93,12 @@ func New(object interface{}, eventType config.EventType, kind string) Event {
 	}
 
 	if kind != "events" {
-		event.Messages = []string{fmt.Sprintf("Resource %sd\n", eventType.String())}
+		switch eventType {
+		case config.ErrorEvent, config.InfoEvent:
+			event.Messages = []string{fmt.Sprintf("Resource %s\n", eventType.String())}
+		default:
+			event.Messages = []string{fmt.Sprintf("Resource %sd\n", eventType.String())}
+		}
```
PrasadG193: @aananthraj Why do you need switch case here if we are doing the same operation for every case?

aananthraj: Have look at the `%s` and `%sd`
Events like `create` , `update` and `delete` comes with an extra `d` at the end as `Resource Created`
Where as `error`, `info` comes as `Resource Error`

aananthraj: I guess its too small to notice 😅

PrasadG193: Aah, got it

PrasadG193: Probably better to add a comment on that

PrasadG193:
Merged. Thanks @aananthraj !

aananthraj:
@PrasadG193 thanks for the support 👍

Issue #160, opened 2019-08-18
[Kubernetes/Helm] Make cert usage generic


rajinator:
**Is your feature request related to a problem? Please describe.**
Currently custom cert is only supported if the mattermost section is enabled, and a cert is specified for it. If behind a network that uses custom SSL, communications to slack or ELK won't work.

**Describe the solution you'd like**
Make cert support generic, move cert value, and logic out of the mattermost: block.

**Additional context**
I'll be adding an example PR soon

rajinator:
Here are example changes for making cert usage generic. will open a PR after #156 is merged, since this contains those changes https://github.com/opensourcebapu/botkube/commit/765bd87edeb9ac002527a06c543ab7c8c7e764ce

PR #163, opened 2019-08-22
[Rebased] Feature/generic ssl

rajinator:
##### ISSUE TYPE
 - Feature Pull Request

##### SUMMARY
Make custom SSL/TLS support generic, instead of just mattermost. Useful for clusters using self-signed certificates or infrastructure behind SSL proxies in enterprise

Applicable changes for #160 

Rebased according to @bhavin192's request


helm/botkube/values.yaml
```diff
@@ -195,6 +195,9 @@ config:
   # about the best practices for the created resource
   recommendations: true
 
+  ssl:                                          # For using custom SSL certificates
```
bhavin192: I think we should just comment out `ssl` section instead of setting the `ssl.cert` to a string which is "None". We shouldn't have done that before as well, can you please fix it?

helm/botkube/templates/deployment.yaml
```diff
@@ -31,7 +31,7 @@ spec:
           volumeMounts:
             - name: config-volume
               mountPath: "/config"
-          {{- if ne .Values.config.communications.mattermost.cert "None" }}
+          {{- if ne .Values.config.ssl.cert "None" }}
```
bhavin192: 1. With the suggested change in the `values.yaml`, we can replace this with
```suggestion
          {{- if .Values.config.ssl.cert }}
```

helm/botkube/templates/deployment.yaml
```diff
@@ -52,7 +52,7 @@ spec:
         - name: config-volume
           configMap:
             name: {{ include "botkube.fullname" . }}-configmap
-      {{- if ne .Values.config.communications.mattermost.cert "None" }} 
+      {{- if ne .Values.config.ssl.cert "None" }}
```
bhavin192: Same as 1.

helm/botkube/templates/secret.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if ne .Values.config.communications.mattermost.cert "None" }}
+{{- if ne .Values.config.ssl.cert "None" }}
```
bhavin192: Same as 1.

helm/botkube/values.yaml
rajinator: had to leave it uncommented due to [secret.yaml](https://github.com/infracloudio/botkube/blob/develop/helm/botkube/templates/secret.yaml) line 13
```yml
  ca-certificates.crt: {{ .Files.Get (printf "%s" .Values.config.communications.mattermost.cert) | b64enc }}
```
new line in pull is
```yml
  ca-certificates.crt: {{ .Files.Get (printf "%s" .Values.config.ssl.cert) | b64enc }}
```
which would cause errors if the ssl section was commented, without an extra if loop

helm/botkube/templates/deployment.yaml
rajinator: Due to my comment above about the templates/secret.yaml file, used this instead
```suggestion
          {{- if .Values.config.ssl.enabled }}
```

helm/botkube/templates/deployment.yaml
rajinator: See reply to 1.

helm/botkube/templates/secret.yaml
rajinator: See reply to 1.

bhavin192:
@mugdha-adhav can you please take a look at this changes once, just to make sure everything is fine.

PrasadG193:
Thanks for the PR @rajinator !

PR #169, opened 2019-08-30
Add support for Webhooks

aananthraj:
##### ISSUE TYPE

 - Feature Pull Request

##### SUMMARY

**Modify return type as pointer:**

This commit,
- modifies return type of  GetLastSeenSlackMessage to pointers.

**Add support for Webhooks:**

This Commit,
- Adds support for event notifications through `Wehbook` 
- updates configs in all yaml files
- updated helm charts
- adds unit tests for webhook notifier
- adds integration test webhook notifier
- adds test/webhook package to enable integration testing.

**Preview**:

```
{
  "meta": {
    "kind": "Pod",
    "name": "command-app-1",
    "namespace": "default",
    "cluster": "demo"
  },
  "status": {
    "type": "create",
    "level": "info",
    "reason": "",
    "messages": [
      "Resource created\n"
    ]
  },
  "summary": "Pod `command-app-1` in of cluster `demo`, namespace `default` has been created:\n```Resource created\nRecommendations:\n- :latest tag used in image 'debian' of Container 'command-demo-container' should be avoided.\n```",
  "timestamp": "2019-08-30T22:29:04+05:30",
  "recommendations": [
    ":latest tag used in image 'debian' of Container 'command-demo-container' should be avoided.\n"
  ]
}
```

### DIFF
--- a/test/e2e/env/env.go
+++ b/test/e2e/env/env.go
@@ -7,6 +7,8 @@ import (
 	"testing"
 
 	"github.com/infracloudio/botkube/pkg/config"
+	"github.com/infracloudio/botkube/test/e2e/utils"
+	"github.com/infracloudio/botkube/test/webhook"
 	"github.com/nlopes/slack"
 	"github.com/nlopes/slack/slacktest"
 	"k8s.io/client-go/kubernetes"
@@ -21,6 +23,7 @@ import (
 type TestEnv struct {
 	K8sClient     kubernetes.Interface
 	SlackServer   *slacktest.Server
+	WebhookServer *webhook.Server
 	SlackMessages chan (*slack.MessageEvent)
 	Config        *config.Config
 }
@@ -45,6 +48,9 @@ func New() *TestEnv {
 	testEnv.Config.Communications.Slack.Token = "ABCDEFG"
 	testEnv.Config.Communications.Slack.Channel = "cloud-alerts"
 
+	//Set Webhook
+	testEnv.Config.Communications.Webhook.Enabled = true
+
 	// Add settings
 	testEnv.Config.Settings.ClusterName = "test-cluster-1"
 	testEnv.Config.Settings.AllowKubectl = true
@@ -55,6 +61,7 @@ func New() *TestEnv {
 	testEnv.K8sClient = fake.NewSimpleClientset()
 	testEnv.SlackMessages = make(chan (*slack.MessageEvent), 1)
 	testEnv.SetupFakeSlack()
+	testEnv.SetupFakeWebhook()
 
 	return testEnv
 }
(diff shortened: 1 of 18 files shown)


sanketsudake:
Lets add unit tests and integration tests as well. We can mock webhook URL and notifier would call that webhook URL with payload. We can verify cross check payload received on Webhook and sent from notifier.

You can also refer integration tests recently added by @PrasadG193

aananthraj:
@sanketsudake, @PrasadG193 made those requested changes.

test/e2e/notifier/create/create.go
```diff
@@ -55,7 +66,14 @@ func (c *context) testCreateResource(t *testing.T) {
 			err := json.Unmarshal([]byte(lastSeenMsg), &m)
 			assert.NoError(t, err, "message should decode properly")
 			assert.Equal(t, c.Config.Communications.Slack.Channel, m.Channel)
-			assert.Equal(t, test.Expected.Attachments, m.Attachments)
+			assert.Equal(t, test.ExpectedSlackMessage.Attachments, m.Attachments)
+
+			// Get last seen webhook payload
+			lastSeenPayload := c.GetLastReceivedPayload()
```
sanketsudake: This may fail if webhook is not called by the time this function call is made. Should be have retries here ?

test/e2e/env/env.go
```diff
@@ -76,3 +83,20 @@ func (e TestEnv) GetLastSeenSlackMessage() string {
...
+	e.WebhookServer = s
+}
+
+// GetLastReceivedPayload return last message received by fake webhook server
+func (e TestEnv) GetLastReceivedPayload() utils.WebhookPayload {
+	allSeenMessages := e.WebhookServer.GetReceivedPayloads()
+	if len(allSeenMessages) != 0 {
+		return allSeenMessages[len(allSeenMessages)-1]
+	}
+	return utils.WebhookPayload{}
```
sanketsudake: If no message is there it should `nil` I guess instead of empty payload. Please let me know your opinion as well on this.

test/e2e/notifier/create/create.go
aananthraj: I guess the delay introduced for slack and time for the assertion of  slack notifier would be sufficient for the mock server receiving data. your thought.?
```
// Get last seen slack message
time.Sleep(time.Second)
```

sanketsudake: Ok. Looks good for now.

test/e2e/env/env.go
aananthraj: > If no message is there it should nil

as it is returning the object struct itself, it is made to return the Zero Value of the struct(empty payload). for returning nil we might need to change it as a pointer.

PrasadG193:
Minor comment, can we rename `test/webhooktest` package to `test/webhook` to maintain consistency in the naming format of all `test/*` packages

aananthraj:
Sure, lets maintain the same.

test/e2e/env/env.go
aananthraj: it is good to use pointers instead of passing the object directly.
shall we change it..? or lets leave this as is for now..? 
@PrasadG193 thoughts..?

PrasadG193: @aananthraj let's use pointers if that makes sense. We might need to modify return type of `GetLastSeenSlackMessage` also have consistency

aananthraj:
made those request changes. 
It is ready now, let's get it merged.

aananthraj:
made those request changes.
It is ready now, let's get it merged.

PrasadG193:
Looks good for now, we might need to add support for the self-signed certificates in future

PR #187, opened 2019-10-08
Fix duplicate notification for Job update

PrasadG193:
##### ISSUE TYPE
 - Bug fix Pull Request

##### SUMMARY

This PR fixes the issue of getting duplicate Job update notifications.
We'll be ignoring update events which are older than 30sec

### DIFF
--- a/pkg/filterengine/filters/job_status_checker.go
+++ b/pkg/filterengine/filters/job_status_checker.go
@@ -4,6 +4,8 @@
 package filters
 
 import (
+	"time"
+
 	"github.com/infracloudio/botkube/pkg/config"
 	"github.com/infracloudio/botkube/pkg/events"
 	"github.com/infracloudio/botkube/pkg/filterengine"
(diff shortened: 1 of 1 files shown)


pkg/controller/controller.go
```diff
@@ -173,8 +173,16 @@ func sendEvent(obj, oldObj interface{}, c *config.Config, notifiers []notify.Not
 		}
 	}
 
-	// Filter events
+	// Run custom filters
 	event = filterengine.DefaultFilterEngine.Run(obj, event)
+
+	// Filter events
+	// Skip older update events if timestamp difference is more than 30sec
+	if !event.TimeStamp.IsZero() && time.Now().Sub(event.TimeStamp) > 30*time.Second {
```
codenio: This code is to be executed only for job update events..?
I guess we need to add the above specific conditon to avoid skipping of other events.

PrasadG193: This is not Job resources specific, this will be checked for each event

PrasadG193: @codenio We are comparing TimeStamp with time.Now(). Any live event won't have time difference more than few Microseconds. We just don't want to send a notification with having an older timestamp. We are not figuring out whether it is duplicate or not here

codenio: @PrasadG193 thanks making suitable changes.

PrasadG193:
Thanks @codenio for reviewing and suggesting changes.

PR #197, opened 2019-10-11
Fix helm/botkube/values.yaml clusterrole namespace info

### LINKED ISSUE #194: [ERROR] unmarshal error from configmap.yaml
There is an error from configmap.yaml, line number 184
after 'namespaces', neither include nor ignore isn't appear.
Maybe it is changed like below
---------------------------------------
namesapces:
  include:
    -all
  ignore:
    -

daegeun-ha:
##### ISSUE TYPE
 - Bug fix Pull Request

##### SUMMARY
Fixes #194 
helm/botkube/values.yaml file fix
missing 'include', 'ignore' keyword, so fix it.

### DIFF
--- a/helm/botkube/values.yaml
+++ b/helm/botkube/values.yaml
@@ -181,7 +181,10 @@ config:
         - error
     - name: clusterrole
       namespaces:
-        - all
+        include:
+          - all
+        ignore:
+          - 
       events:
         - create
         - delete


sanketsudake:
@PrasadG193  I wonder we couldn't catch this in earlier testing. We are missing checks for invalid configuration ? How does this configuration gets parsed if it's invalid ?

PrasadG193:
@sanketsudake While Unmarshal, Golang skips the field if it is not matching the struct. Also, for testing we use config from https://github.com/infracloudio/botkube/blob/develop/config.yaml which doesn't have the field. I'll add this in e2e test

PrasadG193:
@daegeun-ha Could you please also update 
- https://github.com/infracloudio/botkube/blob/develop/deploy-all-in-one-tls.yaml, 
- https://github.com/infracloudio/botkube/blob/develop/deploy-all-in-one.yaml and 
- https://github.com/infracloudio/botkube/blob/develop/config.yaml 
so that they are in sync

PrasadG193:
@daegeun-ha could you please also remove
 
```
  - name: clusterrolebindings
    namespaces:
      include:  
        - all
      ignore:
        - 
```
Redundant field from https://github.com/infracloudio/botkube/blob/develop/config.yaml

PrasadG193:
@daegeun-ha I have cherry-picked your changes and raised separate PR.
Thanks for catching and fixing this!

PR #213, opened 2019-11-12
Remove errant } from clusterrole helm template

baronomasia:
##### ISSUE TYPE
 - Bug fix Pull Request

##### SUMMARY
This removes an errant } from the helm clusterrole template


PrasadG193:
Aah, I wonder why `helm lint` didn't catch that :thinking:

PR #219, opened 2019-11-20
Expose prometheus metrics for BotKube

Surbhidongaonkar:
##### ISSUE TYPE
 - Feature Pull Request

##### SUMMARY
Adding prometheus which exposes different metrics from botkube like no of resources scanned,  no of goroutines that currently exist etc.

Fixes #https://github.com/infracloudio/botkube/issues/182

Dependencies: github.com/prometheus/client_golang/prometheus/promhttp


helm/botkube/templates/service.yaml
```diff
@@ -0,0 +1,26 @@
+{{- if .Values.service }}
+{{- range $port := .Values.service.ports  }}
+{{- $service_type := $port.type | default "ClusterIP" }}
```
PrasadG193: @Surbhidongaonkar We shouldn't be creating service if serviceMonitor.enabled is not set true

Surbhidongaonkar: Have modified service.yaml with required changes

helm/botkube/templates/servicemonitor.yaml
```diff
@@ -0,0 +1,31 @@
+{{- if .Values.serviceMonitor.enabled }}
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  name: {{ template "botkube.fullname" . }}
+  labels:
```
PrasadG193: @Surbhidongaonkar We should take these labels from `Values.serviceMonitor.labels`

bhavin192: @PrasadG193 small correction here: we should append the labels from `Values.serviceMonitor.labels` to these labels

helm/botkube/values.yaml
```diff
@@ -250,6 +250,21 @@ config:
...
     upgradeNotifier: true
 
+service:
+  name: metrics
+  port: 2112
+  targetPort: 2112  
+            
+serviceMonitor:
+  ## If true, a ServiceMonitor CRD is created for a botkube
+  ## https://github.com/infracloudio/botkube
```
bhavin192: We should write `CR` instead of `CRD` here. 
Also how about adding link to Pormetheus Operator GitHub instead of BotKube repository?
May be this? https://github.com/coreos/prometheus-operator/blob/master/Documentation/api.md#servicemonitor

cmd/botkube/main.go
```diff
@@ -30,6 +30,9 @@ func main() {
 		go mb.Start()
 	}
 
+	//Prometheus metrics
```
bhavin192: Small typo here: space after `//`

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,12 @@
+package metrics
+
+import (
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+	"net/http"
+)
+
+// ServeMetrics expose metrics
```
bhavin192: How about writing something like this
```go
// ServeMetrics exposes metrics in Prometheus format
```

Surbhidongaonkar: Thanks! I have made the required changes.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,12 @@
...
+
+import (
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+	"net/http"
+)
+
+// ServeMetrics exposes metrics in Prometheus format
+func ServeMetrics() {
+	http.Handle("/metrics", promhttp.Handler())
+	http.ListenAndServe(":2112", nil)
```
PrasadG193: @Surbhidongaonkar We should read port from env var. Something like `METRICS_PORT`. And pass that env variable through deployment spec (like https://github.com/infracloudio/botkube/blob/develop/helm/botkube/templates/deployment.yaml#L41)

PR #238, opened 2019-12-27
Added restrictAccess flag to enable/disable behavior of botkube. 

mahendrabagul:
This flag will control user's access to botkube via configured channel. If set to true, it will allow user to talk to botkube only through configured channel. In other case, botkube will be accessible from all channels.

Signed-off-by: Mahendra Bagul <mahendra@infracloud.com>

##### ISSUE TYPE

 - Bug fix Pull Request

##### SUMMARY

Fix for https://github.com/infracloudio/botkube/issues/235

### DIFF
--- a/pkg/execute/executor.go
+++ b/pkg/execute/executor.go
@@ -67,11 +67,12 @@ type Executor interface {
 
 // DefaultExecutor is a default implementations of Executor
 type DefaultExecutor struct {
-	Message       string
-	AllowKubectl  bool
-	ClusterName   string
-	ChannelName   string
-	IsAuthChannel bool
+	Message        string
+	AllowKubectl   bool
+	RestrictAccess bool
+	ClusterName    string
+	ChannelName    string
+	IsAuthChannel  bool
 }
 
 // CommandRunner is an interface to run bash commands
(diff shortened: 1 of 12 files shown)


pkg/execute/executor.go
```diff
@@ -53,6 +53,7 @@ const (
 	notifierStartMsg   = "Brace yourselves, notifications are coming from cluster '%s'."
 	notifierStopMsg    = "Sure! I won't send you notifications from cluster '%s' anymore."
 	unsupportedCmdMsg  = "Command not supported. Please run /botkubehelp to see supported commands."
+	restrictAccessMsg  = "Sorry, You can use botkube from configured channel only."
```
PrasadG193:
```suggestion
	restrictAccessMsg  = "Sorry, You can use BotKube from configured channel only."
```

pkg/execute/executor.go
```diff
@@ -217,9 +225,6 @@ func runKubectlCommand(args []string, clusterName string, isAuthChannel bool) st
 		}
 		finalArgs = append(finalArgs, arg)
 	}
-	if isAuthChannel == false {
-		return ""
-	}
```
PrasadG193: Please verify if `BotKube notifier` and `BotKube filters` commands are working only from configured channel regardless of `restrictAccess` flag

mahendrabagul: I tested the flow and its working now.

pkg/execute/executor.go
mahendrabagul: This statement is not required now, so removed the message.

PrasadG193:
Merged! Thanks @mahendrabagul for the PR

PR #243, opened 2020-02-05
Checked cluster name before executing kubectl command

girishg4t:
This commit fix the bug 230 which shows redundant error message
when deployed in miticluster, added check for cluster name and
created new function to get cluster name from kubectl command

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -3,6 +3,7 @@ package utils
 import (
 	"fmt"
 	"os"
+	"regexp"
 	"strconv"
 	"time"
 
@@ -401,4 +402,16 @@ func ExtractAnnotaions(obj *coreV1.Event) map[string]string {
 	}
 
 	return map[string]string{}
+}
+
+//GetClusterNameFromKubectlCmd this will return cluster name from kubectl command
+func GetClusterNameFromKubectlCmd(cmd string) string {
+	r, _ := regexp.Compile(`--cluster-name[=|' ']([^\s]*)`)
+	//this gives 2 match with cluster name and without
+	matchedArray := r.FindStringSubmatch(cmd)
+	var s string = ""
+	if len(matchedArray) >= 2 {
+		s = matchedArray[1]
+	}
+	return s
 }
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -409,7 +409,7 @@ func GetClusterNameFromKubectlCmd(cmd string) string {
 	r, _ := regexp.Compile(`--cluster-name[=|' ']([^\s]*)`)
 	//this gives 2 match with cluster name and without
 	matchedArray := r.FindStringSubmatch(cmd)
-	var s string = ""
+	var s string
 	if len(matchedArray) >= 2 {
 		s = matchedArray[1]
 	}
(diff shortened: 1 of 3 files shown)


pkg/utils/utils_test.go
```diff
@@ -0,0 +1,31 @@
+package utils
+
+import "testing"
+
+func TestGetClusterNameFromKubectlCmd(t *testing.T) {
+	str := "get pods --cluster-name=minikube"
+	result := GetClusterNameFromKubectlCmd(str)
+
+	if result != "minikube" {
+		t.Errorf("Expecte minikube but got %v ", result)
```
sanketsudake: `expected` instead of `Expecte`

pkg/utils/utils_test.go
```diff
@@ -0,0 +1,31 @@
+package utils
+
+import "testing"
+
+func TestGetClusterNameFromKubectlCmd(t *testing.T) {
```
sanketsudake: I see we have pattern here. Can we please write this as table drive test ?

pkg/utils/utils.go
```diff
@@ -402,3 +403,14 @@ func ExtractAnnotaions(obj *coreV1.Event) map[string]string {
 
 	return map[string]string{}
 }
+
+//GetClusterNameFromKubectlCmd this will return cluster name from kubectl command
+func GetClusterNameFromKubectlCmd(cmd string) string {
+	i := strings.Index(cmd, "--cluster-name") + 15 //15 indicates 15 letters in --cluster-name string with space or equal
```
sanketsudake: Please see if we could use valid regex instead.

pkg/utils/utils_test.go
sanketsudake: Also ordering of string update and call for `GetClusterNameFromKubectlCmd` function is not correct.

pkg/utils/utils.go
PrasadG193: Not as a part of this PR but we should use `cli` parser (like github.com/urfave/cli) to parse BotKube commands.

pkg/utils/utils_test.go
girishg4t: corrected

pkg/utils/utils.go
girishg4t: added regex

pkg/utils/utils_test.go
girishg4t: done

pkg/utils/utils_test.go
```diff
@@ -0,0 +1,31 @@
...
+
+	type test struct {
+		input    string
+		expected string
+	}
+
+	tests := []test{
+		{input: "get pods --cluster-name=minikube", expected: "minikube"},
+		{input: "--cluster-name minikube1", expected: "minikube1"},
```
sanketsudake: Can we try `--cluster-nameminikube1` ?

girishg4t: yes added and passed 
{input: "--cluster-nameminikube1", expected: ""},

pkg/utils/utils.go
```diff
@@ -402,3 +403,15 @@ func ExtractAnnotaions(obj *coreV1.Event) map[string]string {
 
 	return map[string]string{}
 }
+
+//GetClusterNameFromKubectlCmd this will return cluster name from kubectl command
+func GetClusterNameFromKubectlCmd(cmd string) string {
+	r, _ := regexp.Compile(`--cluster-name[=|' ']([^\s]*)`)
+	//this gives 2 match with cluster name and without
+	matchedArray := r.FindStringSubmatch(cmd)
+	var s string = ""
```
PrasadG193: Minor: 
```suggestion
	var s string
```

girishg4t: done

PR #255, opened 2020-04-15
Adding ability to make rbac rules and service account configurable

adusumillipraveen:
##### ISSUE TYPE

 - Feature Pull Request

##### SUMMARY

This will enhance helm chart giving ability for users to manage RBAC rules or use a different service account . This is to avoid botkube slack to have access to everything including secrets on slack .

The chart changes are non breaking and behave like older version with defaults set as before.


helm/botkube/values.yaml
```diff
@@ -307,3 +307,16 @@ nodeSelector: {}
...
+  rules:
+    - apiGroups: ["*"]
+      resources: ["*"]
+      verbs: ["get", "watch", "list"]
+serviceAccount:
+  # Specifies whether a service account should be created
+  create: true
+  # The name of the service account to use.
+  # If not set and create is true, a name is generated using the fullname template
+  name: test
```
timja, line 322: shouldn't this be not set by default so fullname is used?

adusumillipraveen: Thanks, was just added to **test** in local :)

helm/botkube/values.yaml
```diff
@@ -307,3 +307,16 @@ nodeSelector: {}
...
+  rules:
+    - apiGroups: ["*"]
+      resources: ["*"]
+      verbs: ["get", "watch", "list"]
+serviceAccount:
+  # Specifies whether a service account should be created
+  create: true
+  # The name of the service account to use.
+  # If not set and create is true, a name is generated using the fullname template
+  name:
```
timja, line 322:
```suggestion
  #name:

```

helm/botkube/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v1
 name: botkube
 home: https://www.botkube.io
-version: v0.9.1
-appVersion: v0.9.1
+version: v0.9.2
+appVersion: v0.9.2
```
PrasadG193, line 5: The chart version will be updated by release scripts when we publish the release
```suggestion
version: v0.9.1
appVersion: v0.9.1
```

PrasadG193:
Thanks @adusumillipraveen for the PR and @timja for reviewing it. 
As @timja suggested, @adusumillipraveen - could you please update the default rbac rules to exclude secrets?

adusumillipraveen:
> Thanks @adusumillipraveen for the PR and @timja for reviewing it.
> As @timja suggested, @adusumillipraveen - could you please update the default rbac rules to exclude secrets?

@PrasadG193 It is not easy to do that as it doesn't support [set subtraction](https://github.com/kubernetes/kubernetes/issues/70387) . So, we will need to list everything and it can become quite big. I would leave it to users to maintain it ?

PrasadG193:
That's true. We already have an issue created to track this. Let's handle that separately

helm/botkube/templates/_helpers.tpl
```diff
@@ -30,3 +30,14 @@ Create chart name and version as used by the chart label.
...
+{{/*
+Create the name of the service account to use
+*/}}
+{{- define "botkube.serviceAccountName" -}}
+{{- if .Values.serviceAccount.create -}}
+    {{ include "botkube.fullname" . }}-sa
+{{- else -}}
+    {{ default "default" .Values.serviceAccount.name }}
+{{- end -}}
+{{- end -}}
```
PrasadG193, line 43: [minor]
```suggestion
{{- end -}}

```

adusumillipraveen:
Thanks @PrasadG193 . Are there any plans to release master anytime soon ? As the older images (v0.9.1) doesn't seem to be compatible with latest chart.  Don't really want use latest tag.

PrasadG193:
@adusumillipraveen We are trying to get Teams support in the next release. But it is being harder than we expected :). If we are not able to do it before next week, we'll probably do a minor release

Issue #258, opened 2020-04-20
nodeSelector is not applied


eddycharly:
In the helm deployment, the nodeSelector configuration is not applied.

eddycharly:
Fixed by #259

Issue #257, opened 2020-04-19
Don't use latest in helm image tag


eddycharly:
https://github.com/infracloudio/botkube/blob/16e90962d65b27d874f0a5129e9f42e25d793aeb/helm/botkube/values.yaml#L12

Having latest in helm image tag is dangerous, it breaks with current latest image because of a change in the config yaml file.

PrasadG193:
Hey @eddycharly,
`develop` branch contains latest changes under development. So one is not expected to deploy BotKube from `develop` unless they want to experiment and test changes under development. Ideally, as described in docs (https://www.botkube.io/installation/), one should refer latest release or master branch for development. The helm charts and deploy manifest we have in `develop` are compatible with latest changes i.e ":latest" branch only. Once we do the release, we tag helm charts and images with the stable tag.

eddycharly:
Hi @PrasadG193 , that's not true, the helm chart v0.9.1 is using `latest` (see https://github.com/infracloudio/botkube/releases/download/v0.9.1/helm-chart-botkube-v0.9.1.tgz), so deploying the chart doesn't work without changing the values.

PrasadG193:
I see. If you are talking about default values, then you are right. We should update the release script to overwrite default image tag as well. Thanks

eddycharly:
Fixed by #259

PR #259, opened 2020-04-20
helm fixes

eddycharly:
##### ISSUE TYPE
 - Bug fix Pull Request

##### SUMMARY
Fixing #257 and #258.

### DIFF
--- a/hack/release.sh
+++ b/hack/release.sh
@@ -29,6 +29,7 @@ update_chart_yamls() {
 
     sed -i "s/version.*/version: ${version}/" helm/botkube/Chart.yaml
     sed -i "s/appVersion.*/appVersion: ${version}/" helm/botkube/Chart.yaml
+    sed -i "s/\btag:.*/tag: ${version}/" helm/botkube/values.yaml
     sed -i "s/\bimage: \"infracloudio\/botkube.*\b/image: \"infracloudio\/botkube:${version}/g" deploy-all-in-one.yaml
     sed -i "s/\bimage: \"infracloudio\/botkube.*\b/image: \"infracloudio\/botkube:${version}/g" deploy-all-in-one-tls.yaml
 
--- a/helm/botkube/templates/deployment.yaml
+++ b/helm/botkube/templates/deployment.yaml
@@ -68,4 +68,7 @@ spec:
         runAsUser: {{ .Values.securityContext.runAsUser }}
         runAsGroup: {{ .Values.securityContext.runAsGroup }}
       {{ end }}
-
+      {{- with .Values.nodeSelector }}
+      nodeSelector:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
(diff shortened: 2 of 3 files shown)


PrasadG193:
Looks good. @eddycharly Could you please also update release script to overwrite the default image tag in values.yaml?
https://github.com/infracloudio/botkube/blob/develop/hack/release.sh#L27

eddycharly:
How do you want to get this done ?

Here the chart version should be bumped but there’s no need to change the image version. 

I mena the chart can evolve without changing the image (nodeSelector for exemple is only a chart concern).

PrasadG193:
We have address the release script change in separate PR. We just need to add one more `sed` statement (like https://github.com/infracloudio/botkube/blob/develop/hack/release.sh#L31) to update the `tag: ` in value.yaml

eddycharly:
Let's say we create a `v0.9.2` release, the version is extracted form the version tag.

How can we set the chart version to `v0.9.2` and image tag to `v0.9.1` ?
I don't see a way to decouple chart version and image tag.

PrasadG193:
@eddycharly We do application and chart release at the same time with the same tag.

eddycharly:
@PrasadG193 i made the requested changes in `release.sh`

hack/release.sh
```diff
@@ -29,6 +29,8 @@ update_chart_yamls() {
 
     sed -i "s/version.*/version: ${version}/" helm/botkube/Chart.yaml
     sed -i "s/appVersion.*/appVersion: ${version}/" helm/botkube/Chart.yaml
+    sed -i "s/\btag:.*/tag: ${version}/" helm/botkube/values.yaml
+    sed -i "s/\bpullPolicy:.*/pullPolicy: IfNotPresent/" helm/botkube/values.yaml
```
PrasadG193, line 33: No need for this statement since we have already set `pullPolicy: IfNotPresent` and that is not going to change even after releases

eddycharly: I thought that with `latest` it should be `Always`.
Not sure if the values.yaml file should contain `tag: latest` and `pullPolicy: Always` in `develop`.

PrasadG193: If image tag is set to latest, the image will always be pulled irrespective of what image policy is set. Ref: https://kubernetes.io/docs/concepts/containers/images/#updating-images. So we can just set pullPolicy to `IfNotPresent` in default helm values

eddycharly: You're right, I forgot about the special handling of `latest` in k8s.
Done

eddycharly:
@PrasadG193 any chance to get a release soon ?

PrasadG193:
@eddycharly We are working on adding/updating documentation. We should be able to release by this weekend

eddycharly:
Thanks 👍

eddycharly:
@PrasadG193 If it helps, you can use [helm-docs](https://github.com/norwoodj/helm-docs) to generate helm chart doc automatically.

I can open a PR for that if you're interested.

PrasadG193:
@eddycharly That looks great. Let me see if I can automate the doc generation. BTW, we maintain documentation in a separate repo here: https://github.com/infracloudio/botkube-docs

PR #266, opened 2020-04-26
Add license headers to source files

### LINKED ISSUE #225: [Refactor] Add copyright license headers in all source code files
We can use the following utility to scan and add required license headers in all files
https://github.com/google/addlicense

PrasadG193:
##### SUMMARY

- Add license headers to source files
- Update copyright years t0 2019-2020 in license

Fixes #225

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -1,3 +1,22 @@
+# Copyright (c) 2019 InfraCloud Technologies
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy of
+# this software and associated documentation files (the "Software"), to deal in
+# the Software without restriction, including without limitation the rights to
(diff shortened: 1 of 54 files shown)


build/Dockerfile
```diff
@@ -1,3 +1,22 @@
+# Copyright (c) 2019 InfraCloud Technologies
```
sanketsudake, line 1: Can you please update the year 2019-2020 in source files as well ?

PrasadG193: @sanketsudake year in the source files license headers should point to the year it was added. e.g https://github.com/infracloudio/botkube/pull/266/files#diff-c30adc0ccd728f66fd5cc8b794cab804R1. That's we have observed in other OSS projects as well
cc: @bhavin192

PR #278, opened 2020-05-05
Fix support for tolerations

jspiro:
##### ISSUE TYPE
 - Bug fix Pull Request

##### SUMMARY

Currently, even if I set toleration values in the `values.yaml` file they are not being picked up as there is not toleration block in  `deployment.yaml` template

### DIFF
--- a/cmd/botkube/main.go
+++ b/cmd/botkube/main.go
@@ -38,20 +38,20 @@ const (
 
 func main() {
 	log.Logger.Info("Starting controller")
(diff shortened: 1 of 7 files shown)


PrasadG193:
Hey @jspiro ,
This is already fixed with https://github.com/infracloudio/botkube/pull/277. Please let me know if you are still facing the issue

jspiro:
Ah so it is!

PR #284, opened 2020-05-30
Make allowed kubectl commands configurable

### LINKED ISSUE #183: Limit kubectl commands
I would like to suggest adding option to limit specific kubectl commands.

For example - I would like to disable `describe secret` command, because it could reveal some sensitive data.

girishg4t:
##### ISSUE TYPE

 - Feature Pull Request

##### SUMMARY
Added configuration in resource_config.yaml for limiting kubectl commands as per the discussion given in ticket 183

The configuration added is used in 
slack.go and mattermost.go to pass it to executor.go and used to limit the commands execution

Fixes #183


PrasadG193:
@girishg4t could you please resolve the errors in CI build?

girishg4t:
@PrasadG193 fixed the issue in test case

PrasadG193:
@girishg4t We should also allow if someone uses singular resource name - like `@BotKube get pod`

girishg4t:
@PrasadG193 i have made the changes as per review comments please check

PrasadG193:
@girishg4t This is yet to be resolved https://github.com/infracloudio/botkube/pull/284#issuecomment-637572333

girishg4t:
> @girishg4t This is yet to be resolved [#284 (comment)](https://github.com/infracloudio/botkube/pull/284#issuecomment-637572333)

@PrasadG193 is this not the one ?
https://github.com/infracloudio/botkube/pull/284/commits/af7dd4e62fd6a5458e414ec50d7a43916de9a3ec

PrasadG193:
If user configured access with - 
```
    commands:
      # method which are allowed
      verbs: ["get"]
      # resource configuration which is allowed
      resources: ["deployments"]
```
Will they be able to execute `@BotKube get deploy` or `@BotKube get deployment`. I don't think so.
I think what we should do is, we should execute `kubectl api-resources` (which include singular name as well as short name for the resources https://godoc.org/k8s.io/apimachinery/pkg/apis/meta/v1#APIResource) during init() and store the response. Then for every user's command, get resources name from shortname if passed from that list. We can use Discovery APIs to get the resource list - https://godoc.org/k8s.io/client-go/discovery#ServerGroupsAndResources

PrasadG193:
@girishg4t [minor] ideally PR heading should be one-liner summary of what you have done. It should inform reader what is the PR about. Please follow this while creating PR next time :slightly_smiling_face:

PrasadG193:
@girishg4t I've done a few changes as per the discussion. Going to merge once CI passes. PTAL

Issue #262, opened 2020-04-21
Refactor logging package


PrasadG193:
Refactor logging package to make log statement simpler. Current format `logging.Logger.Info()` doesn't look good. We can embed Logrus into `pkg/log` to import and use it as `log.Info()`

PrasadG193:
Fixed with https://github.com/infracloudio/botkube/pull/285

PR #285, opened 2020-06-11
Refactor logging package (#262)

hmharsh:
##### ISSUE TYPE

 - Refactor Pull Request  (#262)

##### SUMMARY
Refactor logging package to make log statements simpler. Current format logging.Logger.Info() doesn't look good. We can embed Logrus into pkg/log to import and use it as log.Info()

Fixes #
- Embed Logrus into pkg/log 
- Update the reference to new log package
- Change the reference everywhere form 
    `logging.Logger.Function() --> log.Function()`


PrasadG193:
@hmharsh Let's do following things:
- Remove logging package (as we are not doing much there) move init() function to new `log` pkg
- Create global log object `var log = logrus.New()` in log package
- Use `log` object in wrapper functions

```
var log = logrus.New()

func Info(message ...interface{}) {
	log.Info(message...)
}
```
What do you think?

hmharsh:
That's a brilliant idea, I will be updating the code accordingly.

PR #288, opened 2020-06-16
Create BotKube POC for Namespace.md

hmharsh:
##### ISSUE TYPE
Proposal for the namespace to communication channel configuration input


design/BotKube POC for Namespace.md
```diff
@@ -0,0 +1,144 @@
...
+          verbs: ["get", "logs", "describe", "diff", "explain", "top"]
+          # resource configuration which is allowed
+          resources: ["Deployments", "Pods", "Services", "Nodes", "Ingresses", "Roles"]  
+          # enable notification about specific resources
+          configwatcher: true 
+          
+
+
+  # Profile: BOTKUBE_ADMIN 
+  - name: 'BOTKUBE_ADMIN'
```
sanketsudake, line 120: Maybe Profile should define only define access permisisons and there should be binding between channel and profile.
Once channel can have only one profile.

```
profile:
    namespaces: []
     kubectl: 
          enabled: true/false
          commands:
                verbs: []
                resources: []

communitcations:
          slack:
               enabled: true/false
               token: 'xxxx'
                notiftype: short
                accessbindings:
                    - name: developers
                       profile: dev_profile
                     - name: admin
                        profile: admin_profile
```

design/BotKube POC for Namespace.md
```diff
@@ -0,0 +1,144 @@
...
+      # - ns2
+      kubectl:
+        enabled: true
+        commands:
+          # method which are allowed
+          verbs: ["get", "logs", "describe" ,"api-resources", "api-versions", "cluster-info", "diff", "explain", "top", "auth"]
+          # resource configuration which is allowed
+          resources: [ "Pods", "Services","Namespaces", "Nodes","ReplicationControllers", "PersistentVolumes", "PersistentVolumeClaims", "Secrets", "ConfigMaps", "Deployments", "DaemonSets", "ReplicaSets", "StatefulSets", "Ingresses", "Jobs", "Roles", "RoleBindings", "ClusterRoles"]  
+          # enable notification about specific resources
+          configwatcher: true
```
sanketsudake, line 135: Wonder why configWatcher would be required and profile level ? It should be at comms level ?

design/BotKube POC for Namespace.md
hmharsh: yes, that's even much better idea also by these different channels of the different communication platforms be bonded with the same profile eg.
dev channel of slack and development channel of mattermost can bound to one single profile

design/BotKube POC for Namespace.md
hmharsh: yes, I was also a bit confused to include it here or not but after thinking about some real-life scenario like a channel (say developer) having very specific and limited access should or should not be informed about the same specific resources as described in the profile. 
just as an additional option we can keep it as it is also.

design/BotKube POC for Namespace.md
hmharsh: since code changes will be generic that same pattern can be used with webhook and elastic search, by default they might bound to admin profile but if required multiple webhook, elastic search URL can be configured with different levels of access.

PrasadG193:
This is getting addressed as a part of https://github.com/infracloudio/botkube/pull/291. Hence closing the PR

PR #301, opened 2020-06-30
Allow kubectl commands wihtout namespace and without cluster-name

### LINKED ISSUE #300: Executor fails to run when cluster-name is ignored [BUG]
**Describe the bug**
Executor fails to run when namespace is ignored. For eg : ```get pods``` . While ```get pods -n micro ``` works.
 
**To Reproduce**
1. Go to any channel (slack or telegram)
2. Issue a kubectl command with single param eg " ```get pods```
3. Bot does not respond or returns empty results

**Expected behavior**
Bot should respond with a list of pods.
**Screenshots**

**Additional context**

gmkumar2005:
##### ISSUE TYPE
     - Bug fix Pull Request

    ##### SUMMARY
    Default executor is expecting more than 2 arguments.
    Fixed number of arguments needed to 1.

Fixes #300

### DIFF
--- a/pkg/execute/executor.go
+++ b/pkg/execute/executor.go
@@ -159,7 +159,7 @@ func NewDefaultExecutor(msg string, allowkubectl, restrictAccess bool, defaultNa
 func (e *DefaultExecutor) Execute() string {
 	args := strings.Fields(e.Message)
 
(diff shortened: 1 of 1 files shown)


PrasadG193:
Thanks a lot for fixing this @gmkumar2005 :pray:

PR #315, opened 2020-07-26
add multi-arch image support

xUnholy:
Signed-off-by: Michael Fornaro <20387402+xUnholy@users.noreply.github.com>

##### SUMMARY
Feature Pull Request: Adding multi-architecture support for ARM/ARM64/AMD64
Note: AMD64 was default supported architecture.

Fixes https://github.com/raspbernetes/multi-arch-images/issues/77


PrasadG193:
@xUnholy Thanks for the PR. Can you please check the docker image size after the change?

xUnholy:
The image size shouldn't have changes as it is still the same binary and same final stage image, it's just now an image that has all the manifest layers for different architectures.

PrasadG193:
Sounds good.

Makefile
```diff
@@ -32,12 +32,19 @@ test: system-check
...
 container-image: pre-build
 	@echo "Building docker image"
-	@docker build --build-arg GOOS_VAL=$(shell go env GOOS) --build-arg GOARCH_VAL=$(shell go env GOARCH) -t $(IMAGE_REPO) -f build/Dockerfile --no-cache .
+	@if ! DOCKER_CLI_EXPERIMENTAL=enabled docker buildx ls | grep -q container-builder; then\
+		DOCKER_CLI_EXPERIMENTAL=enabled docker buildx create --name container-builder --use;\
+	fi
+	DOCKER_CLI_EXPERIMENTAL=enabled docker buildx build --platform "linux/amd64,linux/arm64,linux/arm/v7" \
+		-t $(IMAGE_REPO) \
+		-f build/Dockerfile \
+		. --push
```
PrasadG193, line 47: Is `--push` required here? Can't we just build the image locally?

.github/workflows/ci.yml
```diff
@@ -10,54 +9,67 @@ jobs:
...
+
+      - name: Login to GitHub Docker Registry
+        if: success() && github.event_name == 'push' && endsWith(github.ref ,'develop')
+        run: |
+          docker login -u ${{ secrets.DOCKER_USERNAME }} -p ${{ secrets.DOCKER_PASSWORD }}
+
+      - name: Build Container Image
+        if: success() && github.event_name == 'push' && endsWith(github.ref ,'develop')
+        run: |
+          docker buildx build \
```
PrasadG193, line 45: Can we use `make container-image` here?

Makefile
xUnholy: You can run --load if you'd like to build but not push the image.

.github/workflows/ci.yml
xUnholy: Indeed you can however the builder sets up the environment nicely without needing to setup experimental etc..

Makefile
PrasadG193: Let's do it then. Because if someone outside infracloudio org tries to build an image it will fail. Also, if someone has access to infracloudio org, then can overwrite the latest image upstream which is not desirable.

xUnholy: So we can have the Makefile as --load seeing as this can manually be run from a local machine whilst keep CI to be the only --push? I can reflect those changes on my next commit.

.github/workflows/ci.yml
PrasadG193: @xUnholy let's use the make target so that there will be no difference in a way developers test their images and the one we push. Just to avoid any surprises :)

PrasadG193: @xUnholy We can get this merged once this gets addressed.

xUnholy: The only difference is we've changed `container-image` it would only `--load` and not push, would require more re-writing and unfortunately I'm time poor right now.

HaveFun83:
nice will be great to see this here.  Thanks @xUnholy

RickCoxDev:
Any news on this? I would love to see this!

cerebrate:
Would also like to see. I can hack my own multi-arch image together, but it would be much nicer not to have to.

PrasadG193:
@xUnholy Do you mind if I can take this up?

xUnholy:
> @xUnholy Do you mind if I can take this up?

Absolutely, I have it working for myself, the implementation here has vastly improved with the merge of the official docker action and qemu, feel free to do what is necessary.

anthr76:
Can we get this merged :)

anthr76:
Any updates?

PrasadG193:
We should switch to goreleaser to build and publishing images. https://github.com/infracloudio/botkube/issues/504

PrasadG193:
This is getting addressed with: https://github.com/infracloudio/botkube/pull/509

PR #383, opened 2020-10-01
Reorder error check and defer file.close()

seb-ehm:
##### ISSUE TYPE

 - Bug fix Pull Request

##### SUMMARY

Defering a file.close() should be done after the error returned by file.open() is checked, so I've reordered the statements.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -228,10 +228,11 @@ func NewCommunicationsConfig() (*Communications, error) {
 	configPath := os.Getenv("CONFIG_PATH")
 	communicationConfigFilePath := filepath.Join(configPath, CommunicationConfigFileName)
 	communicationConfigFile, err := os.Open(communicationConfigFilePath)
-	defer communicationConfigFile.Close()
+
 	if err != nil {
 		return c, err
 	}
+	defer communicationConfigFile.Close()
 
 	b, err := ioutil.ReadAll(communicationConfigFile)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


codenio:
> Defering a file.close() should be done after the error returned by file.open() is checked, so I've reordered the statements.

@seb-ehm I feel it is a good practice to add a `defer file.close()` immediately after file opening to avoid missing file closer. it works fine irrespective of its call. could you please justify the need for re-ordering it..?

codenio (commented):
could you please justify the need for re-ordering it..?

PrasadG193:
Closing the PR. Change is not required

seb-ehm:
Thank you for the quick response. I agree that the reordering is not strictly neccessary. However, when I try to familiarize myself with a codebase, I run a static analysis, and this is a warning that came up.
Without reordering, file.Close() is going to return an unhandled error when the config file is not found - it's still going to work, just as you say, I just prefer either handling or avoiding any possible errors. Not handling possible errors for file.Close() in those two cases is fine, as the configs are only read and not written to, so no meaningful errors can occur on file.Close().

PrasadG193:
hey @seb-ehm,
Would you mind sharing more info about static analysis you carried out? E.g which tool you used and results?

seb-ehm:
I've used staticheck and Sonarqube with the default settings. In the past, I've used golangci-lint, but I've had some issues with it lately after some upgrades to my system.. I'm not near a computer right now but will try to post a summary of the results as soon as possible.

seb-ehm:
[golang_ci_lint.txt](https://github.com/infracloudio/botkube/files/5320027/golang_ci_lint.txt)
I've managed to get golangci-lint running again and ran it with a selection of linters, staticcheck included. The Sonarqube report does not add a lot of information, so I didn't include it.
 A lot of the warnings are debatable and are mostly about style (e.g. quite a few from the gocritic linter), but some can be helpful. E.g. the dead code warnings at the beginning; for example
`pkg/notify/elasticsearch.go:40:2: `indexSuffixFormat` is unused (deadcode)
	indexSuffixFormat = "02-01-2006"`
is a leftover of a refactoring and can be safely removed.
Other warnings might be worth a look (e.g. unchecked errors, missing cases in switch ).
Let me know if you think that some of them might be worth fixing.

PR #390, opened 2020-10-01
Update bash scripts

SoftCreatR:
- Double quote variables in shell scripts to prevent globbing and word splitting (see [SC2086](https://github.com/koalaman/shellcheck/wiki/SC2086))
- Declare and assign separately to avoid masking return values (see [SC2155](https://github.com/koalaman/shellcheck/wiki/SC2155))
- Remove useless echo
- Fix code style
- Correct small typo in Makefile

I do currently not have an environment available to test these changes, but in general, they're safe.


PrasadG193 (commented):
The indentation is expected

SoftCreatR:
@PrasadG193 Is there a reason for this "wrong" intendation? Furthermore, it's not really nice of you to ignore the other changes and mark the whole PR as invalid. I don't have a problem with restoring the intendation, because that's not the "main change".

PrasadG193:
Hey @SoftCreatR,
We are receiving a lots of spam PRs. It becomes very hard to check if the change is really required or not. If you have other legit fixes, please feel free to raise another PR.
Hope you understand.

PrasadG193:
Btw indentation fixes are not wrong, but they are unnecessary

SoftCreatR:
> Btw indentation fixes are not wrong, but they are unnecessary

Like I said: That's fine and I can revert them. To make it easier for me, I would love, if you could just reopen this PR and remove the label. Thank you.

PR #398, opened 2020-10-03
Bring Alpine/Go current in Dockerfile

gliptak:


gliptak:
@infracloudio please review

PR #424, opened 2020-10-30
feat(slack): clean up empty cmd handling

sgandon:
##### ISSUE TYPE
 - Bug fix Pull Request

##### SUMMARY
The empty command `@BotKube` is managed in the executor.go file for having a identic behaviour on all intergrations but the Slack bot is handling it differently.
So I proposed a cleanup in this  PR.


pkg/bot/slack.go
```diff
@@ -150,10 +150,7 @@ func (sm *slackMessage) HandleMessage(b *SlackBot) {
 	}
 
 	// Trim the @BotKube prefix
-	sm.Request = strings.TrimPrefix(sm.Event.Text, "<@"+sm.BotID+"> ")
-	if len(sm.Request) == 0 {
-		return
-	}
+	sm.Request = strings.TrimPrefix(sm.Event.Text, "<@"+sm.BotID+">")
```
PrasadG193, line 153: Does empty command e.g `@BotKube` return an error message now?

PrasadG193: @sgandon could you confirm this? Rest looks good

sgandon:
I don't understand but my response is in "pending state" !!!
here is it
Yes it does.
this is managed by the executor : https://github.com/infracloudio/botkube/pull/424/files#diff-fccd3962f0d0da63e38e16963f1ac9ecdc37cbbe33beed7e6aee1d0de163f4e1R183
This way there is a common behaviour between all integrations

PrasadG193:
Cool! Thanks @sgandon

PrasadG193 (approved):

PR #430, opened 2020-11-05
[ES] indices to be created with daywise timestamp

kartik-moolya:
##### ISSUE TYPE

 - Feature Pull Request

##### SUMMARY

Elasticsearch Indices will be created with timestamp suffix

### DIFF
--- a/pkg/notify/elasticsearch.go
+++ b/pkg/notify/elasticsearch.go
@@ -98,7 +98,7 @@ func NewElasticSearch(c config.ElasticSearch) (Notifier, error) {
 	}
 	return &ElasticSearch{
 		ELSClient: elsClient,
-		Index:     c.Index.Name + "-" + time.Now().Format(indexSuffixFormat),
+		Index:     c.Index.Name,
 		Type:      c.Index.Type,
 		Shards:    c.Index.Shards,
 		Replicas:  c.Index.Replicas,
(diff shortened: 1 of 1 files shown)


pkg/notify/elasticsearch.go
```diff
@@ -112,7 +113,7 @@ type index struct {
 
 func (e *ElasticSearch) flushIndex(ctx context.Context, event interface{}) error {
 	// Create index if not exists
-	exists, err := e.ELSClient.IndexExists(e.Index).Do(ctx)
+	exists, err := e.ELSClient.IndexExists(e.Index + "-" + time.Now().Format(indexSuffixFormat)).Do(ctx)
```
PrasadG193, line 116: `e.Index + "-" + time.Now().Format(indexSuffixFormat)` This is repeated at several places. Can we make it a part of struct and use that?

kartik-moolya: fixed

pkg/notify/elasticsearch.go
```diff
@@ -91,7 +98,7 @@ func NewElasticSearch(c config.ElasticSearch) (Notifier, error) {
 	}
 	return &ElasticSearch{
 		ELSClient: elsClient,
-		Index:     c.Index.Name,
+		Index:     c.Index.Name + "-" + time.Now().Format(indexSuffixFormat),
```
PrasadG193, line 101: We create ElasticSearch object when BotKube starts - https://github.com/infracloudio/botkube/blob/develop/cmd/botkube/main.go#L59. If we create the index here, it will not be updated if BotKube stays running forever. We should re-calculated index (since it includes timestamp) each time when `SendEvent` is called

kartik-moolya: refactored

PrasadG193 (approved):

PrasadG193:
Thanks for fixing this @kartik-moolya :pray:

PR #437, opened 2020-12-08
Fixed namespace checker to use the correct name

### LINKED ISSUE #419: [BUG] Ignored namespace list gets ignored
Since updating from 0.10.0 to 0.11.0, ignoring some namespaces while including "all" doesn't work--the ignore list seems to simply be ignored:

```
config:
  ## Resources you want to watch
  resources:
  - name: v1/pods
    namespaces:
      include:
        - all
      ignore:
        - test
    events:
      - all
    updateSetting:
      includeDiff: true
      fields:
        - status.availableReplicas

  # Check true if you want to receive recommendations
  # about the best practices for the created resource
  recommendations: true

  # Cluster Setting to manage command execution access
  settings:
    # Set true to enable kubectl commands execution
    allowkubectl: false

    # Set true to enable commands execution from configured channel only
    restrictAccess: false

    # Set true to enable config watcher
    configwatcher: true

    # Set false to disable upgrade notification
    upgradeNotifier: true
```

Specifying certain namespaces to include does work, but isn't ideal

hellozee:
##### ISSUE TYPE
 - Bug fix Pull Request

##### SUMMARY
Apparently the wrong member was used to check the resource name. I don't
know how resillient this patch is, but it works for most of the part.

Fixes #419

### DIFF
--- a/pkg/filterengine/filters/namespace_checker.go
+++ b/pkg/filterengine/filters/namespace_checker.go
@@ -56,8 +56,7 @@ func (f NamespaceChecker) Run(object interface{}, event *events.Event) {
 	}
 	if botkubeConfig != nil {
 		for _, resource := range botkubeConfig.Resources {
-			resourceName := strings.Split(strings.ToLower(event.Title), " ")[0]
-			if resource.Name == resourceName {
+			if resource.Name == event.Resource {
 				// check if namespace to be ignored
 				if isNamespaceIgnored(resource.Namespaces, event.Namespace) {
 					event.Skip = true
--- a/pkg/filterengine/filters/namespace_checker.go
+++ b/pkg/filterengine/filters/namespace_checker.go
@@ -56,7 +56,8 @@ func (f NamespaceChecker) Run(object interface{}, event *events.Event) {
 	}
 	if botkubeConfig != nil {
 		for _, resource := range botkubeConfig.Resources {
-			if resource.Name == strings.ToLower(event.Kind) {
+			resourceName := strings.Split(strings.ToLower(event.Title), " ")[0]
+			if resource.Name == resourceName {
 				// check if namespace to be ignored
 				if isNamespaceIgnored(resource.Namespaces, event.Namespace) {
 					event.Skip = true
--- a/pkg/events/events.go
+++ b/pkg/events/events.go
@@ -51,6 +51,7 @@ type Event struct {
 	Count     int32
 	Action    string
 	Skip      bool `json:",omitempty"`
+	Resource  string
 
 	Recommendations []string
 	Warnings        []string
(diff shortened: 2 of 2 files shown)


pkg/filterengine/filters/namespace_checker.go
```diff
@@ -56,7 +56,8 @@ func (f NamespaceChecker) Run(object interface{}, event *events.Event) {
 	}
 	if botkubeConfig != nil {
 		for _, resource := range botkubeConfig.Resources {
-			if resource.Name == strings.ToLower(event.Kind) {
+			resourceName := strings.Split(strings.ToLower(event.Title), " ")[0]
+			if resource.Name == resourceName {
```
PrasadG193, line 60: @hellozee we have a helper function implemented which returns GVR from given Kind https://github.com/infracloudio/botkube/blob/develop/pkg/utils/utils.go#L258. We can use that method to build "group/version/resource" or "version/resource" (if the group is nil) check

PrasadG193 (changes requested):

pkg/filterengine/filters/namespace_checker.go
hellozee: This is not working for delete events since the resource type return by that method is `/`. Here is how I am trying to do it.
```go
var eventObj *coreV1.Event
err := utils.TransformIntoTypedObject(object.(*unstructured.Unstructured), &eventObj)
if err != nil {
    log.Errorf("Unable to transform object type: %v, into type: %v", reflect.TypeOf(object), reflect.TypeOf(eventObj))
    return
}
tempResourceName, _ := utils.GetResourceFromKind(eventObj.InvolvedObject.GroupVersionKind())
resourceName := utils.GVRToString(tempResourceName)
```

PrasadG193: @hellozee let's do this
1. Add a new field `Resource` in the Event struct
2. Set the Resource value while creating New Event
3. Refer event.Resource while building Title as well as in filters
Sounds good?

hellozee: Cool, will do that instead, 🐱

billimek: @hellozee will you be able to implement the suggested changes?

hellozee: Ahh sorry, completely forgot about it, will do by this weekend, 😸

PrasadG193: Hey @hellozee, please let me know if you are working on this.

hellozee: whoops! expect a PR by evening for sure, 🙃

hellozee:
Done, but this doesn't work in case of a delete event.

PrasadG193:
@hellozee do you mind adding the e2e test for the namespace checker filter (https://github.com/infracloudio/botkube/blob/develop/test/e2e/filters/filters.go#L44)? I am fine if we address that as a separate task.
Please file an issue if you can't add that in this PR.
Thank you!

PrasadG193 (approved):

hellozee:
It would be better if someone else picks it. I cant promise anything this week, sorry. If no one picks it up this week, I would put up a patch next week, 🐱

PrasadG193:
Sounds good @hellozee

PrasadG193:
Filed https://github.com/infracloudio/botkube/issues/443 to track. Merging this PR

PrasadG193:
@hellozee this doesn't seem to be working. I've added a fix https://github.com/infracloudio/botkube/pull/449. PTAL

Issue #429, opened 2020-11-04
defaultNamespace field in config won't work on changing to other than default


A-kanksh-a:
**Describe the bug**
Botkube doesn't produce report of other namespace except default even if resource configuration is changed

**To Reproduce**
1. Go to 'resource_config.yaml'
2. Change the default namespace field other than default for ex- `botkube`
3. deploy botkube and then run the command `@Botkube get deployments` from the applications. 
4. It will show results of default namespace instead of showing result for `botkube` namespace which is set to be default

**Expected behavior**
The result expected should be resources listed in `botkube` namespace

**Screenshots**
![image](https://user-images.githubusercontent.com/41359156/98121750-b5174300-1ed5-11eb-928a-8a849424feed.png)

 Results after deploying botkube
![image](https://user-images.githubusercontent.com/41359156/98121905-ec85ef80-1ed5-11eb-8cb1-fced266dc67e.png)

**Additional context**
If -n flag is provided then the response is correct

sgandon:
Indeed I have also noticed it and fixed it in a PR I am preparing for #421 .

alexavrutin:
I also encountered this problem but in my case adding -n <namespace name> didn't help. The bot simply does not return anything if I specify -n. If I don't specify -n, it returns "No resources found..." Any ideas what might cause it?

sgandon:
@alexavrutin to be sure about what you are doing, the `-n` shall be used to specify a namespace like `@botkube get pods -n default` where `default` is the namespace named `default`.
You can just use `-n` without any string following it.

sramek5:
#458

PrasadG193:
Fixed with: https://github.com/infracloudio/botkube/pull/459

PR #445, opened 2021-01-13
feat(#421) : add default cluster and namespace

### LINKED ISSUE #421: add support for default cluster and namespace on any Slack channels
### **Is your feature request related to a problem? Please describe.**
When inviting @botkube on any channel or using it in @botkube DM,
If you want to issue a series of kubectl command on a cluster it is really annoying to have to specify the cluster name or the namespace all the time.
 

### **Describe the solution you'd like**
It would be nice to have the same behaviour as one expect from the kubectl command line that is having a default cluster and namespace.
I would suggest to add a couple of commands :
```
1/   @botkube cluster [<name of the default cluster>]
2/   @botkube namespace [<name of the default namespace>]
```
1/ the first command would set the default cluster when specified or display the default cluster if name is empty
2/ the second command would set the default namespace when specified or display the default namespace if name is empty

This would be refused for channels that are linked to a cluster using the configuration yaml file.

### **Additional context**
In terms of implementation (if ever possible) this would mean the default cluster that is set would be the only one to respond to this channel when `--cluster-name` is not used.
It would have to store the channel name in order to know if it can answer or not. The default namespace should be stored along with the channel name and used when the backend chooses to respond.
This storage could be done in-memory (for simple implem) but this would mean to loose those settings if the backend pod restarts or it could be persisted (PV or ConfigMap).

sgandon:
##### ISSUE TYPE
 - Feature Pull Request
 - Bug fix Pull Request

##### SUMMARY
This PR add a default behaviour for kubectl command where a channel can now have a default cluster and a default namespace associated with it to avoid repeating those when doing kubectl commands.
There is only an implementation for slack.

Fixes #421 and #429

### DIFF
--- a/pkg/bot/slack.go
+++ b/pkg/bot/slack.go
@@ -134,7 +134,8 @@ func setupDefaultChannelForKubeCtl(client slack.Client, channelName string, clus
 	msgOpt := slack.MsgOptionText(fmt.Sprintf(execute.DefaultClusterForKubectl, clusterName), false)
 	channelID, _, err := client.PostMessage(channelName, msgOpt, slack.MsgOptionAsUser(true))
 	if err != nil {
-		log.Errorf("Error in sending slack message %s", err.Error())
+		log.Errorf("Error in sending slack message to channel %s : %s", channelName, err.Error())
+		return
 	}
 	if namespace == "" {
 		namespace = "default"
--- a/pkg/bot/slack.go
+++ b/pkg/bot/slack.go
@@ -20,6 +20,7 @@
 package bot
 
 import (
+	"fmt"
 	"strings"
 
 	"github.com/infracloudio/botkube/pkg/config"
@@ -85,7 +86,7 @@ func (b *SlackBot) Start() {
 		switch ev := msg.Data.(type) {
 		case *slack.ConnectedEvent:
 			log.Info("BotKube connected to Slack!")
-
+			setupDefaultChannelForKubeCtl(RTM.Client, b.ChannelName, b.ClusterName, b.DefaultNamespace)
 		case *slack.MessageEvent:
 			// Skip if message posted by BotKube
 			if ev.User == botID {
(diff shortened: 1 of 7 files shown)


pkg/config/config.go
```diff
@@ -213,7 +217,7 @@ type Webhook struct {
 type Kubectl struct {
 	Enabled          bool
 	Commands         Commands
-	DefaultNamespace string
+	DefaultNamespace string `yaml:"defaultNamespace"`
```
sgandon, line 220: fixes #429

pkg/bot/slack.go
```diff
@@ -85,7 +86,7 @@ func (b *SlackBot) Start() {
 		switch ev := msg.Data.(type) {
 		case *slack.ConnectedEvent:
 			log.Info("BotKube connected to Slack!")
-
+			setupDefaultChannelForKubeCtl(RTM.Client, b.ChannelName, b.ClusterName, b.DefaultNamespace)
```
PrasadG193, line 89: Hey @sgandon,
In case you have missed it, we don't have to specify `--cluster-name` flag if the command is executed from the configured channel.
In the case of multicluster setup, e.g `Cluster A -> Channel X` and `Cluster B -> Channel Y`, the BotKube commands posted in `Channel X` will be executed on `Cluster A`.
We also follow the same pattern for other integrations - Mattermost and Discord. Please confirm this doesn't change the existing behavior.

pkg/config/config.go
```diff
@@ -88,6 +88,10 @@ var CommunicationConfigFileName = "comm_config.yaml"
 // Notify flag to toggle event notification
 var Notify = true
 
+//KubeCtlLinkedChannels is a set that contains the channel ID (as a key)of all channels linked to
+//this cluster and containing the default namespace as a value
```
PrasadG193, line 92:
```suggestion
// KubeCtlLinkedChannels is a set that contains the channel ID (as a key)of all channels linked to
// this cluster and containing the default namespace as a value
```
Please add leading space in comments to have consistent pattern across the source code

pkosiec:
Hi @sgandon, 
Thank you for your contribution. I'm closing this PR because of inactivity and multiple merging conflicts. I also posted a comment in the issue #421, as in 0.13 we introduced some features which could solve your use-cases. Let's discuss it further under the GH issue! 🙂

PR #448, opened 2021-01-15
Add option to get clustername from google metadata

thijsdejong:
##### ISSUE TYPE

 - Feature Pull Request
 
##### SUMMARY

This adds the option to get the cluster name from google metadata server. This is very useful in a multicluster setup on GCP that we use.

`cloud.google.com/go v0.38.0`

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -24,6 +24,7 @@ import (
 	"os"
 	"path/filepath"
 
+	"cloud.google.com/go/compute/metadata"
 	"gopkg.in/yaml.v2"
 )
 
@@ -275,6 +276,15 @@ func New() (*Config, error) {
 
 	if len(b) != 0 {
 		yaml.Unmarshal(b, c)
+	}
+
+	if c.Settings.ClusterName == "%google_metadata%" {
+		clusterName, err := metadata.InstanceAttributeValue("cluster-name")
+		if err != nil {
+			return nil, err
+		}
+
+		c.Settings.ClusterName = clusterName
 	}
 
 	comm, err := NewCommunicationsConfig()
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -278,7 +278,7 @@ func New() (*Config, error) {
 		yaml.Unmarshal(b, c)
 	}
 
-	if c.Settings.ClusterName == "%google_metadata%" {
+	if c.Settings.ClusterName == "+google_metadata+" {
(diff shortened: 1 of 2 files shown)


timonvw:
This would be really nice !

PrasadG193:
@thijsdejong thanks for the PR. Do we have to set clustername as `google_metadata` to use this feature? Do we need to update documentation for this?

thijsdejong:
> @thijsdejong thanks for the PR. Do we have to set clustername as `google_metadata` to use this feature? Do we need to update documentation for this?

Hi,

You have to set clustername to '%google_metadata%' to use this feature. It would be nice to add a line in the docs that mentions this. Where can I do this?

PrasadG193:
@thijsdejong documentation for botkube.io is placed at https://github.com/infracloudio/botkube-docs. We use hugo to build that. Could you please update the installation and configuration guide as per the change? Please LMK if you need any help with that.

pkg/config/config.go
```diff
@@ -277,6 +278,15 @@ func New() (*Config, error) {
 		yaml.Unmarshal(b, c)
 	}
 
+	if c.Settings.ClusterName == "+google_metadata+" {
```
PrasadG193, line 281: Can we use something like `{{ google_metada }}` instead? Looks more like go templating

hermanbanken: Using {{}} would suggest that Go templating is actually supported. If it is not, then this only leads to confusion, I think.

Would `$GOOGLE_METADATA` or `google-metadata:///cluster` make more sense to you?

PrasadG193: Sounds good to me.

pkosiec:
Hello there,

I'm not really sure if that's the proper way of implementing this 🤔 It feels to me like this is not the BotKube's responsibility. In that way, later we would need to support other cloud providers and that would be hard to maintain.

However, we could be more flexible with cluster name configuration. What @mszostok introduced in #601 could be supported in `config` as well. In that way the `clusterName` could be taken from a different place, e.g. ConfigMap.
That would support such custom scenarios - @thijsdejong you would be able to e.g. create a Job which creates such ConfigMap with cluster name. IMO simple and clean.

@mszostok @PrasadG193 @thijsdejong What do you think?

Cheers!

pkosiec:
Alright, so let's close it and implement in the way as I mentioned above. You can follow the implementation under this issue: https://github.com/kubeshop/botkube/issues/643. Thanks!

PR #459, opened 2021-02-24
corrected defaultNamespace population and avoid appending defaultnamespace if already passed 

girishg4t:
##### ISSUE TYPE

 - Bug fix Pull Request

##### SUMMARY

Resolved issue of DefaultNamespace value is not getting populated in config and appending default namespace if another namespace is already passed

### DIFF
--- a/pkg/execute/executor.go
+++ b/pkg/execute/executor.go
@@ -256,10 +256,8 @@ func trimQuotes(clusterValue string) string {
 func runKubectlCommand(args []string, clusterName, defaultNamespace string, isAuthChannel bool) string {
 
 	// run commands in namespace specified under Config.Settings.DefaultNamespace field
-	if len(defaultNamespace) != 0 {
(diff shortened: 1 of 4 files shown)


pkg/execute/executor.go
```diff
@@ -256,10 +256,8 @@ func trimQuotes(clusterValue string) string {
 func runKubectlCommand(args []string, clusterName, defaultNamespace string, isAuthChannel bool) string {
 
 	// run commands in namespace specified under Config.Settings.DefaultNamespace field
-	if len(defaultNamespace) != 0 {
+	if !utils.Contains(args, "-n") && len(defaultNamespace) != 0 {
```
PrasadG193, line 259: Thank you @girishg4t for catching this. Namespace can also be passed with `--namespace` flag, let's handle that too.

girishg4t: I forgot to add that, Thanks @PrasadG193

PrasadG193 (approved):

PR #461, opened 2021-03-03
remove unwanted hyperlink text from command

girishg4t:
##### ISSUE TYPE

 - Bug fix Pull Request

##### SUMMARY
When clustername is passed to botkube from command and if that clustername contains valid domain then slack automatically adds <http:// to the url  
eg.  
if command is `@BotKube get pods --cluster-name xyz.alpha-sense.org`  then in botkube we get cluster name as `<http://xyz.alpha-sense.org|xyz.alpha-sense.org>`

#### Solution:
Added small logic in executor code to remove the unwanted string

Fixes https://github.com/infracloudio/botkube/issues/460

### DIFF
--- a/pkg/execute/executor.go
+++ b/pkg/execute/executor.go
@@ -181,7 +181,9 @@ func NewDefaultExecutor(msg string, allowkubectl, restrictAccess bool, defaultNa
 
 // Execute executes commands and returns output
 func (e *DefaultExecutor) Execute() string {
-	args := strings.Fields(strings.TrimSpace(e.Message))
+	//Remove hyperlink if it got added automatically
+	command := utils.RemoveHyperlink(e.Message)
+	args := strings.Fields(strings.TrimSpace(command))
 	if len(args) == 0 {
 		if e.IsAuthChannel {
 			return printDefaultMsg(e.Platform)
@@ -278,12 +280,6 @@ func runKubectlCommand(args []string, clusterName, defaultNamespace string, isAu
 		if strings.HasPrefix(arg, ClusterFlag.String()) {
 			// Check if flag value in current or next argument and compare with config.settings.clustername
 			if arg == ClusterFlag.String() {
-				nextArgasClusterName := args[index+1]
-				//This will remove 'http://' from clustername if it is automatically added
-				if strings.Contains(nextArgasClusterName, "<http://") {
-					nerCommands := strings.Split(nextArgasClusterName, "|")
-					args[index+1] = strings.Replace(nerCommands[1], ">", "", 1)
-				}
 				if index == len(args)-1 || trimQuotes(args[index+1]) != clusterName {
 					return ""
 				}
--- a/pkg/execute/executor.go
+++ b/pkg/execute/executor.go
@@ -181,7 +181,7 @@ func NewDefaultExecutor(msg string, allowkubectl, restrictAccess bool, defaultNa
 
 // Execute executes commands and returns output
 func (e *DefaultExecutor) Execute() string {
-	//Remove hyperlink if it got added automatically
+	// Remove hyperlink if it got added automatically
 	command := utils.RemoveHyperlink(e.Message)
 	args := strings.Fields(strings.TrimSpace(command))
 	if len(args) == 0 {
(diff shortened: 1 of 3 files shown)


pkg/execute/executor.go
```diff
@@ -278,6 +278,12 @@ func runKubectlCommand(args []string, clusterName, defaultNamespace string, isAu
 		if strings.HasPrefix(arg, ClusterFlag.String()) {
 			// Check if flag value in current or next argument and compare with config.settings.clustername
 			if arg == ClusterFlag.String() {
+				nextArgasClusterName := args[index+1]
+				//This will remove 'http://' from clustername if it is automatically added
+				if strings.Contains(nextArgasClusterName, "<http://") {
+					nerCommands := strings.Split(nextArgasClusterName, "|")
+					args[index+1] = strings.Replace(nerCommands[1], ">", "", 1)
+				}
```
PrasadG193, line 286: @girishg4t we need the fix generic. Slack actually converts any string that matcher URL pattern into hyperlink. If we want to list CRDs/CRs e.g `@BotKube get prometheuses.monitoring.coreos.com`, in this case, the resource name will be converted into hyperlink. I had fixed this in one of the POCs where I had used regex to do that. Please take a look at this commit for reference: https://github.com/infracloudio/botkube/commit/08fe7731ddb698454cd7f680c4af2257627d266f

girishg4t: Thanks @PrasadG193 i have corrected the code as per review comment, Please check and let me know if any change require

pkg/execute/executor.go
```diff
@@ -181,7 +181,9 @@ func NewDefaultExecutor(msg string, allowkubectl, restrictAccess bool, defaultNa
 
 // Execute executes commands and returns output
 func (e *DefaultExecutor) Execute() string {
-	args := strings.Fields(strings.TrimSpace(e.Message))
+	//Remove hyperlink if it got added automatically
```
PrasadG193, line 184:
```suggestion
	// Remove hyperlink if it got added automatically
```
Please keep consistent format for comments

pkg/utils/utils.go
```diff
@@ -397,3 +397,18 @@ func Contains(a []string, x string) bool {
 	}
 	return false
 }
+
+//RemoveHyperlink removes the hyperlink text from url
+func RemoveHyperlink(hyperlink string) string {
+	var command = hyperlink
+	var hyperlinkRegex = regexp.MustCompile(`(?m)<http:\/\/[a-z.0-9\/\-_=]*\|([a-z.0-9\/\-_=]*)>`)
```
PrasadG193, line 404: Please declare regex string as constant

pkg/utils/utils_test.go
```diff
@@ -83,3 +83,25 @@ func TestContains(t *testing.T) {
...
+func TestRemoveHypelink(t *testing.T) {
+
+	type test struct {
+		input    string
+		expected string
+	}
+
+	tests := []test{
+		{input: "get <http://prometheuses.monitoring.coreos.com|prometheuses.monitoring.coreos.com> --cluster-name <http://xyz.alpha-sense.org|xyz.alpha-sense.org>", expected: "get prometheuses.monitoring.coreos.com --cluster-name xyz.alpha-sense.org"},
+		{input: "get <http://prometheuses.monitoring.coreos.com|prometheuses.monitoring.coreos.com>", expected: "get prometheuses.monitoring.coreos.com"},
```
PrasadG193, line 96: Please add a testcase with hyperlinks for both - resource and cluster name

pkg/utils/utils.go
```diff
@@ -397,3 +397,18 @@ func Contains(a []string, x string) bool {
 	}
 	return false
 }
+
+//RemoveHyperlink removes the hyperlink text from url
+func RemoveHyperlink(hyperlink string) string {
+	var command = hyperlink
```
PrasadG193, line 403:
```suggestion
	command := hyperlink
```

PrasadG193 (changes requested):

pkg/utils/utils_test.go
girishg4t: first case is for resource and clustername

pkg/execute/executor.go
girishg4t: Done

pkg/utils/utils.go
girishg4t: done

pkg/utils/utils.go
girishg4t: chanaged

PrasadG193 (approved):

PR #483, opened 2021-05-11
Fixed BOTKUBE_VERSION

pierluigilenoci:
##### ISSUE TYPE

 - Bug fix Pull Request

##### SUMMARY

The software notifies on the channel if you are using the updated version of BotKube or not. To do this, use the BOTKUBE_VERSION variable which is set up with the `.Chart.Version` value which does not necessarily correspond to the version of the software running on the cluster. The actual version of the software that runs is the one that comes from `.Values.image.tag`. 

Screenshot 2021-05-11 at 09 51 39


PrasadG193 (approved):

PrasadG193:
We need to revisit version logic to use build-time variables. But this can be done as a separate task.

pierluigilenoci:
@PrasadG193 all build failed? Why?

pierluigilenoci:
@PrasadG193 what can we do to push forward?

PrasadG193:
@pierluigilenoci looks like one of the dep breaking builds. Let me fix that.

PR #525, opened 2021-09-17
Notification disable at namespace level (#492)

### LINKED ISSUE #492: Include or explude namespaces based on labels
Hello,

I think that botkube should be able to adapt its selective filter according to labels on the namespace, i.e I would like to be able to either include or exclude namespaces depending on some labels that the user would choose (`production` for example).
I know that a wildcard filter exists for excluding but I would rather my logic not to be dictated by the namespace name but rather by the labels attached to it.

As far as I understood, annotating the namespace with `botkube.io/disable: true` would only disable changes of the namespace itself and not its content and thus it doesn't fill the gap (and it would be too opinionated of a label in my sense anyway).

I hope I haven't missed any already existing feature, and would like to hear what you think.
Thanks!

praddy26:
##### ISSUE TYPE
 - Feature Pull Request

##### SUMMARY
Namespace annotation "botkube.io/notification" has been added to enable/disable notifications of objects present within the namespace. When the value of this annotation is set to "false", all the notifications of objects present within that namespace will be disabled. If this annotation is absent or if it's present and the value of it is set to anything else other than "false", then the notifications will be sent as per the configurations present in the resource config file.

To keep track of annotations, cache has been created using the following module: https://github.com/patrickmn/go-cache

Fixes #492


pkg/filterengine/filters/object_annotation_checker.go
```diff
@@ -82,6 +107,101 @@ func isObjectNotifDisabled(obj metaV1.ObjectMeta) bool {
...
+// isObjectNamespaceNotifEnabled checks annotation botkube.io/notification on the namespace
+// annotation botkube.io/notification enable/disable all object's event notification in the namespace
+func isObjectNamespaceNotifEnabled(obj metaV1.ObjectMeta) bool {
+
+	namespace := obj.Namespace
+	nsObjectNotifEnabled, found := NamespaceCache.Get(namespace)
+
+	if !found {
+		log.Warnf("Cache Miss!! Namespace '%s' not found in the cache!", namespace)
+
```
PrasadG193, line 119: In go, we try to avoid else block for better readability. 
You can return if found, and let move rest of the code outside else
```
	if found {
	        log.Infof("Cache Hit!! Namespace '%s' found in the cache!", namespace)
	        return nsObjectNotifEnabled
	}
       // rest of the logic if not found
```

pkg/filterengine/filters/object_annotation_checker.go
```diff
@@ -82,6 +107,101 @@ func isObjectNotifDisabled(obj metaV1.ObjectMeta) bool {
...
+		// get namespace object
+		nsObject, err := utils.DynamicKubeClient.Resource(schema.GroupVersionResource{Group: "", Version: "v1", Resource: "namespaces"}).Get(context.TODO(), namespace, v1.GetOptions{})
+
+		if err != nil {
+			log.Error(err)
+			nsObjectNotifEnabled = "false"
+		} else {
```
PrasadG193, line 126: Same here. return early and get rid of else block

pkg/filterengine/filters/object_annotation_checker.go
```diff
@@ -82,6 +107,101 @@ func isObjectNotifDisabled(obj metaV1.ObjectMeta) bool {
...
+
+}
+
+// Registers watch for namespace's events
+func RegisterWatchNamespace() error {
+
+	log.Info("Registering namespace watch")
+
+	// register watch on namespaces
+	nsWatcher, err := utils.DynamicKubeClient.Resource(schema.GroupVersionResource{Group: "", Version: "v1", Resource: "namespaces"}).Watch(context.TODO(), v1.ListOptions{})
```
PrasadG193, line 153: The watcher gets closed after some time, please use RetryWatcher - https://pkg.go.dev/k8s.io/client-go/tools/watch#NewRetryWatcher

pkg/filterengine/filters/object_annotation_checker.go
```diff
@@ -82,6 +107,101 @@ func isObjectNotifDisabled(obj metaV1.ObjectMeta) bool {
...
+
+		// check existence of namespace
+		_, err := utils.DynamicKubeClient.Resource(schema.GroupVersionResource{Group: "", Version: "v1", Resource: "namespaces"}).Get(context.TODO(), namespace, v1.GetOptions{})
+		if err != nil {
+			// namespace does not exist
+			log.Debugf("Namespace '%s' does not exist!!", namespace)
+			log.Error(err)
+			// delete namespace entry from cache
+			NamespaceCache.Delete(notifiedObj.GetName())
+			log.Debugf("Namespace Cache: %v", NamespaceCache.Items())
```
PrasadG193, line 189: Return here and get rid of else block

pkg/filterengine/filters/object_annotation_checker.go
```diff
@@ -82,6 +107,101 @@ func isObjectNotifDisabled(obj metaV1.ObjectMeta) bool {
...
+			NamespaceCache.Set(nsObject.GetName(), val, cache.NoExpiration)
+			nsObjectNotifEnabled = val
+			log.Debugf("Namespace Cache: %v", NamespaceCache.Items())
+		}
+
+	} else {
+		log.Infof("Cache Hit!! Namespace '%s' found in the cache!", namespace)
+	}
+
+	b1, _ := strconv.ParseBool(nsObjectNotifEnabled.(string))
```
PrasadG193, line 142: Please handle the returned error as a second return value

PrasadG193:
@praddy26 few suggestions on the caching mechanism you have implemented-
1. While initializing cache, you should list all the namespaces and build the cache
2. Whenever you get namespace event, filter older events comparing object.Medatada.CreationTimestamp 
3. We don't need to Get namespace or check namespace for existence each time when we get event if we filter older events based on creation timestamp

pkg/filterengine/filters/object_annotation_checker.go
praddy26: How to manage common code which has to be executed in either case. Should we wrap it in both if and else condition the same piece of code.

pkosiec:
Hey,
It's been a while - I'm closing the PR because of inactivity 🙂 Let's discuss further under https://github.com/kubeshop/botkube/issues/492. Cheers!

PR #527, opened 2021-09-27
fix: Allow users to disable ANSI color in logging

### LINKED ISSUE #477: Allow users to disable ANSI color in logging
Botkube logs seem to have [ANSI color codes on by default](https://github.com/infracloudio/botkube/blob/develop/pkg/log/log.go#L42) (and provide no configuration to disable them). In our cluster, we ship logs to an ELK stack offsite (and do not often view them using a traditional terminal). The ANSI codes impair readability of logs when not viewed in a terminal.

Provide a setting or configuration that allows users to disable ANSI escape codes in log output.

n/a

**Additional context**
Here's an example log line I see in our Kibana instance:

```
[36mINFO[0m[2021-04-26T15:52:21Z] Registering the filter IngressValidator      
```

sumit-tembe:
##### ISSUE TYPE
 - Bug fix Pull Request

Fixes #477

### DIFF
--- a/pkg/log/log.go
+++ b/pkg/log/log.go
@@ -21,6 +21,7 @@ package log
 
 import (
 	"os"
+	"strconv"
 
 	"github.com/sirupsen/logrus"
 )
@@ -39,7 +40,11 @@ func init() {
 		logLevel = logrus.InfoLevel
 	}
 	log.SetLevel(logLevel)
-	log.Formatter = &logrus.TextFormatter{ForceColors: true, FullTimestamp: true}
+	forceColors, err := strconv.ParseBool(os.Getenv("LOG_FORCE_COLORS"))
+	if err != nil {
+		forceColors = false
+	}
+	log.Formatter = &logrus.TextFormatter{ForceColors: forceColors, FullTimestamp: true}
 }
 
 // Info map logrus.Info func to log.Info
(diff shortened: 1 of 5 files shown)


pkg/log/log.go
```diff
@@ -39,7 +40,8 @@ func init() {
 		logLevel = logrus.InfoLevel
 	}
 	log.SetLevel(logLevel)
-	log.Formatter = &logrus.TextFormatter{ForceColors: true, FullTimestamp: true}
+	forceColors, _ := strconv.ParseBool(os.Getenv("LOG_FORCE_COLORS"))
```
PrasadG193, line 43: what value do we get when `ParseBool` fails? We should check the error and set forceColors=false if is not nil

sumit-tembe: ParseBool returns `false` as default value. (Even in case of an error)

PrasadG193: I still think it's better to handle error for readability

pkosiec:
Hi @sumit-tembe, cool PR - would you like to reopen and rebase it to latest `develop`? It should be even easier to do than before, as now we are using `envconfig`: https://github.com/infracloudio/botkube/blob/develop/cmd/botkube/main.go#L48

Thanks!

PR #537, opened 2021-11-02
feat(helm): allow to set deployment annotations

### LINKED ISSUE #536: [FR] Allow to set deployment annotation in Helm Chart

Currently we cannot set deployment annotations using Helm Chart.

We can only set pod annotations using [`extraAnnotations`](https://github.com/infracloudio/botkube/blob/5cc97c3791f44b2db3370f494d0819362b1bd453/helm/botkube/templates/deployment.yaml#L26), it would be nice to be able to add annotations to deployment too.

- create our own chart from yours and update [`helm/botkube/templates/deployment.yaml`](https://github.com/infracloudio/botkube/blob/develop/helm/botkube/templates/deployment.yaml)
- annotate deployment manually : `kubectl annotate deploy botkube key=value`

**Additional context**

Most charts have values for `deployment.annotations`. For example elastic.co charts have those : 
- podAnnotations (equivalent to `extraAnnotations` in this chart)
- deployment.annotations
- daemonset.annotations

I've created a PR and tested it with/without annotations : https://github.com/infracloudio/botkube/pull/537

lucagouty:
##### ISSUE TYPE

 - Feature Pull Request

##### SUMMARY

Allow to set deployment annotations in helm chart.

Fixes #536

##### Remarks

I hesitated to use `extraAnnotations`, but when I look at other official helm charts most of the time they use different values for pods and deployments annotations. 

Imho we should also rename `extraAnnotations` to `podAnnotations` but would be a breaking change so I left it as it was.

### DIFF
--- a/helm/botkube/templates/deployment.yaml
+++ b/helm/botkube/templates/deployment.yaml
@@ -9,6 +9,10 @@ metadata:
     app.kubernetes.io/managed-by: {{ .Release.Service }}
     component: controller
     app: botkube
+  annotations:
+    {{- if .Values.deployment.annotations }}
+{{ toYaml .Values.deployment.annotations | indent 4 }}
+    {{- end }}
 spec:
   replicas: {{ .Values.replicaCount }}
   selector:
--- a/helm/botkube/values.yaml
+++ b/helm/botkube/values.yaml
@@ -362,6 +362,9 @@ serviceMonitor:
   port: metrics
   labels: {} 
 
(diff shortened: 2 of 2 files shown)


PrasadG193:
Hey @lucagouty thanks for the PR.
@bhavin192 could you please review it?

PrasadG193 (approved):

PrasadG193:
LGTM

PR #540, opened 2021-11-12
[helm] Set empty list as default value for extraEnv

dennybaa:


helm/botkube/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: botkube
 home: https://www.botkube.io
-version: v0.12.3
+version: v0.12.4
```
PrasadG193, line 4: This change is not required, v0.12.3 is the latest release

PrasadG193: @dennybaa the version will be bumped automatically at the time of release. So when we release the next version of the helm chart, it will have the correct version.

PrasadG193 (approved):
Looks good

helm/botkube/Chart.yaml
PrasadG193: @dennybaa could you please resolve the conflict?

dennybaa:
@PrasadG193 resolved

PR #550, opened 2021-11-29
fix: Failed to analyze an event whose lastTimestamp is null (#548)

### LINKED ISSUE #548: [BUG] Failed to analyze an event whose lastTimestamp is null
**Describe the bug**

When I used Botkube to send all events to ElasticSearch for analysis, I couldn't find any scheduling related events. But when I use Kubectl, there are a lot. After my investigation, I found that the structure of events related to scheduling is like this.
```
- action: Scheduling
  apiVersion: v1
  eventTime: "2021-11-29T02:56:10.265245Z"
  firstTimestamp: null
  involvedObject:
    apiVersion: v1
    kind: Pod
    name: eaea1d07-7a6b-4a64-902f-3ef8eeac21b0-x9tz9-git-clone-6858-8x54h
    namespace: montoring
    resourceVersion: "97272811"
    uid: e512eef3-8318-4a06-b3b3-6491528d30a1
  kind: Event
  lastTimestamp: null
  message: '0/20 nodes are available: 10 node(s) didn''t match Pod''s node affinity/selector,
    5 node(s) didn''t match pod affinity rules, 5 node(s) didn''t match pod affinity/anti-affinity
    rules, 5 node(s) had taint {node-role.kubernetes.io/master: }, that the pod didn''t
    tolerate.'
  metadata:
    creationTimestamp: "2021-11-29T02:56:10Z"
    name: eaea1d07-7a6b-4a64-902f-3ef8eeac21b0-x9tz9-git-clone-6858-8x54h.16bbe4e6ffd7cb0f
    namespace: montoring
    resourceVersion: "97272813"
    uid: b6455ac9-a9e7-4e21-a78f-dbc714faf7d4
  reason: FailedScheduling
  reportingComponent: default-scheduler
  reportingInstance: default-scheduler-master03
  source: {}
  type: Warning
```
I found that kubectl get events returns null on events lastTimestamp fields.However, lastTimestamp is used in the code as the TimeStamp of botkube Event.This causes us to be unable to analyze events with lastTimestamp null。

in  botkube/pkg/events/events.go
image

**To Reproduce**

In addition, a null lastTimestamp is normal. Here is a link to the community, please refer to it.https://github.com/kubernetes/kubernetes/issues/90482

**Screenshots**
image

Lion916:
##### ISSUE TYPE

 - Bug fix Pull Request

##### SUMMARY
The lastTimestamp field of the event is sometimes null and sometimes not empty, but these are normal, just because the old and new APIS are called differently. So we need to add a judgment here, using the EventTime field if lastTimestamp is empty

Fixes #548

### DIFF
--- a/pkg/events/events.go
+++ b/pkg/events/events.go
@@ -123,6 +123,10 @@ func New(object interface{}, eventType config.EventType, resource, clusterName s
 		event.Count = eventObj.Count
 		event.Action = eventObj.Action
 		event.TimeStamp = eventObj.LastTimestamp.Time
+		t := time.Time{}
+		if eventObj.LastTimestamp.Time == t {
+			event.TimeStamp = eventObj.EventTime.Time
+		}
 	}
 	return event
 }


Lion916:
I think I should offer a better plan

Lion916:
I think I should offer a better plan

pkg/events/events.go
```diff
@@ -123,6 +123,10 @@ func New(object interface{}, eventType config.EventType, resource, clusterName s
 		event.Count = eventObj.Count
 		event.Action = eventObj.Action
 		event.TimeStamp = eventObj.LastTimestamp.Time
+		t := time.Time{}
+		if eventObj.LastTimestamp.Time == t {
```
PrasadG193, line 127: @Lion916 lets use `time.IsZero()` method to check nil time
https://pkg.go.dev/time#Time.IsZero

Lion916: @PrasadG193    Could you help to review another PR? Thanks~   https://github.com/infracloudio/botkube/pull/572

PrasadG193: @Lion916 could you please make the changes suggested above? :arrow_up:

PR #558, opened 2021-12-29
Ignore  error of type not found in resource

Lion916:
##### ISSUE TYPE

 - Feature Pull Request

##### SUMMARY

After using Botkube for some time, I found a large number of "not found" errors when querying logs for troubleshooting. 
image

Following the code, it turns out that these errors are caused by the Informer cache but are actually normal. Controller-runtime does a similar thing.
https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/client/interfaces.go
```
// IgnoreNotFound returns nil on NotFound errors.
// All other values that are not NotFound errors or nil are returned unmodified.
func IgnoreNotFound(err error) error {
	if apierrors.IsNotFound(err) {
		return nil
	}
	return err
}
```

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -34,6 +34,7 @@ import (
 	log "github.com/infracloudio/botkube/pkg/log"
 
 	coreV1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metaV1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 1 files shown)


PrasadG193 (approved):

PrasadG193:
Awesome! Thank a lot for fixing this @Lion916 :pray:

kelvinwijaya:
@PrasadG193 , Will this changes be merged to the mainstream release? It seem only avail in development branch now

PR #562, opened 2022-02-03
perf: Use `strings.EqualFold` for string comparison

iamrajiv:
## Description

`strings.ToLower/strings.ToUpper` will have to allocate memory for the new strings, as well as convert both strings fully, even if they differ on the very first byte. On the other hand, `strings.EqualFold` compares the strings one character at a time. It doesn't need to create two intermediate strings and can return as soon as the first non-matching character has been found.


PrasadG193 (approved):

PrasadG193:
Thanks for the PR @iamrajiv :pray:

PR #567, opened 2022-02-20
Replace latest tag with v9.99.9-dev in builds

PrasadG193:
Signed-off-by: Prasad Ghangal <prasad.ghangal@gmail.com>

##### ISSUE TYPE
 - Bug fix Pull Request

##### SUMMARY

The latest goreleaser doesn't support non-semver image tags. 
This PR replaces `latest` tags with `v9.99.9-dev` to support latest version of goreleaser

### DIFF
--- a/.github/workflows/artifacts.yml
+++ b/.github/workflows/artifacts.yml
@@ -30,8 +30,7 @@ jobs:
       uses: goreleaser/goreleaser-action@v2
       with:
         install-only: true
-        # Newer goreleaser version doesn't allow builds with "latest" tag
-        version: v0.179.0
+        version: latest
     - name: Run GoReleaser
       run: make release-snapshot
     - name: Install Helm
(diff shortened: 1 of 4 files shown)


.github/workflows/artifacts.yml
```diff
@@ -30,8 +30,7 @@ jobs:
       uses: goreleaser/goreleaser-action@v2
       with:
         install-only: true
-        # Newer goreleaser version doesn't allow builds with "latest" tag
-        version: v0.179.0
+        version: latest
```
sanketsudake, line 33: I am bit confused on this change. Everywhere else we are using `v9.99.9-dev` version where as here we are using latest version.

Also, we have removed the earlier comments.
```
# Newer goreleaser version doesn't allow builds with "latest" tag
```
So Goreleaser accepts latest version now ?

PrasadG193: @sanketsudake the version we have mentioned here is the Goreleaser version we use in CI - not the version we build images with.
Goreleaser doesn't support image builds with `latest` tag, that's why we have replaced tag `latest` with `v9.99.9-dev` tags for BotKube build images.

sanketsudake (approved):

PR #573, opened 2022-04-13
fix: Failed to analyze an event whose lastTimestamp is null

Lion916:
##### ISSUE TYPE
- Bug fix Pull Request
##### SUMMARY
The lastTimestamp field of the event is sometimes null and sometimes not empty, but these are normal, just because the old and new APIS are called differently. So we need to add a judgment here, using the EventTime field if lastTimestamp is empty

Fixes https://github.com/infracloudio/botkube/issues/548

### DIFF
--- a/pkg/events/events.go
+++ b/pkg/events/events.go
@@ -123,6 +123,11 @@ func New(object interface{}, eventType config.EventType, resource, clusterName s
 		event.Count = eventObj.Count
 		event.Action = eventObj.Action
 		event.TimeStamp = eventObj.LastTimestamp.Time
+		// Compatible with events.k8s.io/v1
+		if eventObj.LastTimestamp.IsZero() {
+			event.TimeStamp = eventObj.Series.LastObservedTime.Time
+			event.Count = eventObj.Series.Count
+		}
 	}
 	return event
 }


Lion916:
@PrasadG193  Could you help to review another PR? Thanks~

pkg/events/events.go
```diff
@@ -123,6 +123,11 @@ func New(object interface{}, eventType config.EventType, resource, clusterName s
 		event.Count = eventObj.Count
 		event.Action = eventObj.Action
 		event.TimeStamp = eventObj.LastTimestamp.Time
+		//Compatible with events.k8s.io/v1
```
PrasadG193, line 126:
```suggestion
		// Compatible with events.k8s.io/v1
```

PrasadG193 (approved):

pkg/events/events.go
PrasadG193: @Lion916 can be merged once this minor comment is addressed

PrasadG193: Also please rebase with the latest develop to fix the CI failure

Lion916: @PrasadG193 Could you help to review this PR? Thanks~

PrasadG193: Thanks for being patient. :pray:

PR #577, opened 2022-05-05
Fix [event.go 128] invalid memory address or nil pointer 

krunalhinguu:
##### ISSUE TYPE
 - Bug fix Pull Request

##### SUMMARY
Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer dereference)

event.go : line 128 
event.TimeStamp = eventObj.Series.LastObservedTime.Time

was giving nil pointer.

### DIFF
--- a/pkg/events/events.go
+++ b/pkg/events/events.go
@@ -124,7 +124,7 @@ func New(object interface{}, eventType config.EventType, resource, clusterName s
 		event.Action = eventObj.Action
 		event.TimeStamp = eventObj.LastTimestamp.Time
 		// Compatible with events.k8s.io/v1
-		if eventObj.LastTimestamp.IsZero() {
+		if eventObj.LastTimestamp.IsZero() && !eventObj.Series.LastObservedTime.IsZero() {
 			event.TimeStamp = eventObj.Series.LastObservedTime.Time
 			event.Count = eventObj.Series.Count
 		}
--- a/pkg/events/events.go
+++ b/pkg/events/events.go
@@ -110,6 +110,7 @@ func New(object interface{}, eventType config.EventType, resource, clusterName s
 
 	if objectTypeMeta.Kind == "Event" {
 		var eventObj coreV1.Event
+		var eventSeriesObj coreV1.EventSeries
 		err := utils.TransformIntoTypedObject(object.(*unstructured.Unstructured), &eventObj)
 		if err != nil {
 			log.Errorf("Unable to transform object type: %v, into type: %v", reflect.TypeOf(object), reflect.TypeOf(eventObj))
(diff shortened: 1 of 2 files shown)


PrasadG193:
@krunalhinguu thanks for the PR.
Could you please give the PR an appropriate title? Please go through this guide https://cbea.ms/git-commit/

pkg/events/events.go
```diff
@@ -110,6 +110,7 @@ func New(object interface{}, eventType config.EventType, resource, clusterName s
 
 	if objectTypeMeta.Kind == "Event" {
 		var eventObj coreV1.Event
+		var eventSeriesObj coreV1.EventSeries
```
PrasadG193, line 113: Where are we setting this?

pkg/events/events.go
```diff
@@ -124,9 +125,9 @@ func New(object interface{}, eventType config.EventType, resource, clusterName s
 		event.Action = eventObj.Action
 		event.TimeStamp = eventObj.LastTimestamp.Time
 		// Compatible with events.k8s.io/v1
-		if eventObj.LastTimestamp.IsZero() {
-			event.TimeStamp = eventObj.Series.LastObservedTime.Time
-			event.Count = eventObj.Series.Count
+		if eventObj.LastTimestamp.IsZero() && !eventSeriesObj.LastObservedTime.IsZero() {
```
PrasadG193, line 128: Shouldn't we be using `eventObj.Series` instead of `eventSeriesObj`? Where are we setting `eventSeriesObj`? Is it always going to be nil?

krunalhinguu: you will get null pointer there, because series object is not part of coreV1.Event

https://pkg.go.dev/k8s.io/api/core/v1#EventSeries

https://pkg.go.dev/k8s.io/api/core/v1#Event

when we are trying to use event.series its always giving nil pointer

krunalhinguu: // Data about the Event series this event represents or nil if it's a singleton Event.
	// +optional
	Series *[EventSeries](https://pkg.go.dev/k8s.io/api/core/v1#EventSeries) `json:"series,omitempty" protobuf:"bytes,11,opt,name=series"`

pkg/events/events.go
PrasadG193: @krunalhinguu where we are setting the value of `eventSeriesObj`?
on line no 129 `event.TimeStamp = eventSeriesObj.LastObservedTime.Time` we are trying to access it. But where will `eventSeriesObj` get assigned the value?

krunalhinguu: i thought eventSeries is providing us that value?? or are we manually do that?
@PrasadG193

PrasadG193:
Raised https://github.com/infracloudio/botkube/pull/591 to fix this. @krunalhinguu PTAL

PR #604, opened 2022-06-01
Add option to build&push image on PR via label

### LINKED ISSUE #590: Build Docker image on PRs

Currently, if I want to make easier to test my changes from PR for a reviewer, I need to build my custom image and push it to Docker hub. This should be done automatically.

A GitHub Actions pipeline which builds and pushes Docker images from pull requests. Similar to https://github.com/capactio/capact/blob/main/.github/workflows/pr-build.yaml

Probably it should use `pull_request_target` event as well, to avoid overriding other images (packages) within `ghcr.io`.

mszostok:
##### ISSUE TYPE

 - Feature Pull Request

##### SUMMARY

Add an option to push the BotKube image automatically on PR. Because of   the security reasons, it's enabled only if the `build-pr-image` label is added—since external users do not have the permission to assign labels, this effectively requires repository owners to manually review the changes first.

Here is recorded demo https://www.loom.com/share/2f0fdc621a144d32861d83946bfff1d7 and repository on which I tested that change: https://github.com/mszostok/botkube/pulls

This PR will solve the problem with manual PR builds, e.g. we had that issue here:
- https://github.com/infracloudio/botkube/pull/601
- https://github.com/infracloudio/botkube/pull/593
- https://github.com/infracloudio/botkube/pull/582
- https://github.com/infracloudio/botkube/pull/583

Fixes #590 

#### After merge

A new label `build-pr-image` needs to be created.

#### Further improvements

1. Add `build-pr-image` automatically, for all trusted members _(org collaborators)_. As a result, we don't need to remember about that.
2. Add a comment about successfully pushed images directly under PR. It just simplifies the UX by removing one more mouse click.

#### Alternative solution

The alternative doesn't require manual action. However, it's more complicated.

See: https://github.com/infracloudio/botkube/compare/develop...mszostok:build-images-on-prs-v2?expand=1

To ensure that secrets won't be available for untrusted code, first we need to build the image and share it with the second job, which doesn't check out the untrusted code and can safely push an artifact to ghcr.io.

#### Security

This article describes it well: https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

### DIFF
--- a/.github/workflows/pr-image.yaml
+++ b/.github/workflows/pr-image.yaml
@@ -1,0 +1,72 @@
+name: Publish BotKube image
+
+on:
+  pull_request_target:
+    types: [ opened, synchronize, reopened, labeled ]
+    branches:
+      - "develop"
+
+env:
+  GO_VERSION: 1.18
+
+jobs:
+  publish:
+    name: Build and push
+    runs-on: ubuntu-latest
+
+    # Since external users do not have the permission to assign labels,
+    # this effectively requires repository owners to manually review the changes first.
+    if: contains(github.event.pull_request.labels.*.name, 'build-pr-image')
+
+    permissions:
+      contents: read
+      packages: write
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v2
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Setup Go
+        uses: actions/setup-go@v3
+        with:
+          go-version: ${{ env.GO_VERSION }}
+          cache: true
+
+      - name: Install GoReleaser
+        uses: goreleaser/goreleaser-action@v2
+        with:
+          install-only: true
+          version: latest
+
+      - name: Run GoReleaser
+        env:
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+        run: |
+          make release_pr_snapshot
+
+      - name: Summary
+        env:
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+        run: |
+          cat > $GITHUB_STEP_SUMMARY << ENDOFFILE
+          ### BotKube image published successfully! :rocket:
+
+          To test BotKube with PR changes, run:
+
+              gh pr checkout ${PR_NUMBER}
+
+              helm install botkube -n botkube --create-namespace \\
+              --set image.repository=infracloudio/pr/botkube \\
+              --set image.tag=v${PR_NUMBER} \\
+              ./helm/botkube
+
+          ENDOFFILE
--- a/Makefile
+++ b/Makefile
@@ -61,6 +61,10 @@ gorelease:
 release-snapshot:
 	@./hack/goreleaser.sh release_snapshot
 
+# Build project and push PR images with PR_NUMBER tag
+release_pr_snapshot:
+	@./hack/goreleaser.sh release_pr_snapshot
+
 # system checks
 system-check:
 	@echo "Checking system information"
--- a/hack/goreleaser.sh
+++ b/hack/goreleaser.sh
@@ -43,6 +43,36 @@ release_snapshot() {
   docker manifest push ghcr.io/infracloudio/botkube:${GORELEASER_CURRENT_TAG}
 }
 
+release_pr_snapshot() {
+  prepare
+
+  if [ -z "${PR_NUMBER}" ]
+  then
+    echo "Missing PR_NUMBER."
+    exit 1
+  fi
+
+  export GORELEASER_CURRENT_TAG=v${PR_NUMBER}
+  goreleaser release --rm-dist --snapshot --skip-publish
+  # Re-tag with 'pr' prefix
+  docker tag ghcr.io/infracloudio/botkube:${GORELEASER_CURRENT_TAG}-amd64 ghcr.io/infracloudio/pr/botkube:${GORELEASER_CURRENT_TAG}-amd64
+  docker tag ghcr.io/infracloudio/botkube:${GORELEASER_CURRENT_TAG}-arm64 ghcr.io/infracloudio/pr/botkube:${GORELEASER_CURRENT_TAG}-arm64
+  docker tag ghcr.io/infracloudio/botkube:${GORELEASER_CURRENT_TAG}-armv7 ghcr.io/infracloudio/pr/botkube:${GORELEASER_CURRENT_TAG}-armv7
+
+
+  # Push images
+  docker push ghcr.io/infracloudio/pr/botkube:${GORELEASER_CURRENT_TAG}-amd64
+  docker push ghcr.io/infracloudio/pr/botkube:${GORELEASER_CURRENT_TAG}-arm64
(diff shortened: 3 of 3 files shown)


pkosiec (commented):
Nice PR with a demo recording and nice that you prepared the alternative approach 👌 

Honestly, I'm not a fan of the idea to run the job on label as it requires manual action 🙂 I know we can also automate adding such label, but if we can prevent repushing images without needing reviewer to spot some security issues, without manual action, then why wouldn't we want to go for it? (I mean the alternative approach).

My biggest concern, apart from having the manual action, is that the reviewer is responsible for making sure the PR can be safely built (he/she needs to spot any overrides in Bash scripts, Makefile etc.). With the alternative approach, there's simply no option to mess up with the Packages as pushing is isolated from user.

You wrote the alternative approach is more complex (it saves Docker image and adds another job which pushes the image), but from the projects we worked on in the past, I think this wasn't the problematic part which caused some issues. The most problematic thing was the `pull_request_target` event (sometimes unexpected) behavior and that's something we still have here in this approach. Also, this is quite questionable whether "saving Docker image as an artifact and pushing it in second job" is more complex than "automatically adding label for org memebers/trusted external contributors" (as we would need to have it to say the solutions are on par).

Also, you said on the demo recording, that sometimes the image build is not necessary, so with that label we can avoid builds when doing e.g. documentation. That's true, but we can also ignore this job for specific paths (e.g. all MD files) automatically in GH Actions job configuration.

So personally I would pick the alternative approach as it can be even simpler in a long run. I would like to have as many things automated as possible, and currently I don't see the advantages of the building images on label (I would say otherwise, now I see even more benefits of the alternative approach). But of course I'm open for further discussion 🙂

mszostok:
Hi!

> You wrote the alternative approach is more complex

yes, not only based on code, but also based on the flow. In the `label` example, you just have a `make release_pr_snapshot` and that's all. This script is easy and doesn't require maintaining and understand more complex workflow. In the alternative solution, you need to play with the separated jobs, artifacts, and ensure that the `save`, `load` and `push` are in sync. 


> My biggest concern, apart from having the manual action, is that the reviewer is responsible for making sure the PR can be safely built 

yes, and that's also a good reminder, because even if we will apply the alternative solution, we still **need to pay attention to it.** It's really important to do not remember about that, if you will approve a PR that where you didn't "(..)spot any overrides in Bash scripts, Makefile"  then we are in the same bad position. Hiding that, and forgetting about it, is not a save option too.

>we can also ignore this job for specific paths

yes, but from our experience, we also know that it's sometimes hard to keep the list of files up to date.

----

I will finish the alternative solution and will test that on my fork. I will keep you posted.

mszostok:
I tested the alternative solution and here is a working version: https://github.com/infracloudio/botkube/compare/develop...mszostok:build-images-on-prs-v2?expand=1

Here is GitHub job execution: https://github.com/mszostok/botkube/runs/6714112689?check_suite_focus=true

I'm still voting for the first option, as described in the comment above. After testing the alternative solution, I'm also worried about the artifact size. It's 272 MB, see: https://github.com/mszostok/botkube/actions/runs/2429967388

I added `if: always()` to the delete step, to ensure that even if sth will fail it will be deleted.
```yaml
      - name: Delete Docker image artifact
        uses: geekyeggo/delete-artifact@v1
        if: always()
        with:
          name: ${{ matrix.APP }}-${{github.sha}}
```

I'm not able to check what's the limit here, but it looks like 500 MB: https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions

so it means that no more than 2 concurrent builds cannot be scheduled. It could become problematic in the long run.

pkosiec:
> yes, and that's also a good reminder, because even if we will apply the alternative solution, we still need to pay attention to it. It's really important to do not remember about that, if you will approve a PR that where you didn't "(..)spot any overrides in Bash scripts, Makefile" then we are in the same bad position. Hiding that, and forgetting about it, is not a save option too.

We cannot merge something without review, that's obvious. I was referring to one huge (at least for me) difference: In the alternative approach we can run this job safely before review (which can usually save time spent on review as there might be some build issues which contributor can fix before the review). With the label-based approach we need to review PR every time before running build, which increases time for review. Correct?

When it comes to the artifact size, I think we can think about building multiarch image only on `develop`, and for PRs just `amd64`, but we can take care of that later.

I'm sorry, but personally I still don't see any advantages in the label-based approach over the alternative solution. We can discuss it offline though. I would be happy to review the PR with alternative approach. BTW `matrix.APP` in your job is empty, you can see that in logs: `Artifact -ae61a6323f637ee2ed7bd778fbfa7c3252c9db66 has been successfully uploaded!`. You can fix that before opening PR 🙂

mszostok:
> With the label-based approach we need to review PR every time before running build, which increases time for review. Correct?

Nope, it's only about image. Build is triggered anyway, see: https://github.com/infracloudio/botkube/blob/7735b836e877858e7054caf027055ff421d2c627/.github/workflows/ci.yml#L42-L56

> When it comes to the artifact size, I think we can think about building multiarch image only on develop, and for PRs just amd64, but we can take care of that later.

That will be yet another difference in the setup, which personally, I don't like to introduce.

> I'm sorry, but personally I still don't see any advantages in the label-based approach over the alternative solution. 

as already mentioned:
- is more complicated in the overall workflow, see the implementation, so it will be harder to maintain
- problem with artifact upload size
- hide still a huge security risk, sth is automatically done without any action. If you forget about that, you can forget that this workflow has a security risk included and merge that into `develop`. Of course, you can do that anyway, but here we at least have a "reminder" on each PR.

> In the alternative approach we can run this job safely before review 
yes, and you also don't want to run a docker image on your computer if you don't trust it, right? So the image is not needed before review, it's always a good practice to review the code and later execute it on your computer 🙂

pkosiec:
> Nope, it's only about image.
> yes, and you also don't want to run a docker image on your computer if you don't trust it, right? So the image is not needed before review, it's always a good practice to review the code and later execute it on your computer 🙂

Yes, it is just about building image, but as I wrote before, we can detect some image build issues before, which saves time before actual review. I didn't say anything about running such image on my local machine before review. Moreover, at later point it would be great to have Helm chart installed on some local cluster on CI and do some true E2E tests, to unload maintainers. Which would make the E2E job depending on building image, which would require manual action.

> is more complicated in the overall workflow, see the implementation, so it will be harder to maintain

I already responded why your proposed solution is on par when it comes to complexity, or even more complex (if you want to do automatic labeling for org members and external contributors). And we have that already working.

> hide still a huge security risk, 

This is totally untrue, as an automatically ran job in the alternative solution cannot override any other Docker images. In your labeled-based approach it can, so your labeled-based approach hides a true security risk. People are always the weakest points when it comes to security. That manual action repeated many times over time won't prevent anything - it will end in the opposite. So the alternative approach brings an additional layer of security, which I am voting for.

> problem with artifact upload size

And that's first valid point which resonates with me, however firstly you would need to confirm the limit for open source and if this is a limit per job or not. We need to see why this is so big (even if this is multiarch), and even if we split some components to multiple binaries, I don't think we will split Docker images, maybe we will do something like istiod - let's see. So I wouldn't treat that as a blocker as we need to see how BotKube evolves and how many optimizations we can do first.


I'm still voting for a better security and more automation - that is the alternative approach.

mszostok:
To sum up, I see pros and cons in both approaches, and I'm OK-ish with both of them.

As discussed f2f with Paweł, we would like to ask you @PrasadG193 to check both approaches and let us know about your thoughts 👍 Maybe you know some alternative approach?

mszostok:
Close in favor of #611

PR #611, opened 2022-06-06
Add option to build&push image on PR automatically

### LINKED ISSUE #590: Build Docker image on PRs

mszostok:
##### ISSUE TYPE

 - Feature Pull Request

##### SUMMARY

Add an option to push the BotKube image automatically on PR. It's alternative approach for https://github.com/infracloudio/botkube/pull/604.

This PR will solve the problem with manual PR builds, e.g. we had that issue here:
- https://github.com/infracloudio/botkube/pull/601
- https://github.com/infracloudio/botkube/pull/593
- https://github.com/infracloudio/botkube/pull/582
- https://github.com/infracloudio/botkube/pull/583

Example run: https://github.com/mszostok/botkube/runs/6714112689?check_suite_focus=true

Fixes #590 

To ensure that secrets won't be available for untrusted code, first we need to build the image and share it with the second job, which doesn't check out the untrusted code and can safely push an artifact to ghcr.io.

The flow is as follows:
```
Job1: image build -> image save -> artifact upload 
Job2: artifact download -> image load -> image push
```
Job1—runs untrusted code but without write repo perms
Job2—push built image with package write perms

#### Security

This article describes it well: https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

### DIFF
--- a/.github/workflows/pr-image.yaml
+++ b/.github/workflows/pr-image.yaml
@@ -5,6 +5,25 @@ on:
     types: [ opened, synchronize, reopened ]
     branches:
       - "develop"
+    paths-ignore:
+      - 'branding/**'
+      - 'design/**'
+      - 'helm/**'
+      - '.codespell-whitelist'
+      - '.gitignore'
+      - '.golangci.yml'
+      - '.mergify.yml'
+      - 'botkube-title.jpg'
+      - 'botkube_arch.jpg'
+      - 'CHANGELOG.md'
+      - 'CODE_OF_CONDUCT.md'
+      - 'comm_config.yaml'
+      - 'CONTRIBUTING.md'
+      - 'deploy-all-in-one.yaml'
+      - 'deploy-all-in-one-tls.yaml'
+      - 'LICENSE'
+      - 'README.md'
+      - 'resource_config.yaml'
 
 env:
   GO_VERSION: 1.18
@@ -47,7 +66,7 @@ jobs:
         uses: actions/upload-artifact@v2
         with:
           name: ${{ matrix.APP }}-${{github.sha}}
-          path: /tmp/
+          path: /tmp/*.tar
           retention-days: 1
 
   push-image:
--- a/.github/workflows/pr-image.yaml
+++ b/.github/workflows/pr-image.yaml
@@ -1,0 +1,106 @@
+name: Publish BotKube image
+
+on:
+  pull_request_target:
+    types: [ opened, synchronize, reopened ]
+    branches:
+      - "develop"
+
+env:
+  GO_VERSION: 1.18
+
+jobs:
+
+  save-image:
+    name: Build and save BotKube image
+    runs-on: ubuntu-latest
+
+    permissions:
+      contents: read
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+          persist-credentials: false
+
+      - name: Setup Go
+        uses: actions/setup-go@v3
+        with:
+          go-version: ${{ env.GO_VERSION }}
+          cache: true
+
+      - name: Install GoReleaser
+        uses: goreleaser/goreleaser-action@v2
+        with:
+          install-only: true
+          version: latest
+
+      - name: Save
+        env:
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+        run: |
+          make save-pr-image
+
+      - name: Upload artifact
+        uses: actions/upload-artifact@v2
+        with:
+          name: ${{ matrix.APP }}-${{github.sha}}
+          path: /tmp/
+          retention-days: 1
+
+  push-image:
+    name: Push BotKube image
+    runs-on: ubuntu-latest
+    needs: [save-image]
+
+    permissions:
+      contents: read
+      packages: write
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+        with:
+          persist-credentials: false
+
+      - name: Download artifact
+        uses: actions/download-artifact@v2
+        with:
+          name: ${{ matrix.APP }}-${{github.sha}}
+          path: /tmp
+
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v2
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Push
+        env:
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+        run: make push-pr-image
+
+      - name: Delete Docker image artifact
+        uses: geekyeggo/delete-artifact@v1
+        with:
+          name: ${{ matrix.APP }}-${{github.sha}}
+
+      - name: Summary
+        env:
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+        run: |
+          cat > $GITHUB_STEP_SUMMARY << ENDOFFILE
+
+          ### BotKube image published successfully! :rocket:
+          To test BotKube with PR changes, run:
+
+              gh pr checkout ${PR_NUMBER}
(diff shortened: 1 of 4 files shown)


hack/goreleaser.sh
```diff
@@ -43,6 +43,57 @@ release_snapshot() {
...
+
+  export GORELEASER_CURRENT_TAG=v${PR_NUMBER}
+  goreleaser release --rm-dist --snapshot --skip-publish
+
+  # Re-tag with 'pr' prefix
+  docker tag ghcr.io/infracloudio/botkube:${GORELEASER_CURRENT_TAG}-amd64 ghcr.io/infracloudio/pr/botkube:${GORELEASER_CURRENT_TAG}-amd64
+  docker tag ghcr.io/infracloudio/botkube:${GORELEASER_CURRENT_TAG}-arm64 ghcr.io/infracloudio/pr/botkube:${GORELEASER_CURRENT_TAG}-arm64
+  docker tag ghcr.io/infracloudio/botkube:${GORELEASER_CURRENT_TAG}-armv7 ghcr.io/infracloudio/pr/botkube:${GORELEASER_CURRENT_TAG}-armv7
+
+  # Push images
```
pkosiec, line 63: BTW It's sad that you cannot do `docker save/load for the multiarch image 😞)

.github/workflows/pr-image.yaml
```diff
@@ -0,0 +1,126 @@
...
+
+      - name: Delete Docker image artifact
+        uses: geekyeggo/delete-artifact@v1
+        if: always()
+        with:
+          name: ${{ matrix.APP }}-${{github.sha}}
+
+      - name: Summary
+        env:
+          PR_NUMBER: ${{ github.event.pull_request.number }}
```
pkosiec, line 113: This is defined few times in container steps, let's move it to global envs for all jobs defined in this file (`env.PR_NUMBER`)

pkosiec (commented):
Just a few comments 🙂

hack/goreleaser.sh
```diff
@@ -43,6 +43,57 @@ release_snapshot() {
   docker manifest push ghcr.io/infracloudio/botkube:${GORELEASER_CURRENT_TAG}
 }
 
+save_pr_image() {
```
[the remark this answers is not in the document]
mszostok: but why `load`? not sure if I follow.

>  It shouldn't be a big effort, and we could use the save() command for preparing tar archives and attach them to the GH release 🤔

Let's address that on a separate issue/pr related to GitHub release enhancement. As this would require pipelines and changes (e.g. mention how to use the “air gapped” option).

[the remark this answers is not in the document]
pkosiec: Sorry, I meant `save_image` and `load_image`

> Let's address that on a separate issue/pr related to GitHub release enhancement. 

Fine for me 👍

.github/workflows/pr-image.yaml
```diff
@@ -0,0 +1,121 @@
...
+          version: latest
+
+      - name: Save
+        run: |
+          make save-pr-image
+
+      - name: Upload artifact
+        uses: actions/upload-artifact@v2
+        with:
+          name: ${{ matrix.APP }}-${{github.sha}}
```
pkosiec, line 67: There are still 2 occurrences of `matrix.APP` - here:
```suggestion
          name: botkube-${{github.sha}}
```

.github/workflows/pr-image.yaml
```diff
@@ -0,0 +1,121 @@
...
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+        with:
+          persist-credentials: false
+
+      - name: Download artifact
+        uses: actions/download-artifact@v2
+        with:
+          name: ${{ matrix.APP }}-${{github.sha}}
```
pkosiec, line 89: ...and here:
```suggestion
          name: botkue-${{github.sha}}
```

pkosiec (commented):
Please apply these two small comments and run the job again on your fork, just to confirm it works. We'll be ready to merge this 👍

hack/goreleaser.sh
```diff
@@ -22,6 +22,9 @@
 set -o errexit
 set -o pipefail
 
+IMAGE_REGISTRY="${IMAGE_REGISTRY:-ghcr.io/infracloudio}"
```
PrasadG193, line 25: I am not very comfortable with pushing images to `ghcr.io/infracloudio` on every PR.
How are we going to safeguard this against the abuse? People can raise fake PRs with the intent to steal sensitive info like GitHub creds in this case.
Ideally, I would have liked it if the image gets pushed to the PR authors repo rather than `ghcr.io/infracloudio/botkube`

mszostok: there is such threat but it was mitigated with the usage of:

- permissions
  ```yaml
      permissions:
        contents: read
  ```
- and stopping creds injection
  ```yaml
      steps:
        - name: Checkout code
          uses: actions/checkout@v2
          with:
            ref: ${{ github.event.pull_request.head.sha }}
            persist-credentials: false
  ```

In this way, the `save-image`  job which checks out untrusted source code doesn't have any creds.

The `push-image` has necessary perms, but it checks out code from the `develop` branch, which is the trusted one.

Here is also the alternative approach: https://github.com/infracloudio/botkube/pull/604

There is also a discussion which one is better.

PrasadG193: > In this way, the save-image job which checks out untrusted source code doesn't have any creds.

@mszostok Does that mean the image push won't happen for PRs from the forked repo? What does it mean by untrusted source code?

mszostok: Here is a short introduction to the problem that I recorded previously: https://www.loom.com/share/2f0fdc621a144d32861d83946bfff1d7

In general, we have two options:
- push the image only when there is a given label on a PR, see: https://github.com/infracloudio/botkube/pull/604
- push the image automatically for each new PR, see this PR

Both of those approaches:
1. Build the container image with the changes added on a PR.
2. Push the container image to `ghcr.io/infracloudio/pr`.  


> What does it mean by untrusted source code?

It's a code that was changed/added by contributor on a PR.

PrasadG193: the label-based approach sounds good to me - As it would only build images only when someone needs them. And repo collaborators also have some kind of control over it

pkosiec: @PrasadG193 It would be good to run true E2E tests for each PR, which could make the automatic approach more convenient. I'm working on E2E tests which installs BotKube Helm chart on a local K3d cluster, and then we can run tests against a given test Slack workspace.

pkosiec: @PrasadG193 It would be good to run true E2E tests for each PR (that use the built Docker images), which could make the automatic approach more convenient. I'm working on E2E tests which installs BotKube Helm chart on a local K3d cluster, and then we can run tests against a given test Slack workspace. I think we should consider also that point as well. It would be great if you could read the discussion in https://github.com/infracloudio/botkube/pull/604 and see my some concerns. Generally the automatic approach is safer and less error-prone, and it could be run before reviewing PR (there is no way to override PR images from an open PR). In the label approac,  a person is responsible for labeling the PR after code review - as there might be some changes which overrides current images even before merging the PR.

pkosiec:
Hi @mszostok,

As #615 will require that job with image build, maybe let's merge this? Generally it's LGTM from my side, but I would like to see it running on fork with the applied improvements before merge.

Even after merge, we can apply any other suggestions if necessary 👍  We will refine the solution over time anyway.
Thanks!

PR #612, opened 2022-06-06
fix: Allow users to disable ANSI color in logging

### LINKED ISSUE #477: Allow users to disable ANSI color in logging

sumit-tembe:
Signed-off-by: Sumit Tembe <sumit.tembe@outlook.com>

##### ISSUE TYPE
 - Bug fix Pull Request

Fixes #477


helm/botkube/values.yaml
```diff
@@ -46,6 +46,8 @@ kubeconfig:
 
 # set one of the log levels- info, warn, debug, error, fatal, panic
 logLevel: info
+# set to true to enable ANSI color in logging
+logForceColors: false
```
pkosiec, line 50: Please also update Readme (currently it is done manually): https://github.com/infracloudio/botkube/blob/develop/helm/botkube/README.md

cmd/botkube/main.go
```diff
@@ -51,6 +52,7 @@ type Config struct {
 	ConfigPath            string        `envconfig:"optional"`
 	InformersResyncPeriod time.Duration `envconfig:"default=30m"`
 	KubeconfigPath        string        `envconfig:"optional,KUBECONFIG"`
+	LogForceColors        string        `envconfig:"optional,LOG_FORCE_COLORS"`
```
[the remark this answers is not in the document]
pkosiec: tenor-31552715

helm/botkube/values.yaml
sumit-tembe: Sure

sumit-tembe:
@pkosiec 
Yes, you are right. Earlier we had the below line in code, which was causing this issue.
`log.Formatter = &logrus.TextFormatter{ForceColors: true, FullTimestamp: true}`
So https://github.com/infracloudio/botkube/issues/477 is already fixed. We can close this PR unless we want to allow users to enable colors.

Testing newly added code by adding UT on my local machine and here is the output.
Screenshot 2022-06-06 at 9 12 03 PM

pkosiec:
Hey @sumit-tembe,

I'm sorry, but I'm a bit lost. The line which forces log colors is still here: https://github.com/infracloudio/botkube/blob/develop/cmd/botkube/main.go#L211

I believe this PR is not about enabling colors, but disabling them, right? This was the original issue: https://github.com/infracloudio/botkube/issues/477

And that's something still not possible - at least I wasn't able to force disable colors (regardless my terminal settings), so maybe we should:
- unset ForceColors (set it to the default value, `false`) (as logrus will detect if the terminal supports colors or not)
- have `LOG_DISABLE_COLORS` env variable

cmd/botkube/main.go
```diff
@@ -207,7 +208,7 @@ func newLogger(logLevelStr string) *logrus.Logger {
 		logLevel = logrus.InfoLevel
 	}
 	logger.SetLevel(logLevel)
-	logger.Formatter = &logrus.TextFormatter{ForceColors: true, FullTimestamp: true}
+	logger.Formatter = &logrus.TextFormatter{FullTimestamp: true, DisableColors: logDisableColors}
```
PrasadG193, line 211: What is the default value of `ForceColors`? In case if `logDisableColors=false` do we need to set `ForceColors: true`?

pkosiec: Default for `ForceColors` is `false`. I'm not sure if we need to set `ForceColors` at all, because as the comment says: "Set to true to bypass checking for a TTY before outputting colors.". I think we shouldn't bypass the check 🤔 
Screenshot 2022-06-07 at 14 02 31

pkosiec (approved):
Tested and it works fine 👍 @PrasadG193 what do you think? Waiting for your feedback before merging this 🙂

pkosiec:
@PrasadG193 Can we merge this?

PrasadG193 (approved):

PR #622, opened 2022-06-20
Implement E2E tests against BotKube deployed on K8s and real Slack API

pkosiec:
##### ISSUE TYPE

 - Feature Pull Request

##### SUMMARY

- Implement E2E tests against BotKube deployed on K8s and real Slack API
- Allow building Docker image
- Ensure output from FilterEngine and Executor are predictable (same order every map iteration)
- Change missing TLS secret from Recommendation to Warning
- Add happy path unit test for Ingress validation filter (to make sure we will be able to remove current E2E tests with fake Slack and K8s in near future as the coverage is equal or higher)

This PR will be followed-up by a dedicated PR which run these tests on CI (with `helm test`).

See https://github.com/infracloudio/botkube/issues/615

##### TESTING

1. Check out this PR
1. See the testing instruction in the `./tests` package.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@ IMAGE_REPO=ghcr.io/infracloudio/botkube
 TAG=$(shell cut -d'=' -f2- .release)
 
 .DEFAULT_GOAL := build
-.PHONY: release git-tag check-git-status build pre-build publish lint lint-fix test system-check
+.PHONY: release git-tag check-git-status container-image container-image-e2e-test test test-integration build pre-build publish lint lint-fix system-check
 
 # Show this help.
 help:
@@ -41,6 +41,9 @@ test: system-check
 	@echo "Starting unit and integration tests"
 	@go test -v ./...
 
+test-integration: system-check
+	@go test -v -tags=integration -race -count=1 ./tests/...
+
 # Build the binary
 build: pre-build
 	@cd cmd/botkube;GOOS_VAL=$(shell go env GOOS) CGO_ENABLED=0 GOARCH_VAL=$(shell go env GOARCH) go build -o $(shell go env GOPATH)/bin/botkube
@@ -51,6 +54,13 @@ container-image: pre-build
 	@echo "Building docker image"
 	@./hack/goreleaser.sh build
 	@echo "Docker image build successfully"
+
+# Build image for E2E tests
+container-image-e2e-test:
+	$(eval TEST_NAME := "e2e")
+	$(eval IMAGE_SUFFIX := "$(TEST_NAME)-test")
+	$(eval DOCKER_TAG := "latest")
+	docker build -f ./test.Dockerfile --build-arg TEST_NAME=$(TEST_NAME) -t $(IMAGE_REPO)-$(IMAGE_SUFFIX):$(DOCKER_TAG) .
 
 # Publish release using goreleaser
 gorelease:
--- a/test/doc.go
+++ b/test/doc.go
@@ -1,0 +1,6 @@
+// Package test contains integration tests with fake Slack and Kubernetes APIs.
+//
+// Deprecated: This package is deprecated in favor of `../tests` package
+// which contains real integration tests against real Slack and BotKube installation.
+// Eventually, the fake E2E tests will be converted in unit tests and removed completely.
+package test
--- a/tests/README.md
+++ b/tests/README.md
@@ -10,6 +10,8 @@ This directory contains E2E tests which are run against BotKube installed on Kub
 
 ### Configure Tester Slack application
 
+> **NOTE:** This is something you need to do only once. Once the tester app is configured, you can use its token for running integration tests as many times as you want.
+
 1. Create new Slack Application on [this page](https://api.slack.com/apps)
 2. Use "From an app manifest" option
 3. Copy and paste this manifest:
@@ -38,6 +40,12 @@ This directory contains E2E tests which are run against BotKube installed on Kub
 4. Install this app into your workspace
 5. Navigate to the **OAuth & Permissions** section
 6. Copy the **Bot User OAuth Token** and save it for later.
+
+  You can already export it as environment variable for [running the tests locally](#run-tests-locally):
+
+  ```bash
+  export SLACK_TESTER_APP_TOKEN="{BotKube tester app token}
+  ```
 
 ## Run tests locally
 
@@ -78,19 +86,17 @@ To run the tests manually against the latest development version, follow these s
     extraAnnotations:
       botkube.io/disable: "true"
     image:
-      registry: docker.io
-      repository: pkosiec/botkube 
-      tag: e2e-tests
+      tag: v9.99.9-dev
     ENDOFFILE
     
     helm install botkube --namespace botkube ./helm/botkube -f /tmp/values.yaml --wait
     ```
 
 
-1. Export required environment variables
+1. Export required environment variables:
 
     ```bash
-    export SLACK_TESTER_APP_TOKEN="{BotKube tester app token}" # WARNING: This is a token for Tester, not the BotKube Slack bot!
+    export SLACK_TESTER_APP_TOKEN="{BotKube tester app token}" # WARNING: This is a token for Tester, not the BotKube Slack bot.
     export KUBECONFIG=/Users/$USER/.kube/config # set custom path if necessary
     ```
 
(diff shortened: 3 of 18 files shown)


tests/README.md
```diff
@@ -0,0 +1,93 @@
+# E2E Tests
```
pkosiec, line 1: Currently I named this package as `tests` but I like the singular form more. That's why once we delete current `test` directory, I will rename it again to `test`. For now I wanted to distinguish the difference between the packages and didn't want to mix the files.

But of course I'm also open for different suggestions how to approach this 🙂

test/doc.go
```diff
@@ -0,0 +1,6 @@
+// Package test contains integration tests with fake Slack and Kubernetes APIs.
```
pkosiec, line 1: This package cannot be deleted until the true E2E tests from this PR run on CI. 
I would like to run these newly created tests both on PR and `develop` branch. As the BotKube Tester has really low permissions, I don't see any security issue in such approach. Of course, we will have a separate test workspace configured only for running these tests.

tests/e2e/slack_test.go
```diff
@@ -0,0 +1,366 @@
...
+	MessageWaitTimeout       time.Duration `envconfig:"default=10s"`
+}
+
+const (
+	channelNamePrefix = "test"
+	welcomeText       = "Let the tests begin 🤞"
+	pollInterval      = time.Second
+)
+
+func TestSlack(t *testing.T) {
```
pkosiec, line 55: Initially I wanted to go with `TestMain` approach to prepare test setup, but the code would be a bit more complicated (error handling) and also I'm not really sure if that would be valid. If we added new tests e.g. for Discord bot etc., we would extract the test setup part for Slack to different location anyway.

So I went with a simple approach (IMHO the most readable one) of running subtests, but I'm open to other approaches (creating test suite etc.).

pkosiec (commented):
Few notes for reviewer from my side 🙂

Makefile
```diff
@@ -52,6 +55,13 @@ container-image: pre-build
 	@./hack/goreleaser.sh build
 	@echo "Docker image build successfully"
 
+# Build image for E2E tests
+container-image-e2e-test:
```
pkosiec, line 59: I switched to use Goreleaser for producing test image in the follow-up PR: https://github.com/infracloudio/botkube/pull/624. So you can basically ignore that part when reviewing as this is something temporary, for local dev purposes 🙂

tests/e2e/k8s_helpers_test.go
```diff
@@ -0,0 +1,172 @@
...
+		return containerIdx, false
+	}
+
+	containers := deployment.Spec.Template.Spec.Containers
+	for i, container := range containers {
+		if container.Name != containerName {
+			continue
+		}
+
+		containerIdx = i
```
mszostok, line 125: why not to return?

tests/e2e/k8s_helpers_test.go
```diff
@@ -0,0 +1,172 @@
...
+		if err == wait.ErrWaitTimeout {
+			return lastErr
+		}
+		return err
+	}
+
+	return nil
+}
+
+func findContainerIdxByName(deployment *appsv1.Deployment, containerName string) (int, bool) {
```
mszostok, line 112: IMO it can be simplified to:
```go
func findContainerIdxByName(deployment *appsv1.Deployment, containerName string) int {
	const notFound = -1
	if deployment == nil {
		return notFound
	}

	containers := deployment.Spec.Template.Spec.Containers
	for i, container := range containers {
		if container.Name == containerName {
			return i
		}
	}

	return notFound
}
```
or if you still want to have bool indicator:

```go
func findContainerIdxByName(deployment *appsv1.Deployment, containerName string) (int, bool) {
	notFound := func() (int, bool) {
		return -1, false
	}
	if deployment == nil {
		return notFound()
	}

	containers := deployment.Spec.Template.Spec.Containers
	for i, container := range containers {
		if container.Name == containerName {
			return i, true
		}
	}

	return notFound()
}
// or
func findContainerIdxByName(deployment *appsv1.Deployment, containerName string) (int, bool) {
	if deployment == nil {
		return -1, false
	}

	containers := deployment.Spec.Template.Spec.Containers
	for i, container := range containers {
		if container.Name == containerName {
			return i, true
		}
	}

	return -1, false
}
```

tests/e2e/slack_test.go
```diff
@@ -0,0 +1,416 @@
...
+	channel, cleanupChannelFn := slackTester.CreateChannel(t)
+	defer cleanupChannelFn(t)
+	slackTester.PostInitialMessage(t, channel.Name)
+	botUserID := slackTester.FindUserIDForBot(t)
+	slackTester.InviteBotToChannel(t, botUserID, channel.ID)
+
+	t.Log("Patching Deployment with test env variables...")
+	deployNsCli := k8sCli.AppsV1().Deployments(appCfg.Deployment.Namespace)
+	revertDeployFn := setTestEnvsForDeploy(t, appCfg, deployNsCli, channel.Name)
+	defer revertDeployFn(t)
```
mszostok, line 82:
```suggestion
	t.Cleanup(func() { revertDeployFn(t) })
```

tests/e2e/slack_test.go
```diff
@@ -0,0 +1,416 @@
...
+
+	t.Log("Creating K8s client...")
+	k8sConfig, err := clientcmd.BuildConfigFromFlags("", appCfg.KubeconfigPath)
+	require.NoError(t, err)
+	k8sCli, err := kubernetes.NewForConfig(k8sConfig)
+	require.NoError(t, err)
+
+	t.Log("Setting up test Slack setup...")
+	channel, cleanupChannelFn := slackTester.CreateChannel(t)
+	defer cleanupChannelFn(t)
```
mszostok, line 74:
```suggestion
	t.Cleanup(func() { cleanupChannelFn(t) })
```

tests/e2e/slack_test.go
```diff
@@ -0,0 +1,416 @@
...
+		cfgMap := &v1.ConfigMap{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      channel.Name,
+				Namespace: appCfg.Deployment.Namespace,
+			},
+		}
+		cfgMap, err = cfgMapCli.Create(context.Background(), cfgMap, metav1.CreateOptions{})
+		require.NoError(t, err)
+
+		defer cleanupCreatedCfgMapIfShould(t, cfgMapCli, cfgMap.Name, &cfgMapAlreadyDeleted)
```
mszostok, line 235:
```suggestion
		t.Cleanup(func() {
			cleanupCreatedCfgMapIfShould(t, cfgMapCli, cfgMap.Name, &cfgMapAlreadyDeleted)	
		})
```

tests/README.md
```diff
@@ -0,0 +1,101 @@
...
+          - channels:history
+    settings:
+      org_deploy_enabled: false
+      socket_mode_enabled: false
+      token_rotation_enabled: false
+    ```
+
+4. Install this app into your workspace
+5. Navigate to the **OAuth & Permissions** section
+6. Copy the **Bot User OAuth Token** and save it for later.
```
mszostok, line 40: maybe it will be better to already export it as `SLACK_TESTER_APP_TOKEN`, so there will be no confiusion

mszostok (commented):
Tested—looks great 🚀 🚀 🚀

tests/e2e/k8s_helpers_test.go
```diff
@@ -0,0 +1,172 @@
...
+			nil,
+			[]string{slackEnabledEnvName, slackChannelIDEnvName},
+		)
+
+		if len(originalEnvs) > 0 {
+			deployment.Spec.Template.Spec.Containers[containerIdx].Env = updateEnv(envs, originalEnvs, nil)
+		}
+		_, err = deployNsCli.Update(context.Background(), deployment, metav1.UpdateOptions{})
+		require.NoError(t, err)
+	}
```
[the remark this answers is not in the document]
pkosiec: > oszczędzaj pamięć gdziekolwiek jesteś #pdk

But yes, you're right 😄 Will improve that

tests/README.md
pkosiec: I can mention it here as well 👍

mszostok (approved):
LGTM 💯

PR #623, opened 2022-06-21
Add new configuration syntax proposal

mszostok:
##### ISSUE TYPE
 - Docs Pull Request

##### SUMMARY

- Add document that describes found issue with the current syntax for BotKube configuration.
- Add document that describes how to solve multichannel support
  -  with bindings concept
  - with policies/presets/profiles concept
- Add a document that describes the possible solution to configure BotKube via dedicated CustomResources (CR) instead of configuration files.

Related issues:
- https://github.com/infracloudio/botkube/issues/596


docs/investigation/2022-06-14-policies.md
```diff
@@ -0,0 +1,379 @@
+# Polices
```
pkosiec, line 1: If all these 3 documents are very related to each other (it is about improving configuration) and they was created during the same epic, WDYT about aggregating them in the `configuration` directory?

docs/investigation/2022-06-20-cfg-via-crds.md
```diff
@@ -0,0 +1,581 @@
+# Configure BotKube via CRs
+
+Created on 2022-06-20 by Mateusz Szostok ([@mszostok](https://github.com/mszostok))
+
+| Status      |
+|-------------|
+| `POSTPONED` |
```
pkosiec, line 7: If this is an investigation, maybe it should be deleted? Not sure what about `Created on...`, but I would rather delete that too. But it's up to you 🙂 

```suggestion
```

Once we know we'll go in CRDs, this investigation could be converted to an proposal 👍

docs/investigation/2022-06-20-cfg-via-crds.md
```diff
@@ -0,0 +1,581 @@
...
+    |-------------------------|-----------------------------------------------------------------------------------|-----------------------------------------|
+    | ImageTagChecker         | Checks and adds recommendation if 'latest' image tag is used for container image. | Move as recommendation notificator.     |
+    | IngressValidator        | Checks if services and tls secrets used in ingress specs are available.           | Move as recommendation notificator.     |
+    | ObjectAnnotationChecker | Checks if annotations botkube.io/* present in object specs and filters them.      | Remove it.                              |
+    | PodLabelChecker         | Checks and adds recommendations if labels are missing in the pod specs.           | Move as recommendation notificator.     |
+    | NamespaceChecker        | Checks if event belongs to blocklisted namespaces and filter them.                | Remove it. It will be per resource now. |
+    | NodeEventsChecker       | Sends notifications on node level critical events.                                | Move as K8s events notificator.         |
```
pkosiec, line 97: We can extract it to a dedicated section (`### Mutators`)

docs/investigation/2022-06-20-cfg-via-crds.md
```diff
@@ -0,0 +1,581 @@
+# Configure BotKube via CRs
```
pkosiec, line 1: General comment to all these investigation documents:
It would be great to unify terms (and also describe them, as in the proposal). If that's too much work, let's have the proposal from #626 having agreed terms (e.g. `notifier` will become something different, as long as you agree with me) and here let's just describe used terms for each document (the `notifier` could stay as it is).

docs/investigation/2022-06-20-cfg-via-crds.md
```diff
@@ -0,0 +1,581 @@
...
+  message: "connection failed: 401 Unautorized"
+  # Reflects the generation of the most recently observed change.
+  #observedGeneration:
+  # Last time the condition transitioned from one status to another.
+  #lastTransitionTime:
+```
+
+## Consequences
+
+This section described necessary changes if proposal will be accepted.
```
pkosiec, line 367: Maybe something like this:
```suggestion
This section described necessary changes if the implementation idea will be accepted and planned in the roadmap.
```

pkosiec (commented):
👌 Really good work. Just a few comments, and we'll be able to merge this 👍

pkosiec (approved):
LGTM

Issue #506, opened 2021-07-27
Collect analytics


PrasadG193:
**Is your feature request related to a problem? Please describe.**

Add a way to collect analytics that will help us deciding roadmaps for BotKube development 

> **Note**
> Edited by @pkosiec - as agreed in https://github.com/infracloudio/botkube/pull/625,  the following list declares the scope of this task.

### To do

- [x] Set up Google Analytics and Twilio Segment accounts @pkosiec 
  - Ensure proper access rights

- [x] Implement collecting analytics in BotKube @pkosiec 
  - [x] Collect analytics listed in Initial data to collect section of the proposal
      - https://github.com/kubeshop/botkube/pull/640
      - https://github.com/kubeshop/botkube/pull/644
  - [x] Use [Segment Go Library](https://segment.com/docs/connections/sources/catalog/libraries/server/go/)
  - [x] Enabled by default
  - [x] Make sure users can opt out:
    - a dedicated env variable supported by app
    - a dedicated Helm chart property
  - [x] Document changes on botkube.io: https://github.com/kubeshop/botkube-docs/pull/96
    - Update [privacy policy](https://github.com/infracloudio/botkube-docs/blob/master/content/privacy.md) and describe data we collect
    - Describe configuration - how to opt-out
      - Consider describing it also for each installation document

- [x] Implement analytics on the website @pkosiec https://github.com/kubeshop/botkube-docs/pull/90
  - Use Google Analytics 4
  - If needed: Add cookie banner (some reports say Google Analytics 4 are cookieless, but I also read on official website that `gtag.js` library uses first-party cookies, so a user consent might be also needed)
  - Document changes: Update [privacy policy](https://github.com/infracloudio/botkube-docs/blob/master/content/privacy.md) and describe data we collect 

- [ ] Communicate changes @brampling
  - Write announcement on Slack to gather feedback before/during implementation (before 0.13 release)
  - Explain the reason of introducing the analytics
  - Point to the [epic issue](https://github.com/infracloudio/botkube/issues/506) for further discussion
  - If necessary, schedule a public meeting to discuss the changes
  - Mention analytics in BotKube 0.13 GitHub release description

pkosiec:
I agree, this is a very nice idea. Few thoughts from my side:

- To investigate which solution should be used: Google Analytics, segment.io, others?
- There should be an option to opt-out during installation
- Collect how many installations are there
- Nice to have: Additional metrics: which kind of events are most used
- Update privacy policy on botkube.io

[...]

PR #625, opened 2022-06-24
Investigate analytics collection

pkosiec:
##### ISSUE TYPE
 - Docs Pull Request

##### SUMMARY

- Investigate analytics collection

This PR contains a document with a short summary of my investigation as a part of #506.

See #506

### DIFF
--- a/docs/investigation/analytics-collection.md
+++ b/docs/investigation/analytics-collection.md
@@ -1,10 +1,10 @@
-# Analytics
+# Analytics collection
 
-This document contains a summary of a short research around analytics implementation for BotKube.
+This document contains a summary of a short research around collecting anonymous analytics (a.k.a. telemetry) for BotKube.
 
 ## Goal
 
-Learn how BotKube is used in the most private and least invasive way for users.
+The main goal is learning how BotKube is used, in the most private and least invasive way for users.
 All data should be anonymized to prevent identifying users. All analytics will be used only as aggregated collection of data (in forms of some statistics), to further improve BotKube and adjust its roadmap.
 
 The change should be widely communicated to all users, and there should be an ability to opt out from collecting such analytics.
@@ -15,7 +15,7 @@ I tried to look for some software from CNCF landscape which has telemetry built-
 
 Anyway, here are a few examples I found:
 
-- [`odo`](https://github.com/redhat-developer/odo) - Segment.io ([source](https://github.com/redhat-developer/odo/blob/77d6b6df5cdd05074db8728d1aead76d3e259e25/pkg/segment/segment.go))
+- [`odo`](https://github.com/redhat-developer/odo) - Twillio Segment ([source](https://github.com/redhat-developer/odo/blob/77d6b6df5cdd05074db8728d1aead76d3e259e25/pkg/segment/segment.go))
 - [`config-syncer` (previously `kubed`)](https://github.com/kubeops/config-syncer) - Google Analytics ([source](https://github.com/kubeops/config-syncer/blob/release-0.12/vendor/kmodules.xyz/client-go/tools/cli/cli.go))
 - [`netdata`](https://github.com/netdata/netdata) - PostHog ([see docs](https://learn.netdata.cloud/docs/agent/anonymous-statistics))
 - [Testkube](https://github.com/kubeshop/testkube) - Google Analytics 4 ([source](https://github.com/kubeshop/testkube/blob/34c57dbdb9312b68910e0ad5808485292fa31313/pkg/analytics/analytics.go))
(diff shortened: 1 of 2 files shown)


brampling:
@pkosiec Here's the information collected by OpenShift if you're curious: https://docs.openshift.com/container-platform/4.10/virt/logging_events_monitoring/virt-openshift-cluster-monitoring.html#telemetry-about-telemetry_virt-openshift-cluster-monitoring

They do it through Prometheus push into Prometheus/Thanos: https://github.com/openshift/telemeter. That's probably way more complex than what is needed here, since it has to support tens of thousands of clusters.

pkosiec:
Thanks @brampling 👍 I saw the OpenShift doc before, and such amount of data collected definitely makes sense for such a big platform. For BotKube I think it's more about the application itself, rather than the environment (as BotKube is currently just a single component with different external integrations), but that can change over time.  That's why IMHO the only thing we would need to collect for now is K8s version.

I also thought about failure reporting (what OpenShift does), but users can often misconfigure the application and to know what the problem is, we would need to gather even more data, which could be problematic. I think it makes more sense for such platform as OpenShift, where the app configuration is managed (so it is already known the app should work without issues). Anyway, I can include that on the list of the minimal scope - I'm open for a discussion 🙂 

And yes, probably setting up our own infra for collecting these could be too complex for now. Once we get to the limits of Google Analytics Free, we can definitely go back to this topic 👍

docs/investigation/analytics-collection.md
```diff
@@ -0,0 +1,100 @@
...
+
+- [`odo`](https://github.com/redhat-developer/odo) - Twillio Segment ([source](https://github.com/redhat-developer/odo/blob/77d6b6df5cdd05074db8728d1aead76d3e259e25/pkg/segment/segment.go))
+- [`config-syncer` (previously `kubed`)](https://github.com/kubeops/config-syncer) - Google Analytics ([source](https://github.com/kubeops/config-syncer/blob/release-0.12/vendor/kmodules.xyz/client-go/tools/cli/cli.go))
+- [`netdata`](https://github.com/netdata/netdata) - PostHog ([see docs](https://learn.netdata.cloud/docs/agent/anonymous-statistics))
+- [Testkube](https://github.com/kubeshop/testkube) - Google Analytics 4 ([source](https://github.com/kubeshop/testkube/blob/34c57dbdb9312b68910e0ad5808485292fa31313/pkg/analytics/analytics.go))
+- Rancher backed off from this idea when it comes to k3s: [see issue](https://github.com/k3s-io/k3s/issues/834)
+
+All of the software I found that collect anonymous analytics (not only from Go/Kubernetes space) have opt-out based configuration.
+
+## Initial data to collect
```
brampling, line 26: We should indicate here how we will aggregate the data. We should aggregate it at least per-k8s cluster since we can't really identify the customer. I'm not sure how we'd do this though as I don't think there is a cluster-wide uid.

docs/investigation/analytics-collection.md
```diff
@@ -0,0 +1,100 @@
...
+
+This list contains initial data which we could collect.
+
```
mszostok, line 29: what about collecting errors? For example:
- app crashes e.g. OOMKilled - would inform us whether our app is stable enough.
- configuration errors (syntax, missing fields etc.) - would inform us whether we need to provide more automation around that (CLI, validators etc.) - currently we don't have anything to improve UX for configuration.
- notification errors - e.g. the % of failed request that were not sent to Slack.

pkosiec: > app crashes e.g. OOMKilled 

I was thinking that for the initial scope we could gather metrics only from the inside of the app. Unfortunately, that wouldn't give us information why the app is being shut down. Correct me if I'm wrong 🙂 

But anyway, good ideas, I will apply the suggestions.

docs/investigation/analytics-collection.md
```diff
@@ -0,0 +1,100 @@
...
+- Running installations
```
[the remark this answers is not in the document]
pkosiec: 🤔 Do you mean notifying Segment every time we got an event? This could be resource-expensive when receiving multiple events. We would need to think about batching them somehow 🤔  

EDIT: Sounds like it is even built-in to the Segment library - https://segment.com/docs/connections/sources/catalog/libraries/server/go/#batching. Adding it to the list then 👍

pkosiec:
Thank you @brampling and @mszostok. I pushed the improvements to this document - please take a look again. Thanks!

docs/investigation/analytics-collection.md
brampling: 👍

docs/investigation/analytics-collection.md
mszostok: AFAIK to check the OOMKilled, we will need to check the status of the Pod, that won't be hard to do, though.

mszostok (approved):
LGTM 👍

PR #626, opened 2022-06-27
Add new configuration syntax proposal

mszostok:
##### ISSUE TYPE

 - Proposal Pull Request

##### SUMMARY

Add document that describes how to solve multichannel support and configuration issues [with bindings concept](https://github.com/mszostok/botkube/blob/binding-proposal/docs/proposal/2022-06-14-bindings.md).

It is based on the investigation from https://github.com/infracloudio/botkube/pull/623

Related issue: https://github.com/infracloudio/botkube/issues/596

### DIFF
--- a/docs/proposal/2022-06-14-configuration-bindings.md
+++ b/docs/proposal/2022-06-14-configuration-bindings.md
@@ -16,6 +16,7 @@ Created on 2022-06-14 by Mateusz Szostok ([@mszostok](https://github.com/mszosto
   * [Send notifications to multiple communications platform](#send-notifications-to-multiple-communications-platform)
   * [Run executor only from a dedicated channel](#run-executor-only-from-a-dedicated-channel)
   * [Validate channel annotation](#validate-channel-annotation)
+  * [Reference other sources inside them](#reference-other-sources-inside-them)
 - [Alternatives](#alternatives)
   * [Route notifications to a given channel based on the Kubernetes Namespace](#route-notifications-to-a-given-channel-based-on-the-kubernetes-namespace-1)
     + [Annotate Namespace](#annotate-namespace)
@@ -107,9 +108,13 @@ This section describes the necessary changes in the syntax. **It's not backward 
         namespaces:
           include: ["istio-system"]
         events: ["error"]
+        updateSetting:
+          includeDiff: true
+          fields:
+            - status.phase
       - name: networking.istio.io/v1alpha3/VirtualServices
         namespaces:
-          include: ["@all"]
+          include: ["all"]
         events: ["error"]
 
     # Recommendations about the best practices
@@ -137,10 +142,13 @@ This section describes the necessary changes in the syntax. **It's not backward 
               namespaces: # override the top level Namespace
                 include: ["istio-system"]
               events: ["error"]
+              updateSetting:
+                includeDiff: true
+                fields:
+                  - status.phase
             - name: networking.istio.io/v1alpha3/VirtualServices
               # uses the default Namespace settings from top level definition.
               events: ["error"]
-
 
         # Recommendations about the best practices
         recommendations:
@@ -194,7 +202,7 @@ This section describes the necessary changes in the syntax. **It's not backward 
     </table>
 
 
-2. Communications
+3. Communications
     <table>
     <tr>
     <td> Before </td> <td> After </td>
@@ -309,7 +317,7 @@ communications: # allows to have multiple slacks, or ES configurations
 **Sources**
 
 ```yaml
-Sources:
+sources:
  - name: nodes-errors # name used for bindings
    kubernetes:
      resources:
@@ -552,6 +560,44 @@ executors:
 
 If you use the `botkube.io/channel: <channel_name>` annotation, notifications are sent to a given channel even if not authorized.
 With a new syntax, we can check if there is a matching source binding for a given channel.
+
+### Reference other sources inside them
+
+We won't add a dedicated support to define a new source called "all-errors" and embed "network-errors", "nodes-errors" along with other resources instead of re-declaring all again. However, the [YAML anchors](https://helm.sh/docs/chart_template_guide/yaml_techniques/#yaml-anchors) can be used to overcome this issue.
+
+**Sources**
+```yaml
+sources:
+  - name: nodes-errors
+    kubernetes:
+      resources: &nodes-errors
+        - name: v1/nodes
+          namespaces:
+            include: ["@all"]
+          events:
+            - error
+  - name: network-errors
+    kubernetes:
+      namespaces:
+        include: ["@all"]
+      resources: &network-errors
+        - name: v1/services
+          events:
+            - error
+        - name: networking.k8s.io/v1/ingresses
+          events:
+            - error
+  - name: all-errors
+    kubernetes:
+      namespaces:
+        include: ["@all"]
+      resources:
+        <<: *nodes-errors
+        <<: *network-errors
+          - name: v1/pod
+            events:
+              - error
+```
 
 ## Alternatives
 


docs/proposal/2022-06-14-bindings.md
```diff
@@ -0,0 +1,681 @@
...
+4. The `namespaces.include` and `namespaces.exclude` properties are added to `notifiers[].kubernetes[]`.
+5. The `resource_config.yaml` and `comm_config.yaml` are merged into one, but you can provide config multiple times. In the same way, as Helm takes the `values.yaml` file. It's up to the user how it will be split.
+6. Update documentation about configuration.
+
+### Follow-up changes
+
+1. Change selector for all Namespaces from `all` to `@all`.
+2. Add full channel/user indicator - `@` or `#`.
+3. Recommendations are merged under notifications.
+4. Filters are removed and existing one are moved under `notifiers[].{name}`.
```
pkosiec, line 502: So are they removed or not? Sorry, I don't understand this point - cause right now there's no configuration for each filter, so what is moved? 🤔

docs/proposal/2022-06-14-bindings.md
```diff
@@ -0,0 +1,681 @@
...
+5. Update `@BotKube` commands to reflect new configuration.
+6. **Optional**: Add CLI to simplify creating/updating configuration.
+
+### Syntax
+
+
```
pkosiec, line 508:
```suggestion
```

docs/proposal/2022-06-14-bindings.md
```diff
@@ -0,0 +1,681 @@
...
+    # Recommendations about the
+    # best practices for the created resource
+    recommendations: true
+
+
+
+
+
+
+
```
pkosiec, line 545:
```suggestion
```

docs/proposal/2022-06-14-bindings.md
```diff
@@ -0,0 +1,681 @@
...
+
+
+
```
pkosiec, line 535:
```suggestion
```

docs/proposal/2022-06-14-bindings.md
```diff
@@ -0,0 +1,681 @@
...
+               # uses the default Namespace settings from top level definition.
+               events:
+                 - error
+         sysdig:
+           # ..
+         kubePug:
+           # ...
+     ```
+
+### Use case: Route BotKube notifications to individual channels
```
pkosiec, line 158: We could gather all use cases under `Examples` header. WDYT?

docs/proposal/2022-06-14-bindings.md
```diff
@@ -0,0 +1,681 @@
...
+### Follow-up changes
+
+1. Change selector for all Namespaces from `all` to `@all`.
+2. Add full channel/user indicator - `@` or `#`.
+3. Recommendations are merged under notifications.
+4. Filters are removed and existing one are moved under `notifiers[].{name}`.
+5. Update `@BotKube` commands to reflect new configuration.
+6. **Optional**: Add CLI to simplify creating/updating configuration.
+
+### Syntax
```
pkosiec, line 506: I think we should move this syntax right after Overview? And maybe we can see what is duplicated between Overview and Syntax. I like the nice before/after table 👍

docs/proposal/2022-06-14-bindings.md
```diff
@@ -0,0 +1,681 @@
...
+    ```
+
+    </td>
+    </tr>
+    </table>
+
+
+## Resources
+
+- [PR that adds profiles to BotKube configuration](https://github.com/infracloudio/botkube/pull/291)
```
pkosiec, line 680: Maybe a small comment would be helpful - why it is good to mention it there 🙂  (e.g. first implementation, which was based on X, but it is not worth to invest in time, as it has limitations like Y)

pkosiec (commented):
I really like the idea 👍 Here are a few minor comments - it's more related to the document itself, rather than the idea, which I'm the fan of.

docs/proposal/2022-06-14-bindings.md
```diff
@@ -0,0 +1,681 @@
...
+      namespaces:
+        include: ["team-a"]
+      commands:
+        verbs: ["get", "logs"]
+        resources: ["Deployments", "Pods", "Services"]
+```
+
+### Others
+
+If you use the `botkube.io/channel: <channel_name>` annotation, notifications are sent to a given channel even if not authorized.
```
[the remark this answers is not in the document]
mszostok: Personally, I would remove it. However, let's leave it for now and add a metric to check how many users are using it.

docs/proposal/2022-06-14-bindings.md
mszostok: I wanted to remove this functionality completely and the existing ones renamed as notifiers with configuration added under `notifiers[].{name}`.

However, we don't have the plugin system in place, so I rephrased it and made optional. Maybe it will be better to do that in the scope of the plugin system.

----
5. **Optional**: [Filters](https://www.botkube.io/filters/) are renamed to `notifiers` and configuration is added under `notifiers[].{name}`.
    > **NOTE:** In the future, [Filters](https://www.botkube.io/filters/) should be completely removed from the BotKube and replaced with the plugin system.
----

docs/proposal/2022-06-14-bindings.md
```diff
@@ -0,0 +1,681 @@
...
+- [Support for default cluster and namespace on any Slack channels](https://github.com/infracloudio/botkube/issues/421)
+
+## Proposal
+
+### Overview
+
+This idea is to explicitly extract such configuration domains:
+- Executors, such as `kubectl`, `helm`, `istioctl` etc.
+- Notifiers, such as Kubernetes events, Sysdig, Prometheus, etc.
+- Communications, such as Slack, Discord, MS Teams etc.
```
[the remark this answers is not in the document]
mszostok: Yep, I also don't like the term `Communications`, however I don't have a better name for now. `Notifiers` term doesn't work in this scope, as we need something that will aggregate both the Bot and Sink term.

Maybe others will have some ideas about that.

pkosiec (approved):
LGTM 🚀 Again, nice job!

docs/proposal/2022-06-14-configuration-bindings.md
```diff
@@ -0,0 +1,647 @@
...
+   kubernetes:
+     resources:
+       - name: v1/nodes
+         namespaces:
+           include:
+             - @all
+         events:
+           - error
+  - name: network-errors
+    kubernetes:
```
PrasadG193, line 323: Should we allow referring to another source in a source to avoid duplicated configuration? Eg here we can have "all-errors" -And we can embed `network-errors`, `nodes-errors` along with other resources instead of redeclaring all again

docs/proposal/2022-06-14-configuration-bindings.md
```diff
@@ -0,0 +1,647 @@
...
+
+    ```
+
+    </td>
+    <td>
+
+    ```yaml
+    sources:
+      - name: default
+        kubernetes:
```
bygui86, line 130: why a "kubernetes" stanza? do you plan to support other type of sources different than k8s in the future?

docs/proposal/2022-06-14-configuration-bindings.md
```diff
@@ -0,0 +1,647 @@
...
+          commands:
+            verbs: ["api-resources", "...", "auth"]
+            resources: ["deployments", "..", "nodes"]
+    ```
+    </td>
+    </tr>
+    </table>
+
+
+2. Communications
```
bygui86, line 197: really stupid comment :) it should be point 3. Communications

docs/proposal/2022-06-14-configuration-bindings.md
```diff
@@ -0,0 +1,647 @@
...
+    ```yaml
+    sources:
+      - name: default
+        kubernetes:
+          # New 'namespace' property - allows Namespace restriction
+          # It can be overridden in the nested level.
+          namespace:
+            include: ["@all"]
+          resources:
+            - name: v1/pods
```
bygui86, line 136: what about `updateSetting` ? do you plan to keep this feature?
it's great, I don't think it would be a good idea to remove it :)

docs/proposal/2022-06-14-configuration-bindings.md
```diff
@@ -0,0 +1,647 @@
...
+          events:
+            - error
+```
+
+### Route notifications to a given channel based on the Kubernetes Namespace
+
+> **Note**
+>
+> Currently, you can send notification to non-default channel using [annotation](https://www.botkube.io/usage/#send-notification-to-non-default-channel).
+However, you need to apply `botkube.io/channel: <channel_name>` to each K8s object (Pods, Services, etc.) which is cumbersome.
```
bygui86, line 349: so you plan to move `botkube.io/channel` to namespace level, right?
what about `botkube.io/disable` ?
I suggest to keep the disable annotation on resource level, but extend it to namespace level as well.

docs/proposal/2022-06-14-configuration-bindings.md
```diff
@@ -0,0 +1,647 @@
...
+
+
+    resources:
+      - name: v1/pods
+        namespaces:
+          include: ["istio-system"]
+        events: ["error"]
+      - name: networking.istio.io/v1alpha3/VirtualServices
+        namespaces:
+          include: ["@all"]
```
bygui86, line 112: why `@all` instead of simply `all`?

bygui86 (commented):
@mszostok thanks a lot for this proposal, I love it!

docs/proposal/2022-06-14-configuration-bindings.md
mszostok: Because when we use `all`, we block an option of using such name for Namespaces. With `@all` you can see that it is a special tag based on the "API". Do you think this is not a necessary change? We thought that as we already change the API it will be good to straight this out too.

docs/proposal/2022-06-14-configuration-bindings.md
bygui86: maybe better to specify this :)

docs/proposal/2022-06-14-configuration-bindings.md
bygui86: got it thanks :)

docs/proposal/2022-06-14-configuration-bindings.md
bygui86: It's fine like you described. But what about the other annotation `botkube.io/disable`?

docs/proposal/2022-06-14-configuration-bindings.md
bygui86: wow that could be a real issue!

bygui86:
@mszostok thanks for your answers :)

Issue #592, opened 2022-05-25
Data race issues detected when running integration tests with `-race`


pkosiec:
**Describe the bug**

When running `go test -race`, it detects multiple data race conditions during E2E scenarios.  This should be fixed as such errors result in unpredictable behavior of the app.

**To Reproduce**

1. Pull latest `develop`
2. Modify `hack/runtests.sh` script and add `-race` flag to `go test` in line 31:

```bash
set -o errexit

set -o pipefail

export CONFIG_PATH="$(pwd)/test"

# Run unit and integration tests excluding dependencies
PACKAGES=$(go list ./... | grep -v '/vendor/')
for package in $PACKAGES; do
    go test -race -tags=test ${@} "$package" -v # <- here
done
```

3. Observe multiple `WARNING: DATA RACE` warnings. Because of data race conditions, the tests fail.

**Additional context**

Example log:

```
=== RUN   TestRun/command/Test_disable_notifier/Enable_notifier
==================
WARNING: DATA RACE
Write at 0x0000087b4a18 by goroutine 107:
  github.com/infracloudio/botkube/pkg/execute.(*DefaultExecutor).runNotifierCommand()
      /Users/pkosiec/repositories/botkube/pkg/execute/executor.go:321 +0x551
  github.com/infracloudio/botkube/pkg/execute.(*DefaultExecutor).Execute()
      /Users/pkosiec/repositories/botkube/pkg/execute/executor.go:213 +0xa35
  github.com/infracloudio/botkube/pkg/bot.(*slackMessage).HandleMessage()
      /Users/pkosiec/repositories/botkube/pkg/bot/slack.go:155 +0x6cd
  github.com/infracloudio/botkube/pkg/bot.(*SlackBot).Start()
      /Users/pkosiec/repositories/botkube/pkg/bot/slack.go:100 +0x564
  github.com/infracloudio/botkube/test/e2e.StartFakeSlackBot.func1()
      /Users/pkosiec/repositories/botkube/test/e2e/e2e_test.go:117 +0x39

Previous read at 0x0000087b4a18 by goroutine 67:
  github.com/infracloudio/botkube/pkg/controller.sendEvent()
      /Users/pkosiec/repositories/botkube/pkg/controller/controller.go:177 +0x6e5
  github.com/infracloudio/botkube/pkg/controller.registerEventHandlers.func1()
      /Users/pkosiec/repositories/botkube/pkg/controller/controller.go:135 +0x144
[...]

Issue #589, opened 2022-05-20
Increase test coverage


pkosiec:
This issue is an epic and aggregates multiple subtasks which increase overall test coverage in BotKube.

## To do

- This issue is an epic and aggregates multiple subtasks which increase overall test coverage in BotKube.

## To do

- [x] https://github.com/infracloudio/botkube/issues/615
- [ ] Write unit tests across all packages
    - Go through each package, file by file, and write missing unit tests
    - Split it to multiple pull requests
    - Refactor the code to make unit testing possible
    - Write missing unit tests
    - Reduce cyclomatic complexity, especially in: `pkg/execute/executor.go`, `pkg/controller/controller.go`, and `pkg/notify/slack.go`

- [ ] Review related issues and close/resolve them:
    - #349

- Write missing E2E tests
    - #282
    - #307
    - #322
    - #426
    - Webhook
    - MS Teams

- Write missing unit tests
    - Refactor the code to make unit testing possible
    - Reduce cyclomatic complexity, especially in: `pkg/execute/executor.go`, `pkg/controller/controller.go`, and `pkg/notify/slack.go`
    - It would be good to reduce usage of `time.Sleep` (now there are 7 matches when searching for such usage)
- Consider adding additional integration tests (see related issues for inspiration)

- It would be good to reduce usage of `time.Sleep` (now there are 7 matches when searching for such usage)

## Reason

Currently, BotKube has a minimal testing - mostly by the integration tests which bases on fake K8s client and fake Slack bot.
To ensure all features work as expected, we should improve the test coverage.

pkosiec:
Blocked by #220 as there wil be a bigger refactoring as a part of that task.

pkosiec:
Moving it out from 0.13 after discussion with @mszostok - we will write missing unit tests gradually over time, same with refactoring of each package 👍

PR #627, opened 2022-06-29
Remove old integration tests that use fake K8s and Slack APIs

### LINKED ISSUE #615: Prepare integration tests based on local Kubernetes cluster
## Description

- Prepare basic test scenario with Slack
    - Check if events are posted on channel, and if BotKube reacts for some kubectl commands 
    - Ideally use real Slack with channel create/delete; then these tests will be run on `develop` branch
    - Alternatively, use Slack mock deployed in cluster; then we will be able to run these tests on pull requests 
- Use real K8s API, executor and deploy BotKube with Helm
- Spawn k3d/kind cluster and install Helm chart
- Run tests via `helm test`
- Remove duplicated test cases from current tests based on fake K8s API
- Prepare GitHub pipeline (if we use mocked Slack, then it is both for PR and `develop` and it depends on #590; if not, then it runs only on `develop`)

## Related issue

See the epic #589

pkosiec:
##### ISSUE TYPE
 - Feature Pull Request

##### SUMMARY

⚠️ This PR waits for #624 to be merged.

- Remove old integration tests that use fake K8s and Slack APIs

As now there are new E2E tests (see #624), we can delete old tests and enable `-race` flag to make sure to fix any found race issues (whenever it is in tests or tested code).

There is still a chance that such race conditions exist in the code, however we will fix all found issues during increasing coverage in #589.

Resolves #615 and #592


pkg/controller/controller_test.go
```diff
@@ -7,22 +7,53 @@ import (
...
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 
 	"github.com/infracloudio/botkube/pkg/config"
 	"github.com/infracloudio/botkube/pkg/utils"
-	testutils "github.com/infracloudio/botkube/test/e2e/utils"
 )
 
 // TODO: Tests moved out straight from E2E test package with minimal changes.
```
mszostok, line 17: is it still valid?

test/doc.go
```diff
@@ -1,6 +0,0 @@
-// Package test contains integration tests with fake Slack and Kubernetes APIs.
```
mszostok, line 1: do you want to still address that? 
https://github.com/infracloudio/botkube/pull/622#discussion_r901612701

mszostok (approved):
LGTM 👍

pkg/controller/controller_test.go
pkosiec: This is still valid, as I didn't write additional test cases for the unit tests. This comment links to the test epic, co I think it is fine to keep it as it is. I believe will gradually rework unit tests when we will touch these packages, so the epic would be kept open.

test/doc.go
pkosiec: Thanks, right 👍 Will do that

Issue #634, opened 2022-07-05
Create a new Helm repository for Botkube


mszostok:
**Overview**

After repository transition, we need to fix BotKube Helm charts hosting. We should check such options:

- Use OCI artifacts, see https://helm.sh/blog/storing-charts-in-oci/
    This could be problematic:
    > Please note that there are some features that are not yet available for Helm repositories stored in OCI registries:
    > - Force an existing version to be reindexed by changing its digest

- Use GitHub pages
- Use the https://github.com/kubeshop/helm-charts

**Acceptance Criteria**

- We have a full control over the BotKube Helm repository,
  - Repository is hosted on BotKube or Kubeshop related infrastructure.
- The old release (v0.12.4) is re-pushed 
- We are able to re-push the v9.99.9 version each time there is a new commit on `main`.
  - alternatively we can think of pushing them with the commit SHA suffix.
- New repository is available on Artifact Hub
- The release documentation is updated

mszostok:
## Investigation

### OCI 

It looks nice, however it is lacking two nice features:
- "(..) OCI based registries don't provide standard APIs to facilitate searching.", so the `helm search` doesn't work, see [issue](https://github.com/helm/helm/issues/11000).
- force an existing version to be reindexed by changing its digest (use-full for latest). However, this can be simply changed and we can use the SHA suffix.

PR #637, opened 2022-07-06
Remove last chart release, remove www from botkube's urls

mszostok:
##### SUMMARY

- Remove last chart release from CI. It was used for development purposes. Currently, we don't need it and we will use the local Helm chart version.
- Remove `www` from Botkube's URLs

I also checked whether we can do the release to the local `gh-pages` branch. Here is the working GitHub action workflow: https://github.com/mszostok/botkube/actions/runs/2615913591/workflow#L51-L122

It works without any issues. So if we will need the latest Helm chart to be released, e.g., for installing on dev clusters etc. We can revisit this idea.

Fixes https://github.com/kubeshop/botkube/issues/634


pkosiec (approved):
LGTM

PR #645, opened 2022-07-14
Add automatic Go imports formatting

mszostok:
## Description

- Add automatic Go imports formatting

Example of failed job: https://github.com/kubeshop/botkube/runs/7349795533?check_suite_focus=true


Makefile
```diff
@@ -28,12 +28,12 @@ check-git-status:
 	@if [ -z "$(shell git remote -v)" ] ; then echo 'ERROR: No remote to push tags to' && exit 1 ; fi
 	@if [ -z "$(shell git config user.email)" ] ; then echo 'ERROR: Unable to detect git credentials' && exit 1 ; fi
 
-lint:
```
mszostok, line 31: it was not used, on CI we use a dedicated GitHub Action for `golangci-lint`

.github/workflows/ci.yml
```diff
@@ -26,6 +26,17 @@ jobs:
           args: --timeout=${{ env.GOLANGCI_LINT_TIMEOUT }}
       - name: Verify Go modules
         run: go mod verify
+      - name: Run Go import formatting
+        run: make go-import-fmt
+      - name: Detect git changes
+        run: |
+          if [[ $(git diff --stat) != '' ]]; then
+            echo -e '❌ \033[0;31m. Run 'make go-import-fmt'.\033[0m'
```
mszostok, line 34: or maybe `make lint-fix` ? just in case the `golangci-lint` or `go mod verify` will fail 🤔  However, we will need add `if: alwasy()` to  other checks, to ensure that the next step will be executed.

We can also add `go mod tidy` to `lint-fix` target 🤔

Makefile
```diff
@@ -28,12 +28,14 @@ check-git-status:
...
-	@golangci-lint run "./..."
-
-lint-fix:
+lint-fix: go-import-fmt
+	@go mod tidy
+	@go mod verify
 	@golangci-lint run --fix "./..."
 
+go-import-fmt:
```
pkosiec, line 36: Please also declare it as a phony target in the line 4 🙂

hack/fmt-imports.sh
```diff
@@ -0,0 +1,40 @@
...
+  popd > /dev/null
+}
+
+main() {
+  imports::install
+
+  imports::format
+}
+
+main
```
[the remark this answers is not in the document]
mszostok: It works for me. However, I know that in past you installed that tool, so check with `which goimports-reviser` you use, probably the old version that you installed. I changed the installation func to avoid such problem.

pkosiec (approved):
LGTM

PR #654, opened 2022-07-21
Get Go version from the go.mod file on CI

mszostok:
## Description

- Get Go version from the go.mod file on CI

I found out that there is a nice feature that we can use https://github.com/actions/setup-go#getting-go-version-from-the-gomod-file. As a result, we have a single source of truth about used Go version. As we don't test our code with multiple version (matrix) it's IMO a great solution.


pkosiec (approved):
LGTM

Issue #656, opened 2022-07-21
Update BotKube with a skeleton for the new configuration syntax


mszostok:
## Overview

Introduce only the syntax changes (nest under proper properties etc.). Support only one item on the list (hard-code `.[0].channel` etc.). 

## Acceptance Criteria

- BotKube works with a new syntax as described in the [binding configuration proposal](https://github.com/kubeshop/botkube/blob/main/docs/proposal/2022-06-14-configuration-bindings.md)
- Stretch: introduce the binding syntax.

Out of scope:
- docs changes
- `recommendation` property

## Reason

This will unblock the rest of the work in https://github.com/kubeshop/botkube/issues/596

Issue #655, opened 2022-07-21
Resolve setting array values using environment variables


pkosiec:
## Overview

After introducing the new configuration syntax proposal, we need to figure out how to override array items with environment variables.

For example:

```yaml
    communications:
      - name: tenant-b-workspace
        slack:
          token: 'SLACK_API_TOKEN'
```
What env will override this? `BOTKUBE_COMMUNICATION_0_SLACK_TOKEN`, or `BOTKUBE_COMMUNICATION_TENANT_B_WORKSPACE_SLACK_TOKEN`? 

Alternatively, we can change the syntax to:

```yaml
    communications:
      tenantBWorkspace:
        slack:
          token: 'SLACK_API_TOKEN'
          # ...trimmed...
``` 

## Acceptance Criteria

- Investigate possible options and shortly describe them (e.g. in the comment under this issue, or a separate document if necessary)
- Implement the selected solution

## Reason

See epic #596 for details.

mszostok:
The another thing is how to override a given property from file?
https://github.com/kubeshop/botkube/blob/main/pkg/config/testdata/config-slack-override.yaml

Are we going with full replacement for `communications` array?

PR #659, opened 2022-07-25
Update BotKube to load the new configuration syntax

mszostok:
## Description

- Update BotKube to load the new configuration syntax
- Update BotKube proposal, see https://github.com/kubeshop/botkube/pull/659/commits/9aed76c64a62fa8e3b84f21f4003848d38dd645c
  -  see the new **Use array instead of map** section for more info

## Related issue(s)

Fix https://github.com/kubeshop/botkube/issues/656
Fix https://github.com/kubeshop/botkube/issues/655

## Testing

The integration tests fails as it uses outdated settings from `main`. I tested that locally:

```bash
export SLACK_TESTER_APP_TOKEN="xoxb-"
export IMAGE_REPOSITORY="kubeshop/pr/botkube"
export IMAGE_TAG="659-PR"
export TEST_IMAGE_REPOSITORY="kubeshop/pr/botkube-test"
export PR_NUMBER=659

helm install botkube --namespace botkube ./helm/botkube --wait --create-namespace \
        -f ./helm/botkube/e2e-test-values.yaml \
        --set communications.default-workspace.slack.token="${SLACK_BOT_TOKEN}" \
        --set image.repository="${IMAGE_REPOSITORY}" \
        --set image.tag="${IMAGE_TAG}" \
        --set e2eTest.image.repository="${TEST_IMAGE_REPOSITORY}" \
        --set e2eTest.image.tag="${IMAGE_TAG}" \
        --set e2eTest.slack.testerAppToken="${SLACK_TESTER_APP_TOKEN}" \
        --set e2eTest.slack.additionalContextMessage="Pull request: ${PR_NUMBER} - https://github.com/kubeshop/botkube/pull/${PR_NUMBER}"
```

The result are available on official Testing Workspace, see:  https://testworkspace-6mc9503.slack.com/archives/C03RKGG3SSC/p1658760166208209
(channel: `#test-370005e8-db7e-4522-a3f2-8e5469d91f36`)


helm/botkube/values.yaml
```diff
@@ -292,127 +287,186 @@ config:
...
+      # -- BotKube Application Client ID.
+      botID: 'DISCORD_BOT_ID'
+      # -- Map of configured channels. The `channels` property name is an alias for a given configuration.
+      #
+      ## Format: channels.<alias>
+      channels:
+        'default':
+          # -- Discord channel ID for receiving BotKube alerts.
+          # The BotKube user needs to be added to it.
+          name: 'DISCORD_CHANNEL_ID'
```
pkosiec, line 390: Can we change this field to ID? It's quite confusing as we need to provide channel ID instead of name. Thanks!

pkosiec (commented):
Few small comments after code review. For now I tested Slack and Discord and they work fine. After fixes I will test all integrations 👍

helm/botkube/values.yaml
```diff
@@ -292,127 +287,186 @@ config:
...
+##      communications:
+##        # Here specify settings for each app, like Slack, Mattermost etc.
+##        # NOTE: Use setting format visible below.
+existingCommunicationsSecretName: ""
+
+# Map of enabled communication mediums. The `config` property name is an alias for a given configuration.
+#
+## Format: config.<alias>
 communications:
+  'default-workspace':
```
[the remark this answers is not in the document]
mszostok: I like the `default-group` 👍

pkg/config/config.go
```diff
@@ -355,3 +391,18 @@ func normalizeConfigEnvName(name string) string {
...
+// It's not deterministic if map has more than one element.
+func (t IndexableMap[T]) GetFirst() T {
+	var empty T
+
+	for _, v := range t {
+		return v
+	}
+
+	return empty
+}
```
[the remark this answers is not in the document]
mszostok: Maybe it will be useful later, but not sure, I added a comment :)

pkg/filterengine/with_all_filters.go
```diff
@@ -16,15 +16,21 @@ const (
...
 		filters.NewImageTagChecker(logger.WithField(filterLogFieldKey, "Image Tag Checker")),
 		filters.NewIngressValidator(logger.WithField(filterLogFieldKey, "Ingress Validator"), dynamicCli),
 		filters.NewObjectAnnotationChecker(logger.WithField(filterLogFieldKey, "Object Annotation Checker"), dynamicCli, mapper),
 		filters.NewPodLabelChecker(logger.WithField(filterLogFieldKey, "Pod Label Checker"), dynamicCli, mapper),
-		filters.NewNamespaceChecker(logger.WithField(filterLogFieldKey, "Namespace Checker"), conf.Resources),
 		filters.NewNodeEventsChecker(logger.WithField(filterLogFieldKey, "Node Events Checker")),
-	}...)
+	}
+
+	if len(conf.Sources) > 0 {
```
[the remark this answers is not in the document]
mszostok: I added that as it was a slice previously and I accessed the `[0]` elem. With map I think that it would work without this condidtion, removed 👍

helm/botkube/values.yaml
mszostok: This is quite tricky, if we will change that then we will need to create a dedicated struct just for the discrod bot.
Currently, it's unified under:
```go
// ChannelBindings contains configuration bindings per channel.
type ChannelBindings struct {
	Name     string      `yaml:"name"`
	Bindings BotBindings `yaml:"bindings"`
}
```
What about changing that to the `ID` for all? I know that some uses name and some ID, but IMO in case of "unification" we can go with `ID`. We will need to make a lot of such "unification" later, if we want to have it more extensible. 

The other idea is to have 
```go
// ChannelBindings contains configuration bindings per channel.
type ChannelBindings struct {
	ID       string      `yaml:"id"`
	Name     string      `yaml:"name"`
	Bindings BotBindings `yaml:"bindings"`
}
```

Where in config we can have `channel[*].id` or `channel[*].name` - we validate that only one was set and we use the one that is available. 

WDYT?

pkosiec: I'm not sure if that's a good unification, if some of the comm platforms require channel name (like Slack), and some ID (like Discord). We cannot unify that, as these are two different things - for example, Slack channel has Conversation ID, and also a name. 

Even a single struct with both ID and Name is confusing. I understand the problem, but I think the users' point of view is more important than ours (developer/maintainer one).

I would go with two separate structs to make it clearer to configure: e.g. `ChannelBindingsByName` and `ChannelBindingsByID` or whatever you like 🙂 Is that fine for you?

mszostok: >I'm not sure if that's a good unification, if some of the comm platforms require channel name (like Slack), and some ID (like Discord). We cannot unify that, as these are two different things - for example, Slack channel has Conversation ID, and also a name.

I know that, that the thing, we can say what ID means in our docs. 

Personally, I don't agree that it's necessary to have both ID and name, but let's agree to disagree - I will add a dedicated struct for that.

mszostok:
e2e test execution after applied suggestions:
https://testworkspace-6mc9503.slack.com/archives/C03R37W05A8/p1658836429243559 (`#test-ee9219b0-4bd9-4889-8813-e7d5d47ed961`)

pkosiec (approved):
LGTM 👍

PR #670, opened 2022-08-02
Handle multiple configurations for multiple bots and sinks

### LINKED ISSUE #666: Implement support for handling multiple communicators and multiple channels
## Overview

This is another step to achieve multi-channel support. Currently, even if we have communication groups, we use just the first one. Also, we don't support bindings.

## Acceptance Criteria

- Handle multiple communicators group in main
    - Remove `GetFirst()` usage for communicator configurations
- Support bindings per a given communication platform
   - Scoped to `bot` package only
   - Pass executor bindings for DefaultExecutor for a given channel where user wrote a command

Out of scope:
- Sources (e.g. `WithAllFilters` support)
- Executors (just ensure the bindings configuration in a form of map is passed into DefaultExecutor)

## Reason

See #596 epic issue.

pkosiec:
## Description

- Spawn multiple bots and sinks for each communication group
- Support multiple notification channels for all bots
- Support multiple indices for Elasticsearch sink
- Support toggling notifications on / off by channel for all bots
- Refactor trimming message prefix for all bots (+ add unit tests)
- Do not fall back to sending notifications to default channel (as there are no "default channel") when annotation config is wrong; instead, report an error
- Pass executor bindings to Executor
- Always return when running notifier commands (even if in not authorized channel)
- Add E2E tests
- Fix outdated E2E instructions
- Fix security vulnerability with yaml library

## Testing

I tested all integrations:

- [x] Slack
- [x] Discord
- [x] Mattermost
- [x] Teams
- [x] Elasticsearch
- [x] Webhook  

#### Configuration

Use the following uber config to test multi-channel configuration:

```yaml
communications:
  'default-group':
    # Settings for Slack
    slack:
      enabled: true
      token: {{TOKEN}}
      channels:
        'default':
          name: botkube-test
          bindings:
            executors:
              - kubectl-read-only
            sources:
              - k8s-events
        'two':
          name: general
          bindings:
            executors:
              - kubectl-read-only
            sources:
              - k8s-events

    mattermost:
      enabled: false
      botName: botkube
      notiftype: short
      team: org
      token: {{TOKEN}}
      url: http://mattermost-team-edition.default:8065
      channels:
        'default':
          name: "general"
          bindings:
            executors:
              - kubectl-read-only
            sources:
              - k8s-events
        'test':
          name: "test"
          bindings:
            executors:
              - kubectl-read-only
            sources:
              - k8s-events
      notification:
        type: short                             # Change notification type short/long you want to receive. Type is optional and default is short.

    # Settings for MS Teams
    teams:
      enabled: false
      appID: "{{APP ID}}
      appPassword: "{{APP PASS}}"
      botName: "PKBotKube"
      port: 3978
      notification:
        type: short

    # Settings for Discord
    discord:
      enabled: true
      botID: "{{BOT ID}}
      token: "{{BOT TOKEN}}"
      channels:
        'default':
          id: "{{CHAN ID 1}}"
          bindings:
            executors:
              - kubectl-read-only
            sources:
              - k8s-events
        'test':
          id: "{{CHAN ID 2}}"
          bindings:
            executors:
              - kubectl-read-only
            sources:
              - k8s-events
      notification:
        type: short                             # Change notification type short/long you want to receive. Type is optional and default is short.

    elasticsearch:
      enabled: false
      server: http://elasticsearch-master.default:9200
      username: elastic
      password: changeme
      skipTLSVerify: false                      # toggle verification of TLS certificate of the Elastic nodes. Verification is skipped when option is true. Enable to connect to clusters with self-signed certs
      # ELS index settings
      indices:
        'botkube':
          name: botkube
          type: botkube-event
          shards: 1
          bindings:
            sources:
              - "k8s-events"
            # executors - not allowed in this case, ES is "sink" only.
        'botkube2':
          name: botkube2
          type: botkube-event2
          shards: 1
          bindings:
            sources:
              - "k8s-events"
            # executors - not allowed in this case, ES is "sink" only.
    # Settings for Webhook
    webhook:
      enabled: false
      url: http://echo-server.default

settings:
  clusterName: dev
executors:
  'kubectl-read-only':
    kubectl:
      enabled: true

image:
  tag: '670-PR'
  pullPolicy: Always
e2eTest:
  image:
    repository: "kubeshop/pr/botkube-test"
    tag: "670-PR"
    pullPolicy: Always
```

You can duplicate the config for as `second-group` communication group as well. However, remember that such config doesn't make much sense, as separate communication groups are for different workspaces to avoid duplicate reponse. But I don't believe you'd like to create another test workspaces just to test these features 🙂 

```bash
helm install botkube --namespace botkube ./helm/botkube -f /tmp/values.yaml --wait
```

#### Webhook

You can install Echo server twice to test multiple communication groups:
```
helm install echo-server ealenn/echo-server --set application.logs.ignore.ping=true --set application.enable.environment=false --wait
helm install echo-server ealenn/echo-server --set application.logs.ignore.ping=true --set application.enable.environment=false --wait -n test --create-namespace

kubectl logs -l app.kubernetes.io/name=echo-server -f
kubectl logs -l app.kubernetes.io/name=echo-server -n test -f
```

## To do

- Integration tests
- Manual testing of all bots and sinks

## Related issue(s)

Resolves #666

### DIFF
--- a/pkg/bot/discord.go
+++ b/pkg/bot/discord.go
@@ -26,7 +26,10 @@ var _ Bot = &Discord{}
 
 // customTimeFormat holds custom time format string.
 const (
-	customTimeFormat          = "2006-01-02T15:04:05Z"
+	customTimeFormat = "2006-01-02T15:04:05Z"
+
+	// discordBotMentionRegexFmt supports also nicknames (the exclamation mark).
+	// Read more: https://discordjs.guide/miscellaneous/parsing-mention-arguments.html#how-discord-mentions-work
 	discordBotMentionRegexFmt = "^<@!?%s>"
 )
 
@@ -66,18 +69,17 @@ type discordMessage struct {
 
 // NewDiscord creates a new Discord instance.
 func NewDiscord(log logrus.FieldLogger, cfg config.Discord, executorFactory ExecutorFactory, reporter AnalyticsReporter) (*Discord, error) {
+	botMentionRegex, err := discordBotMentionRegex(cfg.BotID)
+	if err != nil {
+		return nil, err
+	}
+
 	api, err := discordgo.New("Bot " + cfg.Token)
 	if err != nil {
 		return nil, fmt.Errorf("while creating Discord session: %w", err)
 	}
 
 	channelsCfg := discordChannelsConfigFrom(cfg.Channels)
-
-	// Support nicknames (exclamation mark) as well: https://discordjs.guide/miscellaneous/parsing-mention-arguments.html#how-discord-mentions-work
-	botMentionRegex, err := discordBotMentionRegex(cfg.BotID)
-	if err != nil {
-		return nil, err
-	}
 
 	return &Discord{
 		log:             log,
--- a/pkg/bot/discord.go
+++ b/pkg/bot/discord.go
@@ -3,6 +3,7 @@ package bot
 import (
 	"context"
 	"fmt"
+	"regexp"
 	"strings"
 	"sync"
 
@@ -11,7 +12,9 @@ import (
 
 	"github.com/kubeshop/botkube/pkg/config"
 	"github.com/kubeshop/botkube/pkg/events"
+	"github.com/kubeshop/botkube/pkg/execute"
 	"github.com/kubeshop/botkube/pkg/format"
+	"github.com/kubeshop/botkube/pkg/multierror"
 )
 
 // TODO: Refactor this file as a part of https://github.com/kubeshop/botkube/issues/667
@@ -22,7 +25,10 @@ import (
 var _ Bot = &Discord{}
 
 // customTimeFormat holds custom time format string.
-const customTimeFormat = "2006-01-02T15:04:05Z"
+const (
+	customTimeFormat          = "2006-01-02T15:04:05Z"
+	discordBotMentionRegexFmt = "^<@!?%s>"
+)
 
 var embedColor = map[config.Level]int{
 	config.Info:     8311585,  // green
@@ -37,15 +43,13 @@ type Discord struct {
 	log             logrus.FieldLogger
 	executorFactory ExecutorFactory
 	reporter        AnalyticsReporter
-	notifyMutex     sync.RWMutex
-	notify          bool
 	api             *discordgo.Session
-
-	Notification     config.Notification
-	Token            string
-	ChannelID        string
-	BotID            string
-	ExecutorBindings []string
+	notification    config.Notification
+	botID           string
+	channelsMutex   sync.RWMutex
+	channels        map[string]channelConfigByID
+	notifyMutex     sync.Mutex
+	botMentionRegex *regexp.Regexp
 }
 
 // discordMessage contains message details to execute command and send back the result.
@@ -54,7 +58,6 @@ type discordMessage struct {
 	executorFactory ExecutorFactory
 
 	Event         *discordgo.MessageCreate
-	BotID         string
 	Request       string
 	Response      string
 	IsAuthChannel bool
@@ -62,26 +65,29 @@ type discordMessage struct {
 }
 
 // NewDiscord creates a new Discord instance.
-func NewDiscord(log logrus.FieldLogger, c *config.Config, executorFactory ExecutorFactory, reporter AnalyticsReporter) (*Discord, error) {
-	discord := c.Communications.GetFirst().Discord
-
-	api, err := discordgo.New("Bot " + discord.Token)
+func NewDiscord(log logrus.FieldLogger, cfg config.Discord, executorFactory ExecutorFactory, reporter AnalyticsReporter) (*Discord, error) {
+	api, err := discordgo.New("Bot " + cfg.Token)
 	if err != nil {
 		return nil, fmt.Errorf("while creating Discord session: %w", err)
+	}
+
+	channelsCfg := discordChannelsConfigFrom(cfg.Channels)
+
+	// Support nicknames (exclamation mark) as well: https://discordjs.guide/miscellaneous/parsing-mention-arguments.html#how-discord-mentions-work
+	botMentionRegex, err := discordBotMentionRegex(cfg.BotID)
+	if err != nil {
+		return nil, err
 	}
 
 	return &Discord{
 		log:             log,
 		reporter:        reporter,
 		executorFactory: executorFactory,
-		notify:          true, // enabled by default
 		api:             api,
-
-		Token:            discord.Token,
-		BotID:            discord.BotID,
-		ChannelID:        discord.Channels.GetFirst().ID,
-		ExecutorBindings: discord.Channels.GetFirst().Bindings.Executors,
-		Notification:     discord.Notification,
+		botID:           cfg.BotID,
+		notification:    cfg.Notification,
+		channels:        channelsCfg,
+		botMentionRegex: botMentionRegex,
 	}, nil
 }
 
@@ -95,7 +101,6 @@ func (b *Discord) Start(ctx context.Context) error {
 			log:             b.log,
 			executorFactory: b.executorFactory,
 			Event:           m,
-			BotID:           b.BotID,
 			Session:         s,
 		}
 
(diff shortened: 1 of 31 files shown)


pkg/bot/slack.go
```diff
@@ -178,47 +165,54 @@ func (b *Slack) IntegrationName() config.CommPlatformIntegration {
...
-				sm.log.Debugf("Ignoring message as it doesn't contain %q prefix", sm.BotID)
-				return nil
-			}
-			// Serve only if current channel is in config
-			if b.ChannelName == info.Name {
-				sm.IsAuthChannel = true
-			}
-		}
+	// Handle message only if it starts with bot mention
+	if !strings.HasPrefix(sm.Event.Text, "<@"+sm.BotID+">") {
```
mszostok, line 195: it's defined twice, 195 and 213, let's extract that to a const

mszostok, line 195: BTW, it would be nice to have this consistent across all Bots, if `discord` already uses:
```go
r.MatchString
r.ReplaceAllString
```
here we have:
```go
strings.HasPrefix
strings.TrimPrefix
```
so, using `regex` would be more generic solution that could work in all places, so we can have some helper method around that? 🤔

pkg/bot/slack.go
```diff
@@ -265,61 +259,80 @@ func (sm *slackMessage) Send() error {
...
-// SendMessage sends message to slack channel
-func (b *Slack) SendMessage(ctx context.Context, msg string) error {
-	b.log.Debugf(">> Sending to slack: %+v", msg)
-	channelID, timestamp, err := b.Client.PostMessageContext(ctx, b.ChannelName, slack.MsgOptionText(msg, false), slack.MsgOptionAsUser(true))
-	if err != nil {
-		return fmt.Errorf("while sending Slack message to channel %q: %w", b.ChannelName, err)
+func slackChannelsConfigFrom(channelsCfg config.IdentifiableMap[config.ChannelBindingsByName]) map[string]channelConfigByName {
+	channels := make(map[string]channelConfigByName)
+	for _, channCfg := range channelsCfg {
+		// index by Identifier as channelGroupName is not used
```
mszostok, line 330: what you mean by `channelGroupName`?

pkg/bot/slack.go
```diff
@@ -265,61 +259,80 @@ func (sm *slackMessage) Send() error {
...
+		channels[channCfg.Name] = channelConfigByName{
```
mszostok, line 331: should be 
```suggestion
		channels[channCfg.Identifier()] = channelConfigByName{
```
?

pkg/bot/discord.go
```diff
@@ -247,3 +255,38 @@ func (dm discordMessage) Send() {
...
+	r := regexp.MustCompile(fmt.Sprintf(`^<@!?%s>`, b.botID))
+	if !r.MatchString(msg) {
+		return "", false
+	}
+
+	return r.ReplaceAllString(msg, ""), true
+}
+
+func discordChannelsConfigFrom(channelsCfg config.IdentifiableMap[config.ChannelBindingsByID]) map[string]channelConfigByID {
+	res := make(config.IdentifiableMap[channelConfigByID])
```
mszostok, line 282: **nit**
```suggestion
	res := make(map[string]channelConfigByID)
```

pkg/bot/mattermost.go
```diff
@@ -315,100 +299,105 @@ func (b *Mattermost) listen(ctx context.Context) {
...
+func (b *Mattermost) setChannels(channels map[string]channelConfigByID) {
+	b.channelsMutex.Lock()
+	defer b.channelsMutex.Unlock()
+	b.channels = channels
+}
+
+func mattermostChannelsCfgFrom(client *model.Client4, teamID string, channelsCfg config.IdentifiableMap[config.ChannelBindingsByName]) (map[string]channelConfigByID, error) {
+	res := make(config.IdentifiableMap[channelConfigByID])
+	for _, channCfg := range channelsCfg {
+		fetchedChannel, resp := client.GetChannelByName(channCfg.Name, teamID, "")
```
mszostok, line 387:
```suggestion
		fetchedChannel, resp := client.GetChannelByName(channCfg.Identifier(), teamID, "")
```
let's be consistent, use `Identifier()` in all places or explicit field value. If we have a bug, it will be easy to trace and fix in all places.

pkg/bot/mattermost.go
```diff
@@ -315,100 +299,105 @@ func (b *Mattermost) listen(ctx context.Context) {
...
-	if _, resp := b.APIClient.CreatePost(post); resp.Error != nil {
-		return fmt.Errorf("while creating a post: %w", resp.Error)
+func (b *Mattermost) setChannels(channels map[string]channelConfigByID) {
+	b.channelsMutex.Lock()
+	defer b.channelsMutex.Unlock()
+	b.channels = channels
+}
+
+func mattermostChannelsCfgFrom(client *model.Client4, teamID string, channelsCfg config.IdentifiableMap[config.ChannelBindingsByName]) (map[string]channelConfigByID, error) {
+	res := make(config.IdentifiableMap[channelConfigByID])
```
mszostok, line 385:
```suggestion
	res := make(map[string]channelConfigByID)
```

pkg/bot/teams.go
```diff
@@ -46,61 +47,59 @@ const (
...
-	ClusterName      string
-	Notification     config.Notification
-	Adapter          core.Adapter
-	DefaultNamespace string
-
-	conversationRefMutex sync.RWMutex
-	conversationRef      *schema.ConversationReference
+	// TODO: Be consistent with other communicators when Teams supports multiple channels
+	//channels map[string][ChannelBindingsByName]
+	bindings config.BotBindings
```
mszostok, line 62: I cannot see any logic which is setting this field 🤔 The executor would not work properly in this case.

pkg/bot/teams.go
```diff
@@ -46,61 +47,59 @@ const (
...
-	AllowKubectl     bool
-	RestrictAccess   bool
-	ClusterName      string
-	Notification     config.Notification
-	Adapter          core.Adapter
-	DefaultNamespace string
-
-	conversationRefMutex sync.RWMutex
-	conversationRef      *schema.ConversationReference
+	// TODO: Be consistent with other communicators when Teams supports multiple channels
```
mszostok, line 60: I thought that:
1. User sets proper bindings in values YAML with expected channelIDs
2. When we get the "conversation" activity we check if we have a given ID on our list
    2.1 If yes, use a proper binding list
    2.2 If not, set `isAuthChannel` to `false` run executor 

Is this sth that can be implemented?

I know that it's different from Slack approach where user don't need to start the conversation, however if already started we should get sth that identifies a given channel (id/name)

pkg/bot/teams.go
```diff
@@ -218,8 +217,13 @@ func (b *Teams) processActivity(w http.ResponseWriter, req *http.Request) {
...
+
+			ref, err := b.getConversationReferenceFrom(activity)
+			if err != nil {
+				return schema.Activity{}, fmt.Errorf("while getting conversation reference: %w", err)
+			}
+
+			e := b.executorFactory.NewDefault(b.IntegrationName(), newTeamsNotifMgrForActivity(b, ref), true, activity.ChannelID, b.bindings.Executors, msgWithoutPrefix)
 			out := e.Execute()
 
 			actJSON, _ := json.MarshalIndent(turn.Activity, "", "  ")
```
mszostok, line 229: **nit** why the error is ignored 🤔 ?

mszostok (commented):
I didn't test that e2e, as IMO we would need to do that once both our PRs will be merged to the `main` branch (all conflict also resolved). In this way, we will test the full e2e scenario and see whether it behaves as expected.

pkg/bot/slack.go
```diff
@@ -8,12 +8,13 @@ import (
 
 	"github.com/sirupsen/logrus"
 	"github.com/slack-go/slack"
+	"go.uber.org/multierr"
```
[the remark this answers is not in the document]
pkosiec: Yep, right, that was a mistake 🙂

pkg/bot/teams.go
pkosiec: Unfortunately it's hard to get ID from Teams UI (link for channel contains UUIDs, but the Activity and channel IDs are numeric ones). As discussed, it needs to stay as it is 😞

ezodude (commented):
Great (but v massive!) PR 🙌. All makes sense.

There is no duplication exactly but the code feels repetitive.

Maybe in the future we can introduce a higher level abstraction e.g. MultiChanneller (just OTTOMH) to centralise notification logic.

pkg/bot/teams.go
```diff
@@ -294,37 +304,43 @@ func (b *Teams) putRequest(u string, data []byte) (err error) {
 
 // SendEvent sends event message via Bot interface
 func (b *Teams) SendEvent(ctx context.Context, event events.Event) error {
-	if !b.notify {
-		b.log.Info("Notifications are disabled. Skipping event...")
-		return nil
-	}
+	b.log.Debugf(">> Sending to Slack: %+v", event)
```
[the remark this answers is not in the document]
pkosiec: Yep, missed that one 😄 Thanks!

pkg/execute/notifier.go
```diff
@@ -11,23 +11,28 @@ import (
 )
 
 const (
-	notifierStartMsgFmt  = "Brace yourselves, incoming notifications from cluster %q."
-	notifierStopMsgFmt   = "Sure! I won't send you notifications from cluster %q anymore."
-	notifierStatusMsgFmt = "Notifications from cluster %q are %s."
+	notifierStartMsgFmt         = "Brace yourselves, incoming notifications from cluster %q."
```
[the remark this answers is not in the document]
pkosiec: I like `%q` more TBH, but let's be consistent 👍

pkg/bot/teams.go
pkosiec: That was not me 😄 But I will handle the error 👍

pkg/bot/discord.go
```diff
@@ -173,45 +166,60 @@ func (b *Discord) Type() config.IntegrationType {
...
+	return channel.notify
 }
 
-// SetNotificationsEnabled sets a new notification status.
-func (b *Discord) SetNotificationsEnabled(enabled bool) error {
-	b.notifyMutex.Lock()
-	defer b.notifyMutex.Unlock()
-	b.notify = enabled
+// SetNotificationsEnabled sets a new notification status for a given channel ID.
+func (b *Discord) SetNotificationsEnabled(channelID string, enabled bool) error {
```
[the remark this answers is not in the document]
pkosiec: you're right, I will add another mutex 👍

pkg/bot/slack.go
```diff
@@ -192,7 +199,7 @@ func (b *Slack) SetNotificationsEnabled(channelName string, enabled bool) error
 
 func (sm *slackMessage) HandleMessage(b *Slack) error {
 	// Handle message only if it starts with bot mention
-	if !strings.HasPrefix(sm.Event.Text, "<@"+sm.BotID+">") {
+	if !strings.HasPrefix(sm.Event.Text, fmt.Sprintf(slackBotMentionPrefixFmt, sm.BotID)) {
```
mszostok, line 202: it doesn't change, so it would be good to move it to constructor. The best option would be to use the same thing that you have on for other bots based on my previous comment: https://github.com/kubeshop/botkube/pull/670#discussion_r937658507

mszostok (approved):
🚀

Issue #668, opened 2022-08-01
Add logic to handle communicator bindings in `kubectl`


mszostok:
## Overview

Add logic to use settings with from given bindings.(blocked by #666) Proposed merge strategy:

**Policy**

- merge only if enabled `kubectl.enabled == true`
- the merge appends the items of
	- `kubectl.commands.verbs`
	- `kubectl.commands.resources`
		Only if the allowed Namespace is matched.

```yaml
bindings:
  executors:
    - 'kubectl-team-a'
    - 'kubectl-team-b'
    - 'kubectl-global'
    -  'kubectl-exec'
executors:
  'kubectl-team-a':
    kubectl:
      namespaces:
        include: [ "team-a" ]
      commands:
        verbs: [ "get" ]
        resources: [ "deployments" ]
  'kubectl-team-b':
    kubectl:
      namespaces:
        include: [ "team-b" ]
      commands:
        verbs: [ "get", "describe" ]
        resources: [ "deployments", "pods" ]
  'kubectl-global':
    kubectl:
      namespaces:
        include: [ "all" ]
      commands:
        verbs: [ "logs", "top" ]
        resources: [ ]
  'kubectl-exec':
    kubectl:
      enabled: false
      namespaces:
        include: [ "all" ]
      commands:
        verbs: [ "exec" ]
        resources: [ ]
```

## Acceptance Criteria

- `kubectl` executor works only if there is a given binding
- if there are multiple bindings for a given channel, apply merging strategy and execute `kubectl`

## Reason

See https://github.com/kubeshop/botkube/issues/596

PR #672, opened 2022-08-02
Add logic to handle communicator bindings in kubectl

mszostok:
## Description

**Enhancements**
- Currently, we allowed to have only one executor binding on the list. 
  **Now executor manager is able to take the bindings list and calculate proper restrictions for a given channel.**
- E2E and unit test coverage to ensure no regression.
- Change `commands list` output:
    <table>
    <tr>
    <td> Before </td> <td> After </td>
    </tr>
    <tr>
    <td>

    ```yaml
    allowed verbs:
      - api-resources
      - api-versions                        
      - auth
      - cluster-info
      - describe
      - diff
      - explain
      - get
      - logs
      - top
    allowed resources:
      - configmaps
      - daemonsets
      - deployments
      - namespaces
      - nodes
      - pods
      - statefulsets
      - storageclasses
    allowed namespaces:
      include:
        - botkube
        - default
    ```

    </td>
    <td>

    ```yaml
    enabled:
      kubectl:
        kubectl-read-only:                          
          namespaces:
            include:
              - botkube
              - default
          enabled: true
          commands:
            verbs:
              - api-resources
              - api-versions
              - cluster-info
              - describe
              - diff
              - explain
              - get
              - logs
              - top
              - auth
            resources:
              - deployments
              - pods
              - namespaces
              - daemonsets
              - statefulsets
              - storageclasses
              - nodes
              - configmaps
          defaultNamespace: default
          restrictAccess: false
        kubectl-wait-cmd:
          namespaces:
            include:
              - botkube
              - default
          enabled: true
          commands:
            verbs:
              - wait
            resources: []
          restrictAccess: false`
    ```

    </td>
    </tr>
    </table>

**Bugs:**
-  On v0.12.4 `@BotKube  get po/botkube-56c8c5874b-f2bbm -n botkube` results in:
    ```
    Command not supported. Please run /botkubehelp to see supported commands.
    ```
    This was addressed on this PR. And now it's validated properly.
- On `main` the `-A` was returning all data even if not authorized `namespace.include: ["default"]`
   This was addressed on this PR. Now, the `-A` works only if you have binding with `namespace.include: ["all"]`
- On v0.12.4 `BotKube get -n botkube po` returns, 
    ```
    Command not supported. Please run /botkubehelp to see supported commands.
    ```
    On main returns:
    ```
    Sorry, the kubectl command is not authorized to work with '-n' resources on cluster 'dev'. Use 'commands list' to see all allowed resources.
    ```
    In this PR it was fixed to:
    ```
    Please specify the resource name after the verb, and all flags after the resource name. Format <verb> <resource> [flags]
    ```
    The better approach will be to remove all flags and find the resource name, however it's not simple. The proper solution should be to delegate this to `kubectl` which can do a proper parsing.

## Testing

The merger was fully tested with unit-tests 
- https://github.com/kubeshop/botkube/pull/672/files#diff-efacb45b078225d6d6058c5ca9152fb14cb291663d78dd64f3984bcae17d49dd
- https://github.com/kubeshop/botkube/pull/672/files#diff-189b9a00b5d538a5926c3c1543d0039ce151dd3e732dbefa346c103f4db44568

and e2e tests:
- It's tested as a part of the e2e tests. See new values: https://github.com/kubeshop/botkube/blob/c5b791b7ba3d0d0e07f12e2df9618af08658ee93/helm/botkube/e2e-test-values.yaml#L31-L48

It doesn't make sens to test the rest of the communication platform as the logic was not changed for them.

## Related issue(s)

Fixes https://github.com/kubeshop/botkube/issues/668


pkg/execute/kubectl/merger.go
```diff
@@ -0,0 +1,119 @@
...
+	AllowedKubectlVerb     map[string]struct{}
+	AllowedKubectlResource map[string]struct{}
+
+	DefaultNamespace string
+	RestrictAccess   bool
+}
+
+// Merger provides functionality to merge multiple bindings
+// associated with the kubectl executor.
+type Merger struct {
```
mszostok, line 18: We can create a wrapper to cache the calc responses. Example pseudo code:

```go
package kubectl

import (
	"fmt"

	"github.com/kubeshop/botkube/pkg/config"
)

type Merger interface {
	MergeForNamespace(includeBindings []string, forNamespace string) EnabledKubectl
	MergeAllEnabled(includeBindings []string) EnabledKubectl
	IsAtLeastOneEnabled() bool
	MergeAllEnabledVerbs(bindings []string) map[string]struct{}
}

type CachedMerger struct {
	underlying Merger
}

// NewMergerWithCache returns a new Merger instance with caching layer.
func NewMergerWithCache(merger Merger) *CachedMerger {
	return &CachedMerger{
		underlying: merger,
	}
}

func (m *CachedMerger) MergeAllEnabledVerbs(bindings []string) map[string]struct{} {
	const key = "MergeAllEnabledVerbs"
	m.mux[key].Lock()
	defer m.mux[key].UnLock()

	var entryKey = fmt.Sprintf("%s/%s", key, bindings)
	out, found := m.cache[entryKey]
	if !found {
		out = m.underlying.MergeAllEnabledVerbs(bindings)
		m.cache[entryKey] = out
	}

	return out
}

func (m *CachedMerger) MergeForNamespace(includeBindings []string, forNamespace string) EnabledKubectl {
 // TODO
}

func (m *CachedMerger) MergeAllEnabled(includeBindings []string) EnabledKubectl {
 // TODO
}

func (m *CachedMerger) IsAtLeastOneEnabled() bool {
 // TODO
}

```

We can also use some caching Go lib, e.g. https://github.com/allegro/bigcache

pkg/bot/slack.go
```diff
@@ -47,11 +47,8 @@ type Slack struct {
 
 	Client           *slack.Client
```
pkosiec, line 48: Hmm 🤔 I tried to test it and:

```
      channels:
        'default':
          # -- Slack channel name without '#' prefix where you have added BotKube and want to receive notifications in.
          name: 'SLACK_CHANNEL'
          bindings:
            executors:
              - kubectl-read-only
              - kubectl-read-only2
            sources:
              - k8s-events`
```

```
executors:
  'kubectl-read-only2':
    kubectl:
      namespaces:
        # -- List of allowed Kubernetes Namespaces for command execution. The "all" keyword, allows all the Namespaces.
        include:
          - all
        # -- List of ignored Kubernetes Namespace. It can be used only when `include: ["all"]`.
        # It can contain a wildcard (*) that expands to zero or more arbitrary characters
        # Example: `ignore: ["x", "y", "secret-ns-*"]`
        ignore: []
      # -- If true, enables `kubectl` commands execution.
      enabled: true
      ## List of allowed `kubectl` commands.
      commands:
        # -- Configures which `kubectl` methods are allowed.
        verbs: ["get", "logs", "top", "auth"]
        # -- Configures which K8s resource are allowed.
        resources: ["deployments", "pods"]
      # -- Configures the default Namespace for executing BotKube `kubectl` commands. If not set, uses the 'default'.
      defaultNamespace: default
      # -- If true, enables commands execution from configured channel only.
      restrictAccess: true
  'kubectl-read-only':
    kubectl:
      namespaces:
        # -- List of allowed Kubernetes Namespaces for command execution. The "all" keyword, allows all the Namespaces.
        include:
          - all
        # -- List of ignored Kubernetes Namespace. It can be used only when `include: ["all"]`.
        # It can contain a wildcard (*) that expands to zero or more arbitrary characters
        # Example: `ignore: ["x", "y", "secret-ns-*"]`
        ignore: []
      # -- If true, enables `kubectl` commands execution.
      enabled: true
      ## List of allowed `kubectl` commands.
      commands:
        # -- Configures which `kubectl` methods are allowed.
        verbs: ["api-resources", "api-versions", "cluster-info", "describe", "diff", "explain"]
        # -- Configures which K8s resource are allowed.
        resources: [ "namespaces", "daemonsets", "statefulsets", "storageclasses", "nodes", "configmaps"]
      # -- Configures the default Namespace for executing BotKube `kubectl` commands. If not set, uses the 'default'.
      defaultNamespace: default
      # -- If true, enables commands execution from configured channel only.
      restrictAccess: true
```

Screenshot 2022-08-04 at 13 22 10

Screenshot 2022-08-04 at 13 22 18


Looks like the `kubectl-read-only2` binding doesn't work, even if I have that in `commands list`:
Screenshot 2022-08-04 at 13 22 59

Do you know what could be the problem?

pkosiec (commented):
Looks great 🚀 I have one open question when it comes to the behavior. Not sure if that's a bug or I don't understand it (probably the latter). But overall LGTM once we sync f2f about it ✅

pkg/bot/discord.go
```diff
@@ -43,12 +43,9 @@ type Discord struct {
 
 	Notification     config.Notification
 	Token            string
-	AllowKubectl     bool
-	RestrictAccess   bool
-	ClusterName      string
 	ChannelID        string
 	BotID            string
-	DefaultNamespace string
+	ExecutorBindings []string
```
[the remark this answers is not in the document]
mszostok: yep, it was for testing, I didn't want to duplicate your work 👍 so this is a small step for handling binding where still you have opt to have only one channel.

pkg/bot/slack.go
pkosiec: Alright, that was issue on my side, thanks for the sync 😄

pkosiec (approved):
Works well 👌 Please see the comments before merge.

pkg/execute/kubectl/merger.go
```diff
@@ -0,0 +1,138 @@
...
+// IsAtLeastOneEnabled returns true if at least one kubectl executor is enabled.
+func (kc *Merger) IsAtLeastOneEnabled() bool {
+	for _, executor := range kc.executors {
+		if executor.Kubectl.Enabled {
+			return true
+		}
+	}
+	return false
+}
+
```
ezodude, line 62: Having had a cursory look at `MergeAllEnabledVerbs` is used.

To get the semantics right, what if it simply returns a Set? (`map[string]bool`) ?

As a freebie, you just get the value and wont't need `_, ok := ...`

pkg/execute/kubectl.go
```diff
@@ -3,51 +3,58 @@ package execute
...
 // As a result, we are able to detect kubectl command but say that you're simply not authorized to use it instead of "Command not supported. (..)"
-func (e *Kubectl) CanHandle(args []string) bool {
+func (e *Kubectl) CanHandle(bindings []string, args []string) bool {
 	if len(args) == 0 {
 		return false
 	}
 
 	// Check if such kubectl verb is enabled
-	if !e.resMapping.AllowedKubectlVerbMap[args[0]] {
+	if !e.kcChecker.IsKnownVerb(e.merger.MergeAllEnabledVerbs(bindings), args[0]) {
```
ezodude, line 57: Slight nit pick.

What if the merger simply told you if it includes a verb? e.g. `e.merger.includesVerb(args[0]])`

ezodude (commented):
That is one big PR! 😅

I understand the premise. And, left a few small comments mainly regarding style and less indirection.

Will run the code before I approve. 🙂

pkg/execute/executor.go
```diff
@@ -377,32 +374,24 @@ func (e *DefaultExecutor) runVersionCommand(args []string, clusterName string) s
...
 
-		keys = append(keys, key)
+		enabledKubectls[name] = item.Kubectl
 	}
-	sort.Strings(keys)
 
-	if len(keys) == 0 {
-		return fmt.Sprintf("%s: []", header)
+	out := map[string]map[string]kubectlCollection{
+		"enabled": {
```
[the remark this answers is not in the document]
mszostok: I will just replace the `enabled` with `Enabled executors` in that way the indent will be more predictable 👍

pkg/execute/executor.go
```diff
@@ -377,32 +374,24 @@ func (e *DefaultExecutor) runVersionCommand(args []string, clusterName string) s
...
-
-func (e *DefaultExecutor) getNamespaceConfig() (string, error) {
-	ns := e.cfg.Executors.GetFirst().Kubectl.Namespaces
-
-	out := map[string]config.Namespaces{
-		"allowed namespaces": ns,
-	}
 	var buff strings.Builder
 	encode := yaml.NewEncoder(&buff)
 	encode.SetIndent(2)
```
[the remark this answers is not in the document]
mszostok: I used that previously because in default setting it is using 4 spaces, and that was not in sync with the data that was printed "manually" with:
```go
	const itemSeparator = "\n  - "
	items := strings.Join(keys, itemSeparator)
```

Now we use only `yaml.Marshal` but I still think that 2 spaces are ok and not 4 are a bit too wide in the communicator window.

pkg/execute/kubectl.go
```diff
@@ -175,40 +188,71 @@ func (e *Kubectl) getNamespaceFlag(args []string) (string, error) {
...
+func (e *Kubectl) addNamespaceFlag(args []string, defaultNamespace string) []string {
+	return append([]string{"-n", defaultNamespace}, utils.DeleteDoubleWhiteSpace(args)...)
+}
+
+func (e *Kubectl) getResourceName(args []string) string {
+	if len(args) < 2 {
+		return ""
 	}
+	parts := strings.SplitN(args[1], "/", 2)
+	resource := parts[0]
```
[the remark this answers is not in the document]
mszostok: wow, that's neat! 
```go
func Test(t *testing.T) {

	resource, _, _ := strings.Cut("po/test", "/")
	fmt.Println(resource)

	resource, _, _ = strings.Cut("po/test/foo", "/")
	fmt.Println(resource)

	resource, _, _ = strings.Cut("foo", "/")
	fmt.Println(resource)

	resource, _, _ = strings.Cut("/foo", "/")
	fmt.Println(resource)
}
```

so I just simplified to 
```go
func (e *Kubectl) getResourceName(args []string) string {
	if len(args) < 2 {
		return ""
	}
	resource, _, _ := strings.Cut(args[1], "/")
	return resource
}
```

pkg/execute/executor.go
```diff
@@ -140,7 +138,7 @@ func (action FiltersAction) String() string {
 }
 
 // Execute executes commands and returns output
-func (e *DefaultExecutor) Execute() string {
+func (e *DefaultExecutor) Execute(channelBindings []string) string {
```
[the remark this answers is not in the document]
mszostok: >  I think the latter option would be more consistent

Make sens, so feel free to replace that on you PR 👍

pkg/execute/executor_test.go
```diff
@@ -6,121 +6,148 @@ import (
...
+			res, err := executor.getEnabledKubectlConfigs()
 
 			// then
 			require.NoError(t, err)
 			assert.Equal(t, tc.expOutput, res)
 		})
 	}
 }
+
+func fixExecutorsConfig(t *testing.T, raw string) config.IndexableMap[config.Executors] {
```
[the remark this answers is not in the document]
mszostok: The `fixtureExecutorsConfig` is a short for `fixtureExecutorsConfig`. We had this approach and other methods used in testing has this `fix` prefix. Let me know if it's ok for you 👍

pkg/execute/kubectl.go
mszostok: IMO it would be good to have that separated. For now, the boundaries are quite easy to describe. Merger takes care of any sort of merging strategies. Checker, on the other hand, it's used to assert some expectation. I'm not sure if it's worth to mix this functionality even if it's simple. I tried to avoid having a one component that will do all stuff, instead wanted to provide smaller components where the logic is simple enough to be well tested and read.

mszostok:
Hi @ezodude 
From the output that you posted, I can see that you are not using the image from this PR. Please use:
```
--set image.tag=672-PR \
```
In the e2e tests, you can see that we test that scenario. I specify multiple executors:
https://github.com/kubeshop/botkube/blob/d12b76ac1a1ee85c3ed2125d6a6a7dd6e394e34b/helm/botkube/e2e-test-values.yaml#L34-L70

and assert that only those enabled would be displayed:
https://github.com/kubeshop/botkube/blob/d12b76ac1a1ee85c3ed2125d6a6a7dd6e394e34b/test/e2e/slack_test.go#L124-L175

The new format was also mentioned in the PR description. 

It has to be changed as a given list of resources and verbs need to be associated with a given set of namespaces. I could try to come up with some merging strategy. However, for now, IMO it's good to print the config as it is so user don't need to be familiar with different syntaxes. In the long term, it would be nice to revisit all BotKube's commands and come up with some consistent approach (better UX). In this way, configuration would be nicely presented (with some field/modals?) to the user, but he will still have an option to get plain config format e.g. using `raw` flag.

pkg/execute/kubectl/merger.go
mszostok: I learned that a long time ago. I'm using the `map[string]struct{}` not because it's give us [allocation improvements](https://dave.cheney.net/2014/03/25/the-empty-struct) but it gives more explicit API 



>I use and will continue to use map[string]struct{}, because when someone reads that, they know, beyond a shadow of a doubt, that existence is the only thing that matters and there is no place in the code, no matter how far this map travels, that will ever make a decision based on anything else. When I used map[string]bool, the person reading the code does NOT know that. You can assert it in a comment all you like, but the reader must still consult the code to verify that it can't go anywhere where some other programmer (who may even be you at some other point in time) didn't decide to make the bool parameter mean something. 

So when I see `map[string]bool` I start thinking if there is a functionality that is able to enable/disable it, if yes then on what condition? 
On the other hand, using `[]string` give us O(n) instead of `O(1)`. 

I'm open to discuss it further and establish an approach that would be suitable for both of us 👍

pkg/execute/executor.go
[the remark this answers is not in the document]
pkosiec: Alright, but it would be great to have consistency. But keep it as it is, we'll definitely improve configs printed later 👍

pkg/execute/kubectl.go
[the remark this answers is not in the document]
pkosiec: Right, I forgot about this: `// If sep does not appear in s, cut returns s, "", false.` 😄

pkosiec (approved):
LGTM

mszostok:
Merging this PR as it's blocking https://github.com/kubeshop/botkube/pull/670

pkg/execute/kubectl/merger.go
ezodude: All good. 👍

pkg/execute/kubectl.go
ezodude: I see what you're getting at - I just felt there's a bit of jumping around.

Ensuring the components are decoupled trumps everything. Here, I felt a generic (`includes...`)[tell don't ask](https://martinfowler.com/bliki/TellDontAsk.html) might help.

It definitely should not delay merging this PR - so all good.

Issue #596, opened 2022-05-25
Support fine-grained configuration per channels for the same bot (e.g. Slack)


pkosiec:
## Overview

This issue is an epic and aggregates all the subtasks to finalize the complete feature.

## Reason

There is a demand in community—5 open issues with a lot of reactions—to make the configuration more flexible in regards which notification should be collected and where to send.

## Acceptance Criteria

- [x] #643
- [x] User is able to route notifications:
  - [x] to individual channels without needing to run multiple deployments.
  - [x] to a given channel based on the Kubernetes Namespace.
- [x] User is able to define `kubectl` permissions per channel—in particular, configuring what commands can be executed and in which Namespaces.
- [x] User-facing documentation is up-to-date.
- [x] Unit-test/e2e test coverage is added to ensure no regression in the future.

## To do

- [x] Prepare design draft / proposal - @mszostok, see:
    - https://github.com/infracloudio/botkube/pull/623,
    - https://github.com/infracloudio/botkube/pull/626
- [x] Gather community feedback about new design.
- [x] Prepare a list of subtasks. Split them into:
  - [x] required—those which need to be finished for MVP
  - [x] follow-up—those which can be addressed later
- [x] Start the asynchronous work to finish all required subtasks.

## Related issues
This issue aggregates the feature requests from:

- https://github.com/infracloudio/botkube/issues/250
- https://github.com/infracloudio/botkube/issues/486
- https://github.com/infracloudio/botkube/issues/444
- https://github.com/infracloudio/botkube/issues/508
- https://github.com/infracloudio/botkube/issues/542

All the tasks above will be eventually closed and the tracking of the feature will be done here.

mszostok:
## Findings

The https://github.com/infracloudio/botkube/issues/250 is not strictly related to this issue, see: https://github.com/infracloudio/botkube/issues/250#issuecomment-1155030581

New related issues:
- https://github.com/infracloudio/botkube/issues/508
- https://github.com/infracloudio/botkube/issues/542

pkosiec:
Closing as all subtasks have been resolved.

PR #678, opened 2022-08-09
Update Helm chart values description

pkosiec:
## Description

- Update Helm chart values description

Fix for failing tests in #680 

## Related issue(s)

#596

### DIFF
--- a/helm/botkube/values.yaml
+++ b/helm/botkube/values.yaml
@@ -51,8 +51,8 @@ kubeconfig:
   existingSecret: ""
 
 
-# -- Map of enabled sources. The `sources` property name is an alias for a given configuration.
-# Key name used as a binding reference.
+# -- Map of enabled sources. The property name under `sources` object is an alias for a given configuration. You can define multiple sources configuration with different names.
+# Key name is used as a binding reference.
 #
 ## Format: sources.<alias>
 sources:
@@ -267,7 +267,7 @@ sources:
        #     fields:
        #       - status.phase
 
-# Map of enabled executors. The `executors` property name is an alias for a given configuration.
+# Map of enabled executors. The property name under `executors` is an alias for a given configuration. You can define multiple executor configurations with different names.
 # Key name is used as a binding reference.
 #
 ## Format: executors.<alias>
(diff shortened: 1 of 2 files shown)


helm/botkube/values.yaml
```diff
@@ -306,25 +306,27 @@ executors:
 ##        # NOTE: Use setting format visible below.
 existingCommunicationsSecretName: ""
 
-# Map of enabled communication mediums. The `config` property name is an alias for a given configuration.
+# Map of enabled communication mediums. The property name under `communications` object is an alias for a given configuration group. You can define multiple communication groups with different names.
```
mszostok, line 309:
```suggestion
# Map of communication mediums. The property name under `communications` object is an alias for a given configuration group. You can define multiple communication groups with different names.
```

helm/botkube/values.yaml
```diff
@@ -267,7 +267,7 @@ sources:
        #     fields:
        #       - status.phase
 
-# Map of enabled executors. The `executors` property name is an alias for a given configuration.
+# Map of enabled executors. The property name under `executors` is an alias for a given configuration. You can define multiple executor configurations with different names.
```
mszostok, line 270:
```suggestion
# Map of executors. The property name under `executors` is an alias for a given configuration. You can define multiple executor configurations with different names.
```

helm/botkube/values.yaml
```diff
@@ -51,8 +51,8 @@ kubeconfig:
   existingSecret: ""
 
 
-# -- Map of enabled sources. The `sources` property name is an alias for a given configuration.
-# Key name used as a binding reference.
+# -- Map of enabled sources. The property name under `sources` object is an alias for a given configuration. You can define multiple sources configuration with different names.
+# Key name is used as a binding reference.
```
mszostok, line 55: maybe it would be good to add also a default description. Currently, we print the whole object in README.md which is not readable.

We can define sth like
```
# @default -- Enables watching for Kubernetes events and sending recommendations.
```
The property name is "clickable" and will navigate to the definition, which will provide more context.

pkosiec: @mszostok That would be a duplication of the description, but I like the idea to get rid of the full object. WDYT about ```
```
?

mszostok: >That would be a duplication of the description

Not exactly, as there we're saying what kind of k8s notification we enable. However `# @default -- See the `values.yaml` file for full object.` also works for me 👌

pkosiec: Alright then, I will append that to the actual description to make sure we have this info in docs 👍

mszostok (approved):

Issue #679, opened 2022-08-09
Refactor the Namespace config to simplify usage


mszostok:
## Overview

Refactor the Namespace config to simplify usage.

## Acceptance Criteria
- [ ] change `include/ignore`to `include/exclude`
- [ ] support regex in both cases (include/exclude)
- [ ] remove the `all` keyword. We will have regex for `include`, so `*`will match all namespaces.
- [ ] add validation to print warning when `include` matches all namespaces but on the list there are more items. Because in this case, we will match all Namespaces anyway.
- [ ] Update documentation

## Reason

See epic https://github.com/kubeshop/botkube/issues/596

PR #686, opened 2022-08-10
Support regex for namespace.include, replace 'all' with '.*', rename include to exclude

mszostok:
…include to exclude

## Description

- Support regex for `namespace.include`, 
- Replace 'all' with '.*', 
- Rename include to exclude.

## Related issue(s)

Fix https://github.com/kubeshop/botkube/issues/679


global_config.yaml.tpl
```diff
@@ -11,27 +11,32 @@ sources:
```
pkosiec, line 11: BTW The PR name says `rename include to exclude`, but it should be `rename ignore to exclude` 😄

helm/botkube/values.yaml
```diff
@@ -69,31 +69,32 @@ sources:
       resources:
```
pkosiec, line 69: BTW PR title says `rename include to exclude` 😄 but I think you meant `rename ignore to exclude` 😄

pkg/config/config.go
```diff
@@ -198,53 +197,66 @@ type UpdateSetting struct {
...
+	// Exclude contains a list of Namespaces to be ignored even if allowed by Include.
+	// It can also contain a regex expressions:
+	//  - "test-.*" - to specif all Namespaces with `test-` prefix.
+	Exclude []string `yaml:"exclude,omitempty"`
 }
 
 // IsAllowed checks if a given Namespace is allowed based on the config.
-// Copied from https://github.com/kubeshop/botkube/blob/b6b7d449278617d40f05d0792b419a7692ad980f/pkg/filterengine/filters/namespace_checker.go#L54-L76
-// TODO(https://github.com/kubeshop/botkube/issues/596): adjust contract.
 func (n *Namespaces) IsAllowed(givenNs string) bool {
```
pkosiec, line 214: I'm not sure if that's a good place (I don't think so), but will you also apply the last AC besides documentation?

> add validation to print warning when include matches all namespaces but on the list there are more items. Because in this case, we will match all Namespaces anyway.

You can definitely push a commit here as I don't expect this change to be that big 🙂

pkosiec (approved):
LGTM, but please add the warning before merge 🙂

pkg/config/config.go
mszostok: OK, I thought so at the beginning but then, it turns out that the impl is not so straight forward, but here it is: https://github.com/kubeshop/botkube/pull/686/commits/93b5fdcf409e004f8e7687f9d038061f40ab632a

pkosiec (approved):
Tested, works well 👍

Issue #684, opened 2022-08-10
Add ability to configure specific recommendations during BotKube installation/upgrade


pkosiec:
## Overview

Implement ability to turn on/off recommendations according to the proposal:  https://github.com/kubeshop/botkube/blob/main/docs/proposal/2022-06-14-configuration-bindings.md#new-syntax

```yaml
        # Recommendations about the best practices
        recommendations:
          image:     # "Checks if 'latest' image tag is used for container image."
            enabled: true
          pod:       # "Checks if labels are missing in the pod specs."
            enabled: true
          ingress:   # "Checks if services and tls secrets used in ingress are available."
            enabled: true
```

## Acceptance Criteria

- [x] Support recommendation configuration under `source` and enable/disable proper recommendations during BotKube startup
- [x] Ensure enabling/disabling filters in runtime via `@BotKube` command still works
- [x] Update documentation
- [x] Update merging strategy to "override" to be consistent with `Executor.Kubectl.Commands.RestrictAccess`
- [x] Create informers if recommendations are enabled, even if Ingress and Pod resources are not configured in the resources configuration (blocked by https://github.com/kubeshop/botkube/issues/676)

## Reason

See epic https://github.com/kubeshop/botkube/issues/596

Issue #676, opened 2022-08-05
Support routing notifications to a given channel based on Kubernetes Sources


pkosiec:
## Overview

User should be able to define different notification sources per a given channel based on the Kubernetes namespace. See the proposal section: https://github.com/kubeshop/botkube/blob/main/docs/proposal/2022-06-14-configuration-bindings.md#route-notifications-to-a-given-channel-based-on-the-kubernetes-namespace

Currently, the sources bindings introduced in new config are not supported:

https://github.com/kubeshop/botkube/blob/95918ec778dfb4866366bbc96eed389e870df8a9/helm/botkube/values.yaml#L328-L329

The scope of this task is to:
- support source bindings in all communication platforms
- ~support Kubernetes event namespace filtering by another property for `kubernetes` source~ (This has been de-scoped)

## Acceptance Criteria

- ~Add new `namespace.include/exclude` property in the `sources.'group-name'.kubernetes` according to the proposal~ (This has been de-scoped)
- Ensure multiple source definitions are supported
- Support source bindings per channel / index for all communication platforms (bots and sinks)
- Update documentation on `botkube-docs` repo
- Add required unit and E2E tests

## Reason

See epic https://github.com/kubeshop/botkube/issues/596.

ezodude:
Now in code exploration mode.

pkosiec:
BTW @ezodude regarding the comments
> (This has been de-scoped)
I think there was a misunderstanding. I believe that's still a scope of this task, but this is definitely something to do in a follow-up PR (not in the first one) 🙂 And then we will have all parts together of the whole multi-workspace/multi-channel configuration epic.

pkosiec:
Ok, I see you extracted the namespaces functionality to a separate task. Closing it then 👍

PR #687, opened 2022-08-11
Add ability to configure specific recommendations during BotKube installation/upgrade

pkosiec:
## Description

- Add ability to configure recommendations in granular way inside the config
- Nest `recommendations` under Kubernetes source (`sources.'alias'.kubernetes`)
- Post recommendations according to source configurations (full support will be introduced in https://github.com/kubeshop/botkube/issues/676)
- Move recommendations out of filters
- Add unit tests for all recommendations
- Improve recommendation messages
- Update E2E tests
- Fix receiving events on other namespaces
- Remove unused code

## Limitations

Currently, recommendations are not send if someone disables watching for Pod `create` and Ingress `create` events, even if all recommendations toggles are set to `true`. This is a behavior known, but will be fixed in a follow-up as a part of https://github.com/kubeshop/botkube/issues/676 or after (if it's complexity will be bigger).

## Testing

Install BotKube from this PR with all/part of the recommendations enabled:

Overrides:
```
communications:
  'default-group':
    slack:
      enabled: true
      token: {TOKEN}
      channels:
        'default':
          name: botkube-test
          bindings:
            executors:
              - kubectl-read-only
            sources:
              - k8s-events
settings:
  clusterName: dev
image:
  pullPolicy: Always
  tag: 687-PR
```
Command:
```
helm install botkube --namespace botkube ./helm/botkube -f /tmp/values.yaml --wait
```

Apply the following YAML:

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: failing
spec:
  containers:
  - name: nginx
    image: nginx:latest
    command: ["foo"]
---
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: ingress
spec:
  tls:
    - hosts:
        - foo.example.com
      secretName: foo
  rules:
    - host: "foo"
      http:
        paths:
          - path: "/testpath"
            pathType: Exact
            backend:
              service:
                name: "foo"
                port:
                  number: 3001
```

Now, turn some recommendations off during installation and apply the YAML again. 

## Related issue(s)

#684


pkg/recommendation/factory.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+	return newRecommendationsSet(f.logger, set)
+}
+
+func (f *Factory) createIfNotExist(condition bool, set map[string]Recommendation, constructorFn func() Recommendation) {
+	if !condition {
+		return
+	}
+
+	// this solution has a downside that we need to create an instance every time,
```
mszostok, line 61: **nit**, you can just use your own `tag` for `createIfNotExist` func:

```go
const ingressTLSRec = "ingres-tls-rec"
		f.createIfNotExist(ingressTLSRec, source.Recommendations.Ingress.TLSSecretValid, set,
			func() Recommendation { return NewIngressTLSSecretValid(f.dynamicCli) },
		)
```
or even
```go
const (
	IngressTLS Rec = iota
	PodLabels
	// ....    
)
```

pkg/recommendation/factory_test.go
```diff
@@ -0,0 +1,76 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/kubeshop/botkube/pkg/config"
+	"github.com/kubeshop/botkube/pkg/recommendation"
+)
+
+func TestFactory_NewForSources(t *testing.T) {
+	// given
+	sources := config.IndexableMap[config.Sources]{
```
mszostok, line 16: we should stop using the `config.IndexableMap` in places where `GetFirst()` is not needed.

pkg/recommendation/ingress_tls_secret_valid.go
```diff
@@ -0,0 +1,90 @@
+package recommendation
```
mszostok, line 1: I assume that the code logic was not changed, right? there is no need to "recheck" it?

mszostok (approved):
This PR LGTM as we will address issues in the follow-up. Current issues/limitations:
- if `recommendations` are enabled but the `kubernetes.resources` are not specified, then it won't work
- the `recommendations` are a separate "source" but it will be moved under `kubernetes` to make them more BotKube specific and associate directly with K8s as we are not sure about the final approach to "generic" recommendations.

pkg/recommendation/ingress_tls_secret_valid.go
pkosiec: Nope, I just simplified / slightly improved the code overall without changing the way how it works.

For Pod Labels Set I got rid of one snippet which wasn't necessary.

I tested all recommendations manually + implemented unit tests as well.

pkg/recommendation/factory.go
pkosiec: If you don't mind I would keep it as it is as for me is less error-prone aporoach from the extensibility point of view.

PR #691, opened 2022-08-16
Support routing notifications to a given channel based on Kubernetes Sources

### LINKED ISSUE #676: Support routing notifications to a given channel based on Kubernetes Sources

ezodude:
## Description

Resolves #676 

### Completed

- Routing notifications to channels via configured sources using the new `sources.Router`.
- A declarative approach to how informers are registered and setup.
- A declarative approach to easily clarify how events are handled including `error` events.
- A Source's Namespaces influence how a notification is mapped to a channel. 
- Removed now unnecessary `NamespaceChecker` filter from the filters list.

### Pending

- [x]  Handling `update` events. 
- [x] Sinks have no sources to collect yet.  
- [x] Add source bindings for Slack.
- [x] Add source bindings for MatterMost.
- [x] Add source bindings for Discord.
- [x] Add source bindings for Teams (it seems Teams bindings are not channel specific).
- [x] Merge in latest from main, including Recommendations.
- [x] Pass/fix existing unit and functional tests.


pkg/bot/discord.go
```diff
@@ -181,18 +182,19 @@ func (b *Discord) Type() config.IntegrationType {
...
 
 // TODO: Support custom routing via annotations for Discord as well
-func (b *Discord) getChannelsToNotify() []string {
-	// TODO(https://github.com/kubeshop/botkube/issues/596): Support source bindings - filter events here or at source level and pass it every time via event property?
-	var channelsToNotify []string
-	for _, channelCfg := range b.getChannels() {
-		if !channelCfg.notify {
-			b.log.Info("Skipping notification for channel %q as notifications are disabled.", channelCfg.Identifier())
-			continue
+func (b *Discord) getChannelsToNotify(eventSources []string) []string {
```
huseyinbabal, line 185: I see multiple `getChannelsToNotify` private methods in different bot implementations, but they are nearly doing same thing. (I see there is only one exception for teams, but maybe it can also use same logic). This method seems a good candidate for extracting to util package to be consumed by each bot, wdyt?

pkg/sink/webhook.go
```diff
@@ -70,7 +70,7 @@ func NewWebhook(log logrus.FieldLogger, c config.Webhook, reporter AnalyticsRepo
 }
 
 // SendEvent sends event notification to Webhook url
-func (w *Webhook) SendEvent(ctx context.Context, event events.Event) (err error) {
+func (w *Webhook) SendEvent(ctx context.Context, event events.Event, _ []string) (err error) {
```
huseyinbabal, line 73: I have a special comment here, maybe we can improve in the future. It seems not all the bot types requires event sources. To satisfy Single Responsibility Principle, instead of trying to satisfy Bot interface methods, we can come up with 2 intefaces as follows
```go
type Bot interface {
	Start(ctx context.Context) error
	IntegrationName() config.CommPlatformIntegration
	controller.
}

type SourcableBot interface {
	Bot
	SendEvent(ctx context.Context, event events.Event, sources []string)
}

type OutGoingBot interface {
	Bot
	SendEvent(ctx context.Context, event events.Event)
}
```

pkg/bot/discord.go
ezodude: Agreed. It's legacy code.

This can def be improved in the future.

pkg/sink/webhook.go
ezodude: Would have to dig deeper. But, ... yup... this is also legacy code and can be improved in the future.

pkg/sources/registration.go
```diff
@@ -0,0 +1,195 @@
...
+	informer        cache.SharedIndexInformer
+	log             logrus.FieldLogger
+	mapper          meta.RESTMapper
+	dynamicCli      dynamic.Interface
+	events          []config.EventType
+	mappedResources []string
+	mappedEvent     config.EventType
+}
+
+func (i Registration) handleEvent(ctx context.Context, resource string, target config.EventType, sourceRoutes []Route, fn eventHandler) {
```
huseyinbabal, line 29: Any reason for not using pointer receiver? Botkube is concurrent-heavy application, if this or parent function is called concurrently, concurrent changes will not be visible to original one. If you are sure about, you always read from receiver,  and never write to that good to go with non-pointer  receiver. Btw, if you are not sure about what kind of operations this receiver does, just use pointer receiver :) You can see Golang code review comments here https://github.com/golang/go/wiki/CodeReviewComments#receiver-type

pkg/sources/registration.go
```diff
@@ -0,0 +1,195 @@
...
+	}
+}
+
+func (i Registration) handleMapped(ctx context.Context, targetEvent config.EventType, routeTable map[string][]Entry, fn eventHandler) {
+	i.informer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj interface{}) {
+			var eventObj coreV1.Event
+			err := utils.TransformIntoTypedObject(obj.(*unstructured.Unstructured), &eventObj)
+			if err != nil {
+				i.log.Errorf("Unable to transform object type: %v, into type: %v", reflect.TypeOf(obj), reflect.TypeOf(eventObj))
```
huseyinbabal, line 70: Any reason for swallowing the original error here? Passing the returned error messages here would be valuable for better error chain visibility. Especially good for concurrent calls.

pkg/sources/registration.go
```diff
@@ -0,0 +1,195 @@
...
+			}
+			_, err = cache.MetaNamespaceKeyFunc(obj)
+			if err != nil {
+				i.log.Errorf("Failed to get MetaNamespaceKey from event resource")
```
huseyinbabal, line 74: swallowed error, and redundant formatted error log

pkg/sources/registration.go
```diff
@@ -0,0 +1,195 @@
...
+				return
+			}
+			fn(ctx, gvrToString, sources, nil)(obj)
+		},
+	})
+}
+
+func (i Registration) canHandleEvent(target string) bool {
+	for _, e := range i.events {
+		if strings.ToLower(target) == e.String() {
```
huseyinbabal, line 106: what about using `strings.EqualFold`? Here is an example https://go.dev/play/p/DdBgSd-P-Ja

pkg/sources/registration.go
```diff
@@ -0,0 +1,195 @@
...
+	}
+	return false
+}
+
+func sourcesForObjNamespace(ctx context.Context, routes []Route, obj interface{}, log logrus.FieldLogger, mapper meta.RESTMapper, cli dynamic.Interface) []string {
+	var out []string
+
+	objectMeta, err := utils.GetObjectMetaData(ctx, cli, mapper, obj)
+	if err != nil {
+		log.Errorf("while getting object metadata: %s", err.Error())
```
huseyinbabal, line 127: I always prefer to use `%+v` with `err` object since, `Error() string` method custom error implementations may not be always there.

pkg/sources/registration.go
```diff
@@ -0,0 +1,195 @@
...
+) ([]string, []string) {
+	var sources, diffs []string
+
+	candidates := sourcesForObjNamespace(ctx, routes, newObj, log, mapper, cli)
+
+	var oldUnstruct, newUnstruct *unstructured.Unstructured
+	var ok bool
+
+	if oldUnstruct, ok = oldObj.(*unstructured.Unstructured); !ok {
+		log.Error("Failed to typecast object to Unstructured.")
```
huseyinbabal, line 159: what about `Failed to typecast old object to Unstructured.` to separate it from next message?

pkg/sources/registration.go
```diff
@@ -0,0 +1,195 @@
...
+	}
+
+	if newUnstruct, ok = newObj.(*unstructured.Unstructured); !ok {
+		log.Error("Failed to typecast object to Unstructured.")
```
huseyinbabal, line 163: what about `Failed to typecast new object to Unstructured.` to separate it from previous message?

pkg/sources/sources.go
```diff
@@ -0,0 +1,249 @@
...
+	}
+}
+
+func (r *Router) MapWithEventsInformer(srcEvent config.EventType, dstEvent config.EventType, handler registrationHandler) {
+	srcResources := r.resourcesForEvents([]config.EventType{srcEvent})
+	if len(srcResources) == 0 {
+		return
+	}
+
+	dstResource := "v1/events"
```
huseyinbabal, line 167: I am a bit concerned about static resource names since, if there is a change to those versions between k8s cluster versions, botkube will stop working even it can handle those resources even version is changed.

ezodude: Any ideas?

pkg/utils/slices.go
```diff
@@ -0,0 +1,16 @@
+package utils
+
+import (
+	"strings"
+)
+
+func Intersect(this, that []string) bool {
+	for _, i := range this {
+		for _, j := range that {
+			if strings.ToLower(i) == strings.ToLower(j) {
```
huseyinbabal, line 10: you can use `strings.EqualFold`

test/e2e/slack_test.go
```diff
@@ -312,8 +311,6 @@ func TestSlack(t *testing.T) {
 					return strings.Contains(msg.Text, heredoc.Doc(fmt.Sprintf("Cluster: %s", appCfg.ClusterName))) &&
 						strings.Contains(msg.Text, "local-path-provisioner") &&
 						strings.Contains(msg.Text, "coredns") &&
-						strings.Contains(msg.Text, "traefik") &&
```
huseyinbabal, line 315: I also mentioned this part on slack, https://kubeshop.slack.com/archives/C03MRCX7UE9/p1660819668105749

huseyinbabal:
@ezodude thank you for your work, as a general note, will you add anything to slack e2e for newly introduced changes? I think there is a huge change in the event sourcing mechanism, better to verify that in e2e tests. Thanks 👍

pkg/bot/slack.go
```diff
@@ -289,23 +290,24 @@ func (b *Slack) SendEvent(ctx context.Context, event events.Event) error {
...
-			continue
+	var out []string
+	for _, cfg := range b.getChannels() {
+		switch {
+		case !cfg.notify:
+			b.log.Info("Skipping notification for channel %q as notifications are disabled.", cfg.Identifier())
+		default:
+			if utils.Intersect(eventSources, cfg.Bindings.Sources) {
+				out = append(out, cfg.Identifier())
+			}
```
pkosiec, line 307: I think that's easier to read - and we have less indentations:
```suggestion
		if !cfg.notify {
			b.log.Info("Skipping notification for channel %q as notifications are disabled.", cfg.Identifier())
			continue
		}

		if utils.Intersect(eventSources, cfg.Bindings.Sources) {
			out = append(out, cfg.Identifier())
		}
```

pkg/utils/slices.go
```diff
@@ -0,0 +1,16 @@
+package utils
+
+import (
+	"strings"
+)
+
+func Intersect(this, that []string) bool {
```
pkosiec, line 7: Can we move that out of `utils` package? I think it's better to have packages with more meaningful names. Here, I think we should go with `slices` or `sliceutil`, or something else 🤔

BTW It would be good to have a simple unit test for that.

test/e2e/slack_test.go
```diff
@@ -116,7 +116,6 @@ func TestSlack(t *testing.T) {
 		command := "filters list"
```
pkosiec, line 116: As Huseyin wrote, it would be great to cover that in E2E tests. We have two Slack channels, so we can easily test different bindings 👍

pkg/filterengine/with_all_filters.go
```diff
@@ -15,11 +15,10 @@ const (
 )
 
 // WithAllFilters returns new DefaultFilterEngine instance with all filters registered.
-func WithAllFilters(logger *logrus.Logger, dynamicCli dynamic.Interface, mapper meta.RESTMapper, res []config.Resource) *DefaultFilterEngine {
+func WithAllFilters(logger *logrus.Logger, dynamicCli dynamic.Interface, mapper meta.RESTMapper, _ []config.Resource) *DefaultFilterEngine {
```
pkosiec, line 18: If we don't need `res []config.Resource`, don't pass it 🙂 

```suggestion
func WithAllFilters(logger *logrus.Logger, dynamicCli dynamic.Interface, mapper meta.RESTMapper) *DefaultFilterEngine {
```

pkg/sink/elasticsearch.go
```diff
@@ -172,7 +172,7 @@ func (e *Elasticsearch) flushIndex(ctx context.Context, indexCfg config.ELSIndex
 }
 
 // SendEvent sends event notification to Elasticsearch
-func (e *Elasticsearch) SendEvent(ctx context.Context, event events.Event) (err error) {
+func (e *Elasticsearch) SendEvent(ctx context.Context, event events.Event, _ []string) (err error) {
```
pkosiec, line 175: Shouldn't Elasticsearch also use the `eventSources`? For outgoing webhook it makes sense to not use it, but Elasticsearch supports multiple indices, so... 🤔

pkg/sources/registration.go
pkosiec: I agree, while previously BotKube usually just logged errors, in new code we should return them as an output argument.

pkg/sources/registration.go
pkosiec: Can we return error from this function as well? BTW we can convert it to method, same with `qualifySourcesForUpdate`.

pkg/sources/sources.go
```diff
@@ -0,0 +1,249 @@
...
+			}
+			for _, e := range flattenEvents(resource.Events) {
+				out[resource.Name][e] = struct{}{}
+			}
+		}
+	}
+	return out
+}
+
+func mergeEventRoutes(resource string, sources config.IndexableMap[config.Sources]) map[config.EventType][]Route {
```
pkosiec, line 110: Can they be converted to methods as well?

pkg/sources/sources.go
```diff
@@ -0,0 +1,249 @@
...
+				case resource == r.Name && e == config.UpdateEvent:
+					out[e] = append(out[e], Route{
+						source:     srcGroupName,
+						namespaces: r.Namespaces,
+						updateSetting: config.UpdateSetting{
+							Fields:      r.UpdateSetting.Fields,
+							IncludeDiff: r.UpdateSetting.IncludeDiff,
+						}})
+				case resource == r.Name && e != config.UpdateEvent:
+					out[e] = append(out[e], Route{source: srcGroupName, namespaces: r.Namespaces})
```
pkosiec, line 125: We can simplify this:


```suggestion
				if resource != r.Name {
					continue
				}

				route := Route{source: srcGroupName, namespaces: r.Namespaces}

				if e == config.UpdateEvent {
					route.updateSetting = config.UpdateSetting{
						Fields:      r.UpdateSetting.Fields,
						IncludeDiff: r.UpdateSetting.IncludeDiff,
					}
				}
				
				out[e] = append(out[e], route)
```

pkg/sources/sources.go
```diff
@@ -0,0 +1,249 @@
...
+	for evt := range events {
+		if _, ok := r.table[resource]; !ok {
+
+			r.table[resource] = []Entry{{
+				event:  evt,
+				routes: routes[evt],
+			}}
+
+		} else {
+			r.table[resource] = append(r.table[resource], Entry{event: evt, routes: routes[evt]})
```
pkosiec, line 143: I think it would work if we simplify this and append to nil slice 🤔  But please test it:
```suggestion
		r.table[resource] = append(r.table[resource], Entry{event: evt, routes: routes[evt]})
```

pkg/sources/sources.go
huseyinbabal: I prefer to do;
- Collect kubernetes built in resources during application startup and put them in a data structure.
- Refer to that variable to see what are the resource names
Again, this is not a release blocker, just wanted to point out for background compatibility.

pkosiec: Calling K8s API client to get the api group and version of `events` would still require us to have some assumptions like resource name + API Group. And also, I believe the v1 K8s API will stay there for a long time with us, so IMHO moving it to consts to the top of this file would be enough for now 👍

pkg/sources/sources.go
```diff
@@ -0,0 +1,249 @@
...
+func (r *Router) mappedInformer(event config.EventType) (Registration, bool) {
+	for _, informer := range r.registrations {
+		if informer.mappedEvent == event {
+			return informer, true
+		}
+	}
+	return Registration{}, false
+}
+
+func flattenEvents(events []config.EventType) []config.EventType {
```
pkosiec, line 239: Could be a method as well 🙂

pkg/sources/sources.go
```diff
@@ -0,0 +1,249 @@
...
+				routes: routes[evt],
+			}}
+
+		} else {
+			r.table[resource] = append(r.table[resource], Entry{event: evt, routes: routes[evt]})
+		}
+	}
+}
+
+func (r *Router) RegisterInformers(targetEvents []config.EventType, handler registrationHandler) {
```
pkosiec, line 148: 🤔 Why are the event types are passed as an input argument, but the resources are loaded from inside of the function? For me it's a bit unintuitive - could you please explain the reason about such approach?

pkg/sources/sources.go
```diff
@@ -0,0 +1,249 @@
...
+		mapper:          r.mapper,
+		dynamicCli:      r.dynamicCli}
+}
+
+func (r *Router) HandleEvent(ctx context.Context, target config.EventType, handlerFn eventHandler) {
+	for resource, informer := range r.registrations {
+		if informer.canHandleEvent(target.String()) {
+			sourceRoutes := r.sourceRoutes(resource, target)
+			informer.handleEvent(ctx, resource, target, sourceRoutes, handlerFn)
+		}
```
pkosiec, line 183: One nesting less 🙂  Same can be done with `HandleMappedEvent`.
```suggestion
		if !informer.canHandleEvent(target.String()) {
			continue
		}
		
		sourceRoutes := r.sourceRoutes(resource, target)
		informer.handleEvent(ctx, resource, target, sourceRoutes, handlerFn)
```

pkosiec (commented):
First batch of comments after code review 🙂 

Generally looks good! It would be great to:
- support source bindings for Elasticsearch
- return errors from most of the functions/methods
- have E2E tests
- have unit tests for new parts of code.

I will test it shortly and let you know if everything works as expected 🤞

pkg/utils/slices.go
mszostok, line 7: I saw that it's used only for sources. What's more sources are used only with this method. Can we just create a type with our method?

```go

type Sources map[string]struct{}

func (s Sources) Intersect(with []string) bool {
	if s == nil {
		return false
	}
	
	for _, name := range with {
		if _, found := s[strings.ToLower(name)]; found {
			return true
		}
	}
	return false
}
```

a small benefit is that we would not require 2 `for` loops

pkg/sink/webhook.go
pkosiec: As I pointed in https://github.com/kubeshop/botkube/pull/691#discussion_r950047949, Elasticsearch should also support bindings, same with outgoing webhook 🙂 So that would mean this comment is no longer applicable.

pkosiec:
Tested Slack, Discord and Elasticsearch. Slack and Discord work well 👍 Elasticsearch doesn't have the sources support implemented, same with Webhook. Please support them as well, as it is defined in our config 👍 Thanks!

mszostok (commented):
overall good jobs 🚀

pkg/sources/registration.go
pkosiec: True 👍 But I also saw in other places logging error instead of returning. Let's make sure we always return these errors, unless we can't 🙂

ezodude: For sure, we should always keep an eye for that. But, as @mszostok `AddFunc` stops this here.

pkg/sink/webhook.go
ezodude: Elasticsearch and outgoing webhook source bindings now added.

pkg/sink/elasticsearch.go
ezodude: Elasticsearch source bindings now added.

pkg/sources/sources.go
ezodude: This marks which events you're looking to target for this informer registration.

ezodude: For example, here you don't want to target the `errors` event type as that is handles differently.

pkg/sources/registration.go
ezodude: Yup, for now this is a pure data object. Not wanting to be heavy handed.

Happy to change if the context changes in the future.

pkg/sources/registration.go
ezodude: No reason at all.

The code was moved from another section in `pkg/controller/controller.go`. When error returns fit they will be returned 😄.

pkg/sources/registration.go
ezodude: Done.

pkg/sources/registration.go
ezodude: Actually after digging in, notice that the container func `AddFunc: func(obj interface{}) {` defines no error returns.

Meaning, the error cannot be returned.

pkg/sources/registration.go
ezodude: Again, code moved from controller.go. Now updated 👌.

pkg/utils/slices.go
ezodude: So, I completely agree with your thinking about `utils` packages. I don't like them.

In this case, I was going for the principle of least surprise. There is similar/general code in utils so I moved it there.

pkosiec: I understand 👍 If you could move it out of there that'd be great - let's reduce the functions from there instead of adding new ones 🙂

pkg/sources/sources.go
ezodude: Interesting. The `entry` wrapper struct was introduced to work with the `update` event type. It simplified everything. 

In fact, you no longer need this function at all 👍 ... less code FTW.

ezodude:
@pkosiec a quick note regarding the comments to convert functions to methods:

Generally, I try to keep the code as precise, decoupled and testable as possible. So, in most instances this means "functions over methods".

I use methods, when I need to immediately read / update state. Otherwise, I delegate to functions.

Functions have explicit inputs and outputs. They are easier to understand and test without worrying about messing up state.

I hope this makes sense. Happy to discuss further.

Saying that, I'll def add unit tests to cover the routing functions where possible. 🙂

pkg/controller/controller.go
```diff
@@ -108,13 +91,76 @@ func New(log logrus.FieldLogger,
...
+				c.sendEvent(ctx, obj, resource, config.DeleteEvent, sources, nil)
+			}
+		})
+
+	c.sourcesRouter.HandleEvent(
+		context.Background(),
+		config.UpdateEvent,
+		func(ctx context.Context, resource string, sources []string, updateDiffs []string) func(obj interface{}) {
+			return func(obj interface{}) {
+				c.log.Debugf("Processing update to resource: %q, event: %q, sources: %+v\nobject: %+v", resource, config.UpdateEvent, sources, obj)
```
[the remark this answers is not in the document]
ezodude: I like very much ☝️

pkg/sources/registration.go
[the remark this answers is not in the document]
ezodude: oops.

pkg/sources/registration.go
```diff
@@ -0,0 +1,195 @@
...
+	for _, route := range routes {
+		if route.namespaces.IsAllowed(targetNs) {
+			out = append(out, route.source)
+		}
+	}
+
+	return out
+}
+
+func qualifySourcesForUpdate(
```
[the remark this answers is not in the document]
ezodude: Already handled from previous comments.

pkg/utils/slices.go
ezodude: I see what you mean. 🙂

But given the massive changes in this PR, I'm happy to take this in a future refactoring - if that's ok.

ezodude:
@huseyinbabal @pkosiec @mszostok 

I responded to all feedback with most if it implemented including: `webhook` and `elasticsearch` integration.

Feel free to have a look or wait until the pending items are completed (they're just addendums to the feedback).

Pending:
- New E2E test to work with 2 channels.
- Some minor unit testing..

pkg/sources/registration.go
pkosiec: > Generally, I prefer functions over methods as they can be used by any one. 

I agree, but are they used anywhere else? Will someone else use them? If not, then IMHO they should be not exported methods 🙂 To me, they look exactly as a part of private logic inside the `registration` struct.

But don't treat is as a blocking comment, I'm fine with different opinion on that 👍

pkg/bot/slack.go
pkosiec: Marking this comment as unresolved, as I don't see a place where it was applied 🤔 This is a minor comment, but consider simplifying the code as suggested in all bots 🙂

ezodude: Will have a look. Thanks for the heads up.

ezodude: It's a matter of style. I'm happy with guard clauses. Updated.
(running tests locally to double check all is well)

helm/botkube/e2e-test-values.yaml
```diff
@@ -20,7 +20,7 @@ communications:
             executors:
```
pkosiec, line 20: Please update the `values.yaml` file for Helm chart to indicate the webhook also supports bindings:

```yaml
    ## Settings for Webhook.
    webhook:
      # -- If true, enables Webhook.
      enabled: false
      # -- The Webhook URL, e.g.: https://example.com:80
      url: 'WEBHOOK_URL'
      bindings:
        # -- Notification sources configuration for the webhook.
        sources:
          - k8s-events
```

Then, regenerate the readme for Helm chart. Thanks!

mszostok (approved):
LGTM 🚀

helm/botkube/e2e-test-values.yaml
ezodude: Make sense.

ezodude: Done.

huseyinbabal (approved):

pkg/bot/slack.go
pkosiec: Thanks for applying it for Slack. I think it looks better as there's less indentations now.

> consider simplifying the code as suggested **in all bots** 🙂

Please consider to apply it for all other places 🙂

pkosiec (approved):
Tested Slack and Discord, works well 👍

PR #696, opened 2022-08-22
 Handle recommendations even if no such resources are configured for notifications

### LINKED ISSUE #684: Add ability to configure specific recommendations during BotKube installation/upgrade

pkosiec:
## Description

-  Handle recommendations even if no such resources are configured for notifications

Now the recommendations are passed to all configured

**NOTE:** The recommendations don't have configurable namespaces. If they are enabled, BotKube notifies about Pod/Ingress resources from all namespaces if they contain any warning/recommendation.

## Testing

The changes has been tested in E2E tests as well as unit ones.

You can use sample config:
```yaml
sources:
  'k8s-events':

    kubernetes:

      recommendations:
        pod:
          noLatestImageTag: true
          labelsSet: true
        ingress:
          backendServiceValid: true
          tlsSecretValid: true

      resources: []

  'k8s-events2':

    kubernetes:

      recommendations:
        pod:
          noLatestImageTag: true
          labelsSet: true
        ingress:
          backendServiceValid: true
          tlsSecretValid: true

      resources:
        - name: v1/pods
          namespaces:
            include:
              - ".*"
          events:
            - create
            - delete
            - error

  'k8s-events3':

    kubernetes:

      recommendations:
        pod:
          noLatestImageTag: true
          labelsSet: true
        ingress:
          backendServiceValid: false
          tlsSecretValid: false

      resources:
        - name: v1/pods
          namespaces:
            include:
              - ".*"
          events:
            - delete
            - error
communications:
  'default-group':
    # Settings for Slack
    slack:
      enabled: true
      channels:
        'alias':
          name: 'botkube-test'
          bindings:
            executors:
              - 'kubectl-read-only'
            sources:
              - 'k8s-events'
        'alias2':
          name: 'general'
          bindings:
            executors:
              - 'kubectl-read-only'
            sources:
              - 'k8s-events2'
        'alias3':
          name: 'random'
          bindings:
            executors:
              - 'kubectl-read-only'
            sources:
              - 'k8s-events3'
      token: "{TOKEN}"
      notification:
        type: short
```

Apply recommendations file:

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: failing
spec:
  containers:
  - name: nginx
    image: nginx
    command: ["foo"]
---
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: ingress
spec:
  rules:
    - host: "foo"
      http:
        paths:
          - path: "/testpath"
            pathType: Exact
            backend:
              service:
                name: "foo"
                port:
                  number: 3001
```

Create and delete the resources:
```bash
kubectl create -f /tmp/recommendations.yaml
kubectl delete -f /tmp/recommendations.yaml
```
Observe the events.

## Related issue(s)

Resolves #684

### DIFF
--- a/helm/botkube/e2e-test-values.yaml
+++ b/helm/botkube/e2e-test-values.yaml
@@ -25,6 +25,13 @@ communications:
 sources:
   'k8s-events':
     kubernetes:
+      recommendations:
+        pod:
+          noLatestImageTag: true
+          labelsSet: true
+        ingress:
+          backendServiceValid: false
+          tlsSecretValid: false
       resources:
         - name: v1/configmaps
           namespaces:
@@ -34,12 +41,6 @@ sources:
             - create
             - update
             - delete
-        - name: v1/pods
-          namespaces:
-            include:
-              - botkube
-          events:
-            - create
   'k8s-updates':
     kubernetes:
       resources:
--- a/pkg/recommendation/resource_events.go
+++ b/pkg/recommendation/resource_events.go
@@ -8,7 +8,7 @@ import (
 
 const (
 	podsResourceName    = "v1/pods"
-	ingressResourceName = "networking.k8s.io/v1/ingress"
+	ingressResourceName = "networking.k8s.io/v1/ingresses"
 )
 
 // ResourceEventsForConfig returns the resource event map for a given source recommendations config.
--- a/pkg/recommendation/resource_events.go
+++ b/pkg/recommendation/resource_events.go
@@ -1,0 +1,64 @@
+package recommendation
+
+import (
+	"github.com/kubeshop/botkube/pkg/config"
+	"github.com/kubeshop/botkube/pkg/events"
+	"github.com/kubeshop/botkube/pkg/ptr"
+)
+
+const (
+	podsResourceName    = "v1/pods"
+	ingressResourceName = "networking.k8s.io/v1/ingress"
+)
+
+// ResourceEventsForConfig returns the resource event map for a given source recommendations config.
+func ResourceEventsForConfig(recCfg config.Recommendations) map[string]config.EventType {
+	resNames := make(map[string]config.EventType)
+
+	if ptr.IsTrue(recCfg.Ingress.TLSSecretValid) || ptr.IsTrue(recCfg.Ingress.BackendServiceValid) {
+		resNames[ingressResourceName] = config.CreateEvent
+	}
+
+	if ptr.IsTrue(recCfg.Pod.NoLatestImageTag) || ptr.IsTrue(recCfg.Pod.LabelsSet) {
+		resNames[podsResourceName] = config.CreateEvent
+	}
+
+	return resNames
+}
+
+// ShouldIgnoreEvent returns true if user doesn't listen to events for a given resource, apart from enabled recommendations.
+func ShouldIgnoreEvent(recCfg config.Recommendations, sources map[string]config.Sources, sourceBindings []string, event events.Event) bool {
+	if event.HasRecommendationsOrWarnings() {
+		// shouldn't be skipped
+		return false
+	}
+
+	res := ResourceEventsForConfig(recCfg)
+	recommEventType, ok := res[event.Resource]
+	if !ok {
+		// this event doesn't relate to recommendations, finish early
+		return false
+	}
+
+	if event.Type != recommEventType {
+		// this event doesn't relate to recommendations, finish early
+		return false
+	}
+
+	// Resource + event type matches the ones configured from recommendation.
+	// Check if user listens to this event.
+	for _, key := range sourceBindings {
+		source, exists := sources[key]
+		if !exists {
+			continue
+		}
+
+		// sources are appended, so we need to check the first source that has a given resource with event
+		if source.Kubernetes.Resources.IsAllowed(event.Resource, event.Namespace, event.Type) {
+			return false
+		}
+	}
+
+	// The event is related to recommendations informers. No recommendations there, so it should be skipped.
+	return true
+}
--- a/pkg/sources/sources.go
+++ b/pkg/sources/sources.go
@@ -9,6 +9,7 @@ import (
 	"k8s.io/client-go/tools/cache"
 
 	"github.com/kubeshop/botkube/pkg/config"
+	"github.com/kubeshop/botkube/pkg/recommendation"
 )
 
 const eventsResource = "v1/events"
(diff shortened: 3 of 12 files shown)


helm/botkube/e2e-test-values.yaml
```diff
@@ -34,12 +41,6 @@ sources:
             - create
             - update
             - delete
-        - name: v1/pods
-          namespaces:
-            include:
-              - botkube
-          events:
-            - create
```
pkosiec, line 42: Not needed anymore, as enabled recommendations will also send a notification 🙂 In that way we're testing the changes from this PR.

pkg/recommendation/resource_events.go
```diff
@@ -0,0 +1,64 @@
...
+
+	if ptr.IsTrue(recCfg.Pod.NoLatestImageTag) || ptr.IsTrue(recCfg.Pod.LabelsSet) {
+		resNames[podsResourceName] = config.CreateEvent
+	}
+
+	return resNames
+}
+
+// ShouldIgnoreEvent returns true if user doesn't listen to events for a given resource, apart from enabled recommendations.
+func ShouldIgnoreEvent(recCfg config.Recommendations, sources map[string]config.Sources, sourceBindings []string, event events.Event) bool {
```
ezodude, line 30: Reading this here makes sense, but over in the controller it's a bit confusing why an event is ignored.

Can we rename this to something that better explains the intent?

ezodude (changes requested):
All in all makes sense.

Two genuine comments with a couple of nit picks.

pkg/sources/sources.go
```diff
@@ -222,10 +232,48 @@ func mergeEventRoutes(resource string, sources config.IndexableMap[config.Source
...
 }
 
+func setEventRouteForRecommendationsIfShould(routeMap *map[config.EventType][]route, resForRecomms map[string]config.EventType, srcGroupName, resourceName string) {
+	if routeMap == nil {
+		return
+	}
+
+	eventType, ok := resForRecomms[resourceName]
+	if !ok {
+		// no recommendation for this resource
```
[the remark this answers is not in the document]
pkosiec: I will remove the comment and rename this to `found`, as it's more usual to name this variable like this 👍

pkg/sources/sources.go
```diff
@@ -222,10 +232,48 @@ func mergeEventRoutes(resource string, sources config.IndexableMap[config.Source
...
+		setEventRouteForRecommendationsIfShould(&out, resForRecomms, srcGroupName, resource)
 	}
+
 	return out
 }
 
+func setEventRouteForRecommendationsIfShould(routeMap *map[config.EventType][]route, resForRecomms map[string]config.EventType, srcGroupName, resourceName string) {
+	if routeMap == nil {
+		return
+	}
```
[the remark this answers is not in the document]
pkosiec: This map shouldn't be nil, but it's a good practice to ensure if a pointer is not nil until you access its value to avoid panics.

Theoretically I could return error from this function, but then it would need to be passed up to the `BuildTable` method (3 levels to top). Given the fact that the only error case is when a dev makes a mistake, I don't think this is reasonable to do.

Also, as the function name says `IfShould`, that means IMHO it's ok to not do anything if the `routeMap` is nil. That's why I went with such approach.

I'm open for alternative ideas though 🙂

pkg/recommendation/resource_events.go
pkosiec: I believe the function says exactly what it does (tells if an event should be ignored), and the comment explains when a given event could be ignored. Also, it is in the `recommendation` package so it tells that it is related somehow to recommendations. Honestly, I don't have any other ideas which are not too long. Do you have any? 🙂 

---

Alternatively, WDYT about such change:

```go
	if recommendation.ShouldIgnoreEvent(recCfg, c.conf.Sources, sources, event) {
		c.log.Debugf("Skipping event as it is related to recommendation informers and doesn't have any recommendations: %#v", event)
		return
	}
```

I think that explains better what's happening. Thoughts?

pkg/sources/sources.go
```diff
@@ -199,11 +200,20 @@ func mergeResourceEvents(sources config.IndexableMap[config.Sources]) mergedEven
 				out[resource.Name][e] = struct{}{}
 			}
 		}
+
+		// add resource events related to recommendations
```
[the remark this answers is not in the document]
pkosiec: 🤔 Generally I'm a fan of extracting some parts to separate methods/functions (that's why I defined `recommendation.ResourceEventsForConfig` as a separate function). However, here, if we're building a single resource event map, I think it's easier to understand the whole logic if it's encapsulated in as single function, especially that it's all about a few lines (8 and 7 lines for these two parts). I would need to take the pointer of the map and modify it in the extracted functions, which IMO it's harder to read.

If that's fine for you, I would keep it as it is. Removed the comment though 🙂

pkg/recommendation/resource_events.go
ezodude: Thinking out aloud (to confirm):

... The function ensures the event should be skipped if the recommendation event has no recommendation?
 
You have a recommendation "hit" and a recommendation "miss". All I can offer is: `recommendation.Miss(...)`.

The idea is not to have to dig into the guts of the function to figure out why we are returning.

The Log definitely helps. Not a blocker to shipping this.

pkg/sources/sources.go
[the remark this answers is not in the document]
ezodude: Personally, I think it might be too defensive and if a map is nil, when it's not supposed to be, the runtime will never tell you.

You'll end in a false sense security. But ... def not a blocker. 👍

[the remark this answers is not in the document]
pkosiec: > the runtime will never tell you.

I understand your point of view and I agree - yes, that's the downside of this if 👍 The upside is we'll avoid some panic which will crash the app. I came up with another possiblity: WDYT about converting these functions to methods and then use debug log to notify that the function was skipped because the map is nil?

pkg/recommendation/resource_events.go
pkosiec: >... The function ensures the event should be skipped if the recommendation event has no recommendation?

Yep, that's very nice explanation what's happening here 👍 It would be great to highlight that a given function returns bool and also, as you wrote, it's about events 🤔 It can be something with `Miss`, but I find it a bit too short (it doesn't give me the same idea as `ShouldIgnoreEvent`: `Should` <- returns true/false, `Event` <- it's about events 🤔 `recommendation.Miss` could be understood in different way - `Miss` is when there's no recommendation for a given event 🤔

ezodude: Cool - makes sense. Leave as is.

ezodude (approved):

PR #699, opened 2022-08-24
Generate only draft release, and process chart during release

mszostok:
## Description

- Generate only draft release, and process chart during release


pkosiec (approved):
LGTM

PR #721, opened 2022-09-05
Mattermost v7.0 support added

### LINKED ISSUE #673: Update Bot integration to work with MatterMost v7+.
## Overview

MatterMost bot integration API has changed for version v7+.

BotKube's integration works well with lower versions but breaks with this recent update.

This results in the following error:
_(tested against MatterMost Cloud running the latest version of MatterMost)_

```shell
2022/08/03 13:32:27 while creating Mattermost bot: : Unable to find the existing team.,
```

## Acceptance Criteria

- Successful MatterMost v7+ Bot API integration.
- Documentation updates to simplify integrating with MatterMost v7+.

## Reason

If this is not updated, BotKube users will only be able to use MatterMost versions lower than v7.

huseyinbabal:
## Description

- Mattermost v7.0 support added
- Client is upgraded to v6 which is the latest version as you can see [here](https://pkg.go.dev/search?q=mattermost-server)

The root of cause of this issue was, what we see for teams in Mattermost v7 Console refers to team's display name not name. 
Screen Shot 2022-09-05 at 18 26 06

That's why we started to get team not found error. Now it works as expected 👇 

Screen Shot 2022-09-05 at 18 18 03
 

## Related issue(s)
Fixes #673


pkg/bot/mattermost.go
```diff
@@ -91,11 +91,20 @@ func NewMattermost(log logrus.FieldLogger, cfg config.Mattermost, executorFactor
 	client := model.NewAPIv4Client(cfg.URL)
 	client.SetOAuthToken(cfg.Token)
 
-	botTeam, resp := client.GetTeamByName(cfg.Team, "")
+	botTeams, resp := client.SearchTeams(&model.TeamSearch{
+		Term: cfg.Team,
+	})
```
ezodude, line 96: What you're doing here makes sense. But, I have a quick question: shouldn't this just be a documentation update?

That is, can we just ask users to add the Team name, not display name, in `values.yaml` - or is it hard to obtain that the actual team name?

huseyinbabal: The problematic part is, Mattermost does not expose `name` field to UI. Of course we can say user users to execute the curl request I provided in PR description, but it adds additional complexity to flow. I thought same as you did but it was my findings, Wdyt?

ezodude: Thx. You've answered my question. Works for me.

ezodude (approved):

PR #722, opened 2022-09-06
Disables analytics reporting when running e2e tests locally

ezodude:
## Description

- Ensures analytics reporting is disabled when running e2e tests locally.
- Removes redundant CI analytics reporting setup.


helm/botkube/e2e-test-values.yaml
```diff
@@ -106,6 +106,13 @@ settings:
 
 extraAnnotations:
   botkube.io/disable: "true"
+
+## Parameters for anonymous analytics collection.
+analytics:
+  # -- If true, sending anonymous analytics is disabled. To learn what date we collect,
+  # see [Privacy Policy](https://botkube.io/privacy#privacy-policy).
+  disable: true
```
pkosiec, line 114: We can remove the comments:

```suggestion
analytics:
  disable: true
```

pkosiec (commented):
Currently the value is provided here: https://github.com/kubeshop/botkube/blob/main/.github/workflows/branch-build.yml#L92 - could you please remove the `--set` from there to avoid duplication?

helm/botkube/e2e-test-values.yaml
ezodude: I left them on purpose.

ezodude: As it's an important setting.

pkosiec: I just think of consistency, as we don't have other comments anywhere else in this file, as they are overrides. Especially that the comments are copied with special characters (e.g. `# --`) and Markdown syntax which are used to generate our docs 🙂 We have all values documented in the main `values.yaml` file. 

Of course this is non-blocking comment, but it would be great to sort it out. I see two main options:
- remove these comments
- remove the syntax for docs:

```
# Parameters for anonymous analytics collection.
analytics:
  # If true, sending anonymous analytics is disabled. To learn what date we collect,
  # see Privacy Policy: https://botkube.io/privacy#privacy-policy.
```

Please apply the changes before merge 🙂

pkosiec (approved):
LGTM, but before merge please resolve this comment in either way: https://github.com/kubeshop/botkube/pull/722#discussion_r963358849 👍

pkosiec (approved):
Go go go!

PR #725, opened 2022-09-07
Add interactive Help message for Slack platform

mszostok:
This depends on https://github.com/kubeshop/botkube/pull/693 which needs to be merged first.

## Description

Changes proposed in this pull request:

- Add interactive Help message for Slack platform

Implemented on this PR:
- [x] Help is shown when BotKube is installed and links to a Slack channel (first connection)
	- [x] public message shown for everyone only once, right after installation for configured channels (initial config)
	- [x] if the bot is invited to other channel, the message won't be printed
- [x] The help dialog shows the dialog as per the design
- [x] The buttons run the appropriate commands

## Testing

To run e2e test locally:

1. Create cluster: `k3d cluster create`
2. Checkout PR: `gh pr checkout 725`
3. Install BotKube: 
    ```bash
    export SLACK_BOT_TOKEN="xoxb-"
    export SLACK_APP_TOKEN="xapp-1-"
    
    helm upgrade botkube --install --namespace botkube ./helm/botkube --wait --create-namespace \
	    -f ./helm/botkube/e2e-test-values.yaml \
	    --set communications.default-group.slack.botToken="${SLACK_BOT_TOKEN}" \
	    --set communications.default-group.slack.appToken="${SLACK_APP_TOKEN}" \
	    --set image.repository=kubeshop/pr/botkube \
	    --set image.tag=725-PR \
	    --set image.pullPolicy="Always" \
	    --set analytics.disable=true
    ```
4. Run tests:
    ```bash
    export KUBECONFIG="config"
    export SLACK_BOT_NAME="botkube"
    export SLACK_TESTER_NAME="tester"
    export SLACK_ADDITIONAL_CONTEXT_MESSAGE="local"
    export SLACK_TESTER_APP_TOKEN="xoxb-"
    make test-integration
    ```

## TODO

In follow-up, I will:
- [ ] Help is shown when BotKube is installed and links to a Slack channel (first connection)
	- [ ] save state to not show it per app restart
- [ ] A user runs `@BotKube help`
- [ ] Print a command header (so we can easily correlate the output)
- [ ] Print a plaintext message for others
- [ ] Add source bindings customization

## Related issue(s)

- #683

Issue #683, opened 2022-08-10
Show interactive help command to the user


mszostok:
## Overview

Currently, when an invalid command is executed or when only `@BotKube` is typed, you get:

![Screen Shot 2022-08-09 at 12 44 52](https://user-images.githubusercontent.com/17568639/183629548-8b3cd71e-2988-441a-8649-f7e5dc7cdeee.png)

In the response, you get at least info that `/botkubehelp` will tell you more about possible commands. 

**Cons about such approach:**
- you need to type one more thing
- you get the response in a `plain text`. The best what you can do, is to copy-paste that command.
    ![Screen Shot 2022-08-09 at 12 46 49](https://user-images.githubusercontent.com/17568639/183629895-9c62cd14-fd84-46e5-8616-b0742726bacf.png)

## Desing

Slack design: https://www.figma.com/proto/iGGgUG9vn6C9WvDAFwKS9G/Botkube-Help---Slack?page-id=209%3A1337&node-id=269%3A1038&viewport=282%2C271%2C0.38&scaling=min-zoom&starting-point-node-id=269%3A1038

### Outdated design

![](https://user-images.githubusercontent.com/17568639/183634228-3014d5d3-183d-46f2-98bd-6d7343d61e6e.png)

## Acceptance criteria

- Help is shown when BotKube is installed and links to a Slack channel (first connection)
  - save state to not show it per app restart 
  - public message shown for everyone only once, right after installation for configured channels (initial config)
  - if the bot is invited to other channel, the message won't be printed
- A user runs `@BotKube help`
- The help dialog shows the dialog as per the design
- The buttons run the appropriate commands

## Out of scope

- A new user joins a BotKube channel (shown only to that user)
   - Reason: BotKube can be configured for multiple channels
- Notification presets configuration

## Notes

- Unknown: whether we can write E2E tests that check if the interactive buttons work
  - if Slack Go client doesn't have such method, skip that part - test the `@BotKube test` command only
- All messages are visible for everyone
- Assume #631 is resolved (merged into `main`)
- Initial focus on Slack
- Follow-up in future releases
- Ensure telemetry is in place 
- **it's ok that if you have multiple cluster on the same channel, then all return helper message**

mszostok:
Related:
- https://github.com/kubeshop/botkube/issues/681

mszostok:
If agreed, introduce the dedicated `@Botkube help` command as describe in https://github.com/kubeshop/botkube/issues/393

[...]

PR #727, opened 2022-09-08
Add command to get interactive help message

mszostok:
## Description

> **Warning**
> This depends on https://github.com/kubeshop/botkube/pull/725 which needs to be merged first.

- [x] A user can run `@BotKube help` on **all** platforms, not only Slack 🎉 
  - On other platforms, I render it as plaintext. Of course, MS Teams is like no others... so there I had to apply some hacks..
- [x] Replace the `/botkubehelp` with `@BotKube help`
  - **Reason**: The `Command not supported. Please use 'help' to see supported commands.` message was printed but `/botkubehelp` was not supported on Discord.
  - Now it's unified and all communicators support `@BotKube help` and we say to use `help` in case of errors.
- [x] Print a header (so we can easily correlate the output)
  - **Reason**: Because with buttons you can run command without typing, is hard to correlate the output with a given command.
    Screen Shot 2022-09-08 at 15 46 30

## Testing

To run e2e test locally:

1. Create cluster: `k3d cluster create`
2. Checkout PR: `gh pr checkout 727`
3. Install BotKube: 
    ```bash
    export SLACK_APP_TOKEN="xapp-1-"
    
    helm upgrade botkube --install --namespace botkube ./helm/botkube --wait --create-namespace \
        -f ./helm/botkube/e2e-test-values.yaml \
        --set communications.default-group.slack.token="${SLACK_BOT_TOKEN}" \
        --set communications.default-group.slack.botToken="${SLACK_BOT_TOKEN}" \
        --set communications.default-group.slack.appToken="${SLACK_APP_TOKEN}" \
        --set image.tag=727-PR \
        --set image.pullPolicy="Always" \
        --set analytics.disable=true
    ```
4. Run tests:
    ```bash
    export KUBECONFIG="config"
    export SLACK_BOT_NAME="botkube"
    export SLACK_TESTER_NAME="tester"
    export SLACK_ADDITIONAL_CONTEXT_MESSAGE="local"
    export SLACK_TESTER_APP_TOKEN="xoxb-"
    make test-integration
    ```

## Related issue(s)

#683


pkg/execute/executor.go
```diff
@@ -368,31 +345,12 @@ func (e *DefaultExecutor) findBotKubeVersion() (versions string) {
...
-		}
-		if strings.HasPrefix(arg, ClusterFlag.String()) {
-			if arg == ClusterFlag.String() {
-				checkFlag = true
-			} else if strings.SplitAfterN(arg, ClusterFlag.String()+"=", 2)[1] != clusterName {
-				return ""
-			}
-			continue
-		}
-	}
```
pkosiec, line 395: Okay, so this is a part I asked a question about in previous comment:

> Should we ignore the execution if the cluster name doesn't match? (---cluster-name is provided)? I think we should 🤔

cmd/botkube/main.go
```diff
@@ -394,18 +394,13 @@ func reportFatalErrFn(logger logrus.FieldLogger, reporter analytics.Reporter) fu
```
pkosiec, line 394: Remember about updating the docs afterwards - change configuration of all apps to get rid of the `/botkubehelp` command 👍

pkg/bot/interactive/help.go
```diff
@@ -47,9 +47,9 @@ func Help(platform config.CommPlatformIntegration, clusterName, botName string)
 					Description: fmt.Sprintf("You can run kubectl commands directly from %s!", cases.Title(language.English).String(string(platform))),
```
pkosiec, line 47: BTW The default configuration has `kubectl` disabled. Should we hide the section in such case? I think we should, but it will be easy to fix I think 🤔 I see at least two options:

1. Explicitly return to the user that kubectl functionality is disabled for this cluster. Don't show the commands, or have the buttons disabled.
2. At least add another sentence to the description that the kubectl should be enabled (if the first option would be too difficult).

First option is much better and should be easy to implement. Current behavior can be confusing - if you click the button, you'll get `Command not supported. Please use 'help' to see supported commands.` 😄

pkg/bot/discord.go
```diff
@@ -10,10 +10,10 @@ import (
 	"github.com/bwmarrin/discordgo"
```
pkosiec, line 10: Out of curiosity: Can we somehow disable link previews from the help message? 🤔 I don't expect it's possible with a simple message, but maybe there's is a simple way in `discordgo` 🤔 

Screenshot 2022-09-19 at 12 14 48

pkosiec (commented):
Tested Slack and Discord, works well 👍 LGTM, but I have a question about behavior for info command + suggestion to fix the behavior for `help` command when `kubectl` executor is disabled (default settings).

cmd/botkube/main.go
mszostok: Docs updated, PTAL: https://github.com/kubeshop/botkube-docs/pull/142

pkg/bot/discord.go
mszostok: Unfortunately, I don't see such option.

pkg/bot/interactive/help.go
mszostok: Let's address that in folllow-up PR and I will sync with @lpetkov about the approach for that.

1. "(..)  have the buttons disabled." - that's not possible, however we are able to don't show the commands and "(..) should be easy to implement." - yep, but requires a bit of new code and e2e testing.

pkg/execute/executor.go
mszostok: it's already handled in line 142

pkg/bot/interactive/msg.go
```diff
@@ -0,0 +1,98 @@
...
+)
+
+// mdEmojiTag finds the emoji tags
+var mdEmojiTag = regexp.MustCompile(`:(\w+):`)
+
+// MSTeamsLineFmt represents new line formatting for MS Teams.
+// Unfortunately, it's different from all others integrations.
+func MSTeamsLineFmt(msg string) string {
+	// e.g. `:rocket:` is not supported by MS Teams, so we need to replace it with actual emoji
+	msg = replaceEmoijTagsWithActualOne(msg)
```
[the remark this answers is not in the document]
mszostok: It's tricky, if you will use actual emoji then it's not converted to a built-in one (e.g. on Discord it doesn't look good). Additionally, slack normalized that and even if you sent and actual emoji in "response" you will have `:rocket:` which later need to be handled in test. So even if we will send the actual emoji we won't be able to get rid of that func as we will need to do the normalization in e2e tests then.

pkg/bot/interactive/help.go
pkosiec: A follow-up PR totally works for me 👍I just wanted to make sure we won't forget about this 👍

pkg/bot/interactive/msg_test.go
```diff
@@ -0,0 +1,37 @@
...
+)
+
+// go test -run=TestInteractiveMessageToMarkdown ./pkg/bot/interactive/... -test.update-golden
+func TestInteractiveMessageToMarkdown(t *testing.T) {
+	tests := []struct {
+		name    string
+		lineFmt func(msg string) string
+	}{
+		{
+			name:    "render with Markdown new lines",
```
pkosiec, line 17: Right? 😄 
```suggestion
			name:    "render with MS Teams new lines",
```

pkosiec (approved):

pkg/bot/interactive/msg_test.go
mszostok: homer-simpson-hide-in-shrubs

Issue #710, opened 2022-08-29
Notification Presets


brampling:
A BotKube user needs to be able to set notification settings using a predefined set of configuration settings so they can quickly and easily tune the notification verbosity in a BotKube channel.

## Design

https://www.figma.com/proto/iGGgUG9vn6C9WvDAFwKS9G/Botkube-Help---Slack?page-id=209%3A1337&node-id=269%3A1038&viewport=282%2C271%2C0.38&scaling=min-zoom&starting-point-node-id=269%3A1038

## Acceptance Criteria

- A user can select from a predefined set of notification verbosity settings (Slack)
- A user can set the verbosity using `@BotKube edit source bindings "source1, source2"`  (all platforms)
- The setting is channel-scoped
- We provide OOTB:
  - k8s-err-events - only error events
  - k8s-info-events - same as we had already
  - k8s-all-events - all possible events for all resources

PR #739, opened 2022-09-19
Define different k8s source presets, add option to set event list globally

mszostok:
## Description

- Add option to set events list globally
- Define different k8s source presets
  - `k8s-recommendation-events` - enables all possible k8s recommendations
  - `k8s-all-events` - is the name of the previous configuration
  - `k8s-err-events` - enables only `error` events for resources that were watched previously

## Related issue(s)

- #710

### DIFF
--- a/helm/botkube/values.yaml
+++ b/helm/botkube/values.yaml
@@ -448,6 +448,7 @@ communications:
         # -- Notification sources configuration for the webhook.
         sources:
           - k8s-err-events
+          - k8s-recommendation-events
 
 ## Global BotKube configuration.
 settings:
--- a/helm/botkube/values.yaml
+++ b/helm/botkube/values.yaml
@@ -58,8 +58,7 @@ kubeconfig:
 #
 ## Format: sources.<alias>
 sources:
-  'k8s-events':
-
+  'k8s-recommendation-events':
     # -- Describes Kubernetes source configuration.
     kubernetes:
       # -- Describes configuration for various recommendation insights.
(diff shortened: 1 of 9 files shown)


helm/botkube/values.yaml
```diff
@@ -446,7 +447,7 @@ communications:
       bindings:
         # -- Notification sources configuration for the webhook.
         sources:
-          - k8s-events
+          - k8s-err-events
```
mszostok, line 450:
```suggestion
          - k8s-err-events
          - k8s-recommendation-events
```

pkg/config/testdata/TestLoadConfigSuccess/config.golden.yaml
```diff
@@ -8,17 +8,15 @@ sources:
                 pod:
                     noLatestImageTag: false
                     labelsSet: true
+            events:
```
pkosiec, line 11: If now we have global `namespace.include`, should it be also visible in the golden file? I can't see it 🤔

pkosiec (approved):
LGTM - please apply your suggestion and also see my question 🙂 

Disclaimer: Didn't test it, as tests cover that functionality. Will test the whole notification presets functionality E2E

pkg/config/testdata/TestLoadConfigSuccess/config.golden.yaml
mszostok: There was a typo (missing `s`), should be `namespaces` instead. Fixed 👍

PR #770, opened 2022-09-27
Add analytics to slack interactive buttons

### LINKED ISSUE #756: Add telemetry to the Slack interactivity blocks
## Overview

We already have a few things that supports interactivity in Socket Slack:
1. Help message—posted automatically or via `@BotKube help`
2. Feedback command—included in Help message and via `@BotKube feedback`
3. Edit source bindings—included in Help message and via `@BotKube edit SourceBindings`

All of the above should be covered with dedicated telemetry.

## Acceptance Criteria

- Metric in segment.io is visible when user clicks the interactive buttons / modals
- We can distinguish what button was clicked

## Reason

Learn more how the interactivity in Slack is used by our users and decide on priorities in the next releases.

mszostok:
## Description

- Add analytics to slack interactive buttons. IMO that's the simplest implementation.

   I tried to not modify the code where the `ReportCommand` was used and instead pass func there (see: [bd5020d](https://github.com/kubeshop/botkube/pull/770/commits/bd5020da642d7f55afea53c07e918518ed7470b8#diff-6fe247c9584512f88190e551aeeb30dfedf84d5a4de4aa0a35a9ab1ec221b32aR116-R119)) but the problematic part is for instances such as notifier or edit manager where we already pass the `analyticsReporter`. In this case, to avoid reaces we would need to always do `NewEditExecutor`,`NewNotifierExecutor`, each time we construct executor. IMO this is not worth it. 

We don't need to update our privacy policy as we still only work in "command" area.

## Testing

To test that you need to build the image manually with embedded API key:
```bash
env ANALYTICS_API_KEY="" goreleaser release --rm-dist --snapshot --skip-publish
```

Later you can import that image into your cluster: `k3d image import ghcr.io/kubeshop/botkube:v9.99.9-dev` (if you use k3d).

Then access segment.io, and run `@BotKube help` and then click button to run some  command. You should see two events:
- help command with `"origin": "typed"`
- your command but with `"origin": "buttonClick"`

## Related issue(s)

Fix #756

### DIFF
--- a/internal/analytics/segment_reporter.go
+++ b/internal/analytics/segment_reporter.go
@@ -17,9 +17,13 @@ import (
 const (
 	kubeSystemNSName  = "kube-system"
 	unknownIdentityID = "00000000-0000-0000-0000-000000000000"
+)
 
-	typedCommand  = "typed"
-	buttonCommand = "buttonClick"
+type commandOrigin string
+
+const (
+	buttonClickCommandOrigin commandOrigin = "buttonClick"
+	typedCommandOrigin       commandOrigin = "typed"
 )
 
 var (
@@ -63,9 +67,9 @@ func (r *SegmentReporter) RegisterCurrentIdentity(ctx context.Context, k8sCli ku
 // ReportCommand reports a new executed command. The command should be anonymized before using this method.
 // The RegisterCurrentIdentity needs to be called first.
 func (r *SegmentReporter) ReportCommand(platform config.CommPlatformIntegration, command string, isInteractiveOrigin bool) error {
-	origin := typedCommand
+	origin := typedCommandOrigin
 	if isInteractiveOrigin {
-		origin = buttonCommand
+		origin = buttonClickCommandOrigin
 	}
 	return r.reportEvent("Command executed", map[string]interface{}{
 		"platform": platform,
--- a/internal/analytics/segment_reporter.go
+++ b/internal/analytics/segment_reporter.go
@@ -19,7 +19,7 @@ const (
 	unknownIdentityID = "00000000-0000-0000-0000-000000000000"
 
 	typedCommand  = "typed"
-	buttonCommand = "button"
+	buttonCommand = "buttonClick"
 )
 
 var (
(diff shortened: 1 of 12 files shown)


internal/analytics/segment_reporter.go
```diff
@@ -59,10 +62,15 @@ func (r *SegmentReporter) RegisterCurrentIdentity(ctx context.Context, k8sCli ku
...
-func (r *SegmentReporter) ReportCommand(platform config.CommPlatformIntegration, command string) error {
+func (r *SegmentReporter) ReportCommand(platform config.CommPlatformIntegration, command string, isInteractiveOrigin bool) error {
+	origin := typedCommand
+	if isInteractiveOrigin {
+		origin = buttonCommand
+	}
 	return r.reportEvent("Command executed", map[string]interface{}{
 		"platform": platform,
 		"command":  command,
+		"origin":   origin,
```
pkosiec, line 73: WDYT about `trigger`?

```suggestion
		"trigger":   trigger,
```

internal/analytics/segment_reporter.go
```diff
@@ -59,10 +62,15 @@ func (r *SegmentReporter) RegisterCurrentIdentity(ctx context.Context, k8sCli ku
 
 // ReportCommand reports a new executed command. The command should be anonymized before using this method.
 // The RegisterCurrentIdentity needs to be called first.
-func (r *SegmentReporter) ReportCommand(platform config.CommPlatformIntegration, command string) error {
+func (r *SegmentReporter) ReportCommand(platform config.CommPlatformIntegration, command string, isInteractiveOrigin bool) error {
+	origin := typedCommand
+	if isInteractiveOrigin {
+		origin = buttonCommand
+	}
```
pkosiec, line 69: This part of code tells me that we need an enum passed here instead of bool to have it extensible. I think we cannot assume interactive = button click. We can keep bool, but only if we want to have `interactive=true/false`. Will we have other triggers, like `selectOnChange` etc.? I can imagine such events 🤔


WDYT?

internal/analytics/segment_reporter.go
```diff
@@ -17,6 +17,9 @@ import (
 const (
 	kubeSystemNSName  = "kube-system"
 	unknownIdentityID = "00000000-0000-0000-0000-000000000000"
+
+	typedCommand  = "typed"
+	buttonCommand = "buttonClick"
```
pkosiec, line 22: Please create enum (even if you don't agree about passing enum to the ReportCommand function, in such case just make it unexported):

```suggestion
type CommandTrigger string

const (
	ButtonClickCommandTrigger  = "buttonClick"
	TypedMessageCommandTrigger = "typedMessage"
)
```

pkosiec (commented):
Generally it looks OK - A few open questions and suggestions. Will test it after the comments are resolved 🙂

internal/analytics/segment_reporter.go
mszostok: I thought about it but at the same time it's not a problem to do that later as it's not user-facing API. Using bool right now simplifies a lot, see the code of slack in another places. The zero value of bool works here as intended. I would apply that once we will have such situation. However, the event that is sent to the segment.io has more generic properly, so we won't break this contract externally. So that's the reason I don't want to use the `interactive` properly externally, but "internally" I wanted to make it simple.

internal/analytics/segment_reporter.go
mszostok: done, (unexported)

internal/analytics/segment_reporter.go
mszostok: Honestly, I like the origin even more 🤔 I discuss that also with Blair before implementing and he also was OK with this property name.

internal/analytics/segment_reporter.go
pkosiec: 👍 Alright, makes sense. I thought to do the internal work now, but yeah, let's postpone it until it's necessary 👍

pkosiec: BTW consider changing `isInteractiveOrigin` to `isButtonClickOrigin` or something like this 🤔

pkosiec:
As discussed, please add reporting of Feedback button click (or any other URL) 👍

internal/analytics/segment_reporter.go
mszostok: changed to `isButtonClickOrigin` 👍

pkosiec (approved):
🚀

PR #776, opened 2022-09-28
Adjust formatting for help message and notification presets

mszostok:
## Description

- Adjust formatting for help message and notification presets


pkg/bot/slack_renderer.go
```diff
@@ -58,7 +58,7 @@ func (b *SlackRenderer) RenderModal(msg interactive.Message) slack.ModalViewRequ
 	return slack.ModalViewRequest{
 		Type:          "modal",
 		Title:         b.plainTextBlock(title),
-		Submit:        b.plainTextBlock("Select"),
+		Submit:        b.plainTextBlock("Adjust"),
```
pkosiec, line 61: I would consider `Edit` word as it would be consistent with the command 🤔 Or `Apply`?

BTW 'Adjust notifications' could be also changed to `Edit notifications` to have it consistent with command

But I don't know 😄

pkosiec (approved):
LGTM ✅

pkg/bot/slack_renderer.go
lpetkov: Good call. Regarding the dialog button: I think Apply is the better verb here as it communicates finality, not intent like Edit and Adjust. Let's go with Apply.

However regarding the Help message button _Adjust notifications_ and the dialog title, I think you are literally _adjusting_ a couple of small things, you are not _editing_, as in editing a file or editing text (which is the case with the command line). So we end up having this small inconsistency that I'm personally fine with.

PR #783, opened 2022-10-03
Botkube extensibility design for sources and executors

### LINKED ISSUE #717: Event Manager Extensibility Design
This task is to create the design for [Event Manager](https://github.com/kubeshop/botkube/issues/649).

**AC**

- A design for the event manager extensibility interface has been created and approved by the team and community (as applicable)
- A proposed process for adding/removing extended notifiers has been documented and reviewed
- Among other things, include also:
    - requirements
    - executor contribution flow
    - Possible implementation options (with pros & cons)
- Goal: send notifications based on keptn events
    - Allow rendering for a given event specified by user (e.g. Go template?)
    - Kubernetes source can be extracted based on the proposal without missing functionalities
- Use keptn as an example integration
    - probably use incoming webhook with CloudEvents schema
- Ideally, the filters should be supported for multiple sources https://github.com/kubeshop/botkube/issues/657 

## Notes

- PoC out of scope - it will be created as a follow-up if team feels it should be done (for 1 or e.g. 2 selected options)
- Create it as a public proposal 
- Keep the ceremonies to the minimal
    - Public meeting is out of scope
    - Pull request will be public, anyone will be welcome to review and comment 

**UPDATE (13.09.2022):**
- As Testkube would like to use BotKube to send notifications, reach out to @exu and the team and ensure they are happy with the design.

huseyinbabal:
## Description

- Design documentation is added
- PoC project added in the design documentation.

## Related issue(s)
Fixes #717


mszostok (commented):
I agree with the usage of the HashiCorp plugin lib. It's also used by a well known projects such as Terraform, or Porter.

The things that I'm still missing:
  - How they are managed? Installing/Uninstalling/disable/enable/discover new ones, see install parameters/versioning/upgrades/ etc.
    - how plugin will be configured (e.g. Kubernetes was extracted to become a plugin, how it's initiated?)
      For example, consider an optino on UI where we list all available plugins and some of them are not yet installed so we see "enable" button.
  - Where the plugins are executed. 
    - On the same pod? on the same container, or separated container? it's a long leaving? 
  - How Botkube knows the metadata of the plugin?
      For example, Porter has index.json: https://github.com/getporter/packages/blob/main/plugins/index.json
      We would need to have at least information about the install params (JSON Schema?), name, icon, description.
  - What is the unification layer on event type? I saw your cloud event schema, please described how we process that so we can send events to our platform in the same shape. 

Basically, I would like to see how all this can be fit into the Botkube ecosystem, so after this proposal we can start the implementation (define epics/sub-tasks etc.)
See the section of consequences in
- Bindings proposal: https://github.com/kubeshop/botkube/blob/main/docs/proposal/2022-06-14-configuration-bindings.md#consequences
- or in Analytics investigation: https://github.com/kubeshop/botkube/blob/main/docs/investigation/analytics-collection.md#consequences

docs/proposal/2022-09-28-event-manager-extensibility.md
```diff
@@ -0,0 +1,194 @@
+# Event Manager Extensibility
+
+Created on `2022-09-28` by Huseyin BABAL ([@huseyinbabal](https://github.com/huseyinbabal)))
```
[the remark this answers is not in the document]
huseyinbabal: added

docs/proposal/2022-09-28-event-manager-extensibility.md
```diff
@@ -0,0 +1,194 @@
...
+  * [Golang and Extensibility](#golang-and-extensibility)
+  * [Hashicorp's Go Plugin](#hashicorps-go-plugin)
+  * [Source Plugin Structure](#source-plugin-structure)
+  * [Executor Plugin Structure](#executor-plugin-structure)
+  * [Cloud Events](#cloud-events)
+    + [BotKube Cloudevents Data Structure](#botkube-cloudevents-data-structure)
+
+<!-- tocstop -->
+## Motivation
+Currently, we have support for 5 integrations for BotKube and BotKube is designed to listen events from kubernetes
```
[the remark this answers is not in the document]
huseyinbabal: applied

docs/proposal/2022-09-28-event-manager-extensibility.md
```diff
@@ -0,0 +1,302 @@
...
+}
+```
+
+`id`: This is the unique identifier for the event, and this should be decided by end users. UUID is not a mandatory stuff, it can
+be also ARN for example `ec2:myinstance:12345`.
+
+`source`: This refers to plugin name. Once Botkube receives and event on `/events` endpoint, `source` field would be validated
+to see we already have a registered plugin for this event or not. 
+
+`specversion`: To have proper versioning for Botkube event format.
```
mszostok, line 138:
```suggestion
`specVersion`: To have proper versioning for Botkube event format.
```
I don't see a reason to not use the camel case here.

docs/proposal/2022-09-28-event-manager-extensibility.md
```diff
@@ -0,0 +1,302 @@
...
+
+`time`: Event time
```
mszostok, line 140: Please specify the time format.

docs/proposal/2022-09-28-event-manager-extensibility.md
```diff
@@ -0,0 +1,302 @@
...
+
+`data`: This is the event data that contains important information about source. what you see in `data` section is just example,
```
mszostok, line 142:
```suggestion
`data`: This is the event data that contains important information about source. What you see in `data` section is just example,
```

docs/proposal/2022-09-28-event-manager-extensibility.md
```diff
@@ -0,0 +1,302 @@
...
+we can also add some kind of validation in Botkube to better use received data. For example, if we force user to send key value pairs,
+we can easily apply templating/filtering to this event to full-fill our future stories about sending custom notifications to slack.
+Data structure is strongly opened to suggestion.
+
+As you can also understand, data field contains generic data which can vary for plugins. For internal usage, we can create a general event data structure, but
+I haven't included the final structure for that, I believe it can be generated from scratch during real implementation.
```
mszostok, line 148: Let's state that in the consequences as this will also impact the estimation

docs/proposal/2022-09-28-event-manager-extensibility.md
```diff
@@ -0,0 +1,302 @@
...
+`plugin.go` is for plugin definition which is a simple struct for Hashicorp's Go plugin system.
+
+### Plugins Folder
+In this folder, you can see examples `kubectl` and `Kubernetes`. Each plugin has its own dedicated folders which are typically go module projects. 
+Each plugin folder contains a simple Go file which contains the actual business logic of plugin. You can also see a `Makefile` to manage the build
+process of that plugin. They might not be a simple Go program, that's why there is no unified `Makefile` that builds any Go project, it can contain
+customer specific build flows. For `kubectl` and `Kubernetes`, they are same for now. 
+
+### How to build plugins?
+`plugins` folder only contains the source code of the plugins, and consumer (Botkube or playground project on our case), needs the executable versions
```
mszostok, line 180: This means that we will have multiple releases on our repository? We will do in the same way release for the latest binary? how this will be build and used on PR's CI for e2e tests?

docs/proposal/2022-09-28-event-manager-extensibility.md
```diff
@@ -0,0 +1,302 @@
...
+    AllowedProtocols: []plugin.Protocol{plugin.ProtocolGRPC},
+    HandshakeConfig: plugin.HandshakeConfig{
+        ProtocolVersion:  1,
+        MagicCookieKey:   "BOTKUBE_MAGIC_COOKIE",
+        MagicCookieValue: "BOTKUBE_BASIC_PLUGIN",
+    },
+})
+```
+
+`type` can be `source` or `executor` since we decide on plugin implementation contract based on that value.
```
mszostok, line 221: How this `type` correlates with the above snippet?

docs/proposal/2022-09-28-event-manager-extensibility.md
```diff
@@ -0,0 +1,302 @@
...
+    "name": "kubectl",
+    "type": "executor",
+    "description": "Kubectl executor plugin to handle given command in configured k8s cluster.",
+    "version": "v1.0.9"
+  }
+]
+```
+
+### How can I provide plugin configuration?
+If you use terraform before, they use a notation like `<pluginname>_...` to pass settings to specific plugin. So, in Botkube, if somebody provides `KUBERNETES_CONFIG_PATH=something`, the `kubernetes` plugin executable can be run as follows.
```
mszostok, line 206: probably will be good to also require a prefix `EXECUTOR_` and `SOURCE_` to have dedicated namespaces in case of plugin name conflict.

docs/proposal/2022-09-28-event-manager-extensibility.md
```diff
@@ -0,0 +1,302 @@
...
+```go
+client := plugin.NewClient(&plugin.ClientConfig{
+    Plugins:          m.pluginMap(metadata),
+    VersionedPlugins: nil,
+    Cmd:              exec.Command(metadata.Path, "--config", "something"), // Plugin specific params goes here
```
mszostok, line 211: IMO we should just add an option to pass a raw config via `ConfigYAML string` property. In this way, each plugin can easily unmarshal it to a given struct. Providing such configuration: https://github.com/kubeshop/botkube/blob/ef6d42a8da292503664e744fb0224dec1d83ce19/helm/botkube/values.yaml#L61-L170
via env variables will too complicated.

docs/proposal/2022-09-28-event-manager-extensibility.md
```diff
@@ -0,0 +1,302 @@
...
+        MagicCookieKey:   "BOTKUBE_MAGIC_COOKIE",
+        MagicCookieValue: "BOTKUBE_BASIC_PLUGIN",
+    },
+})
+```
+
+`type` can be `source` or `executor` since we decide on plugin implementation contract based on that value.
+
+### How to install/uninstall plugins?
+Once you initialize plugin, it collects plugin metadatas from plugin index and initializes all of them by using Hashicorp's plugin system.
```
mszostok, line 224: So in the first implementation, each change will simply restart the Botkube as we do right now?

docs/proposal/2022-09-28-event-manager-extensibility.md
```diff
@@ -0,0 +1,302 @@
...
+helps us to communicate with this sub process. In Kubernetes environment, it will be in same pod. No need to worry about size, since only the required plugins will be donwloaded
+to local cache (user's home folder/.Botkube-plugins).
+
+### Example
+You can see a working example [here](https://github.com/huseyinbabal/Botkube-plugins-playground)
+
+### Consequences
+If we agree on this design, we can extend this epic with following description to come up with new tasks.
+
+- [ ] Maintain Botkube plugins under Kubeshop organization
```
mszostok, line 243: Is there an option to just maintain them under the existing `kubeshop/botkube` repo? and once we will have a lot of them we can think of some dedicated repo.

docs/proposal/2022-09-28-event-manager-extensibility.md
```diff
@@ -0,0 +1,302 @@
...
+  - Currently, Kubernetes source events are tightly coupled with Botkube, we can add sourcing mechanism as `kubernetes` plugin. You can see independent example [here](https://github.com/huseyinbabal/botkube-plugins/tree/main/plugins/kubernetes)
+  - It should be background compatible.
+  - Plugin specific parameters should be passed during plugin initialization. Hashicorp's Go plugin system, runs plugin executable as sub process, so it also accepts exec arguments where we can resolve configuration and pass them as separate arguments
+  - Plugin parameters will be documented in [botkube-plugins](https://github.com/huseyinbabal/botkube-plugins) repo, and they can be provided from Botkube as an environment variable like `KUBERNETES_...` so that they can be passed to executable [in this line](https://github.com/huseyinbabal/botkube-plugins-playground/blob/c85cb7b84296a2f41c2dcdd8cac77c0e9dd9c69a/plugin/manager.go#L159)
+  - Data structure for events
+    - Since each source plugin has different data structure, we cannot have a general data structure except we specified #Cloud Events. `data` field has free data structure, but we can convert them to map then extract
```
mszostok, line 265: but the questions is whether we really need to accept a free form? We can enforce a layout of the `data` but not necessary the naming. Sth like:

```yaml
data:
  fields:
     {field_name}: {field_value}
```

and tell them that how they will be processed by the Botkube, so they know how to shape them to achieve a given result. This can also be done as a part of the first implementation to see how this comes together.

docs/proposal/2022-09-28-event-manager-extensibility.md
```diff
@@ -0,0 +1,302 @@
...
+      "data" : {                                         
+        "namespace" : "Botkube",
+          "deployment" : {
+            "name" : "test",
+            "replicas" : 5
+          }
+        }
+      }
+      ```
+- [ ] Extract Kubectl executor feature as a plugin
```
mszostok, line 284: Will we give them a proper rights? How you see extracting the `kubectl` to a dedicated plugin when it comes to the `commands.verb` and `commands.resources` configuration?
Will be good to establish some common agreement about short and long term solution before jumping into implementation.

docs/proposal/2022-09-28-event-manager-extensibility.md
huseyinbabal: Not necessarily, there is an api in Hashicorp's Go plugin, I haven't fully tested that, the process would be update configuration and call Hashicorp plugin API to remove it.

docs/proposal/2022-09-28-event-manager-extensibility.md
huseyinbabal: Sure, during plugin initialization, we can pass it with the config file.

docs/proposal/2022-09-28-event-manager-extensibility.md
huseyinbabal: I just wanted to follow conventions as you can see an example [here](https://github.com/keptn/spec/blob/master/cloudevents.md), but we can try using camelcase and see if cloudevent SDK works.

docs/proposal/2022-09-28-event-manager-extensibility.md
huseyinbabal: I am not sure if I understand correctly, but when we extract to plugin, there will be no difference between current format and pluggable format. Configurations are blindly passed by plugin manager, but the actual validation and process should be handle by plugin. If conf is incorrect, then once Botkube calls this plugin, it will receive error response and we can handle exceptions.

mszostok: This is about RBAC, so Botkube currently has a cluster-admin role and kubectl executor simpily uses that but on the top of that we perform some validation to ensure that it's aligned with the permission defined in the configuration

docs/proposal/2022-09-28-event-manager-extensibility.md
huseyinbabal: Stated in the consequences

docs/proposal/2022-09-28-event-manager-extensibility.md
huseyinbabal: Ok, you are right, we cannot use multiple service accounts in one pod that means we cannot use different service accounts for kubectl and helm. However, I cannot see any concern about that in the beginning since;
- Think about your workstation machine that you use kubectl. Once you install helm, you don't really try to think which service account you should use. So you have unified service account and some roles are attached to it, you use it for both helm and kubectl
- Botkube is like a workstation machine where you virtually jump and execute command there. We can think same for Botkube, and customer designs a unified service account with respective roles then attach it to Botkube deployment.
- If we got a feedback about this, then we might think to come up pod-per-plugin notation, which would have operational cost at the beginning.
wdyt?

huseyinbabal:
> 

Thanks, I addressed them all.

ezodude:
Great deep dive. It feel you've drilled into a lot of detail.

My 2 pence => should we make use of the CloudEvents [type field](https://github.com/cloudevents/spec/blob/v1.0/spec.md#type)?

In case of `sources`/`executors`, this can help us understand the intent.

BotKube can catalogue, as part of the plugin docs, what available intents are available, e.g. 
- for a source plugin, this can be an example event type: `io.botkube.resource.created`
- for a executor plugin, this can be an example event type: `io.botkube.command.execute`

huseyinbabal:
> 

I am totally fine with that

huseyinbabal:
> Great deep dive. You've drilled into a lot of detail.
> 
> My 2 pence => should we make use of the CloudEvents [type field](https://github.com/cloudevents/spec/blob/v1.0/spec.md#type)?
> 
> In case of `sources`/`executors`, this can help us understand the intent.
> 
> BotKube can catalogue, as part of the plugin docs, what available intents are available, e.g.
> 
> * for a source plugin, this can be an example event type: `io.botkube.resource.created`
> * for a executor plugin, this can be an example event type: `io.botkube.command.execute`

I am totally fine with that

docs/proposal/2022-09-28-event-manager-extensibility.md
mszostok: Here is my 2 cents:
1. Having an option to consume the plugin from an external repo is a good thing that should be supported, so we enable external (community maintained) plugins
2. Having core plugins such as kubectl, k8s and later helm, argo etc. can still stay close to the code.

Why I have issue with separate repository:
1. We will still have logic in both places in `botkube` and `botkube-plugins` repo that at the beginning will need to be synced. For example, current permission management, interactive dialogs, and similar.
2. I see a huge problem and slow down when it comes to development and testing. If we have two repos, then I need to play with both of them to do the release of plugin (put it somewhere too?) and to run botkube to consume it.
3. How we will eventually test that on CI? When a plugin is changed, will it trigger it with latest Botkube? When Botkube is changed, will it see the latest release of all plugins? What if feature must be done in both repos, how it will be tested?

**What I propose:**
1. Have an option to consume external plugins as described, but for now do it only for 3rd party plugins.
2. Core plugins still leave in dedicated folders under main `botkube` repository.

**Workflow:**
1. The plugins are built only on official Botkube release and added to the release assets.
2. On official release, the Botkube docker image doesn't have binaries included. Instead, it fetched them from a given release as described in the proposal.
3. On CI the plugins are built on change and incorporated in the Botkube image — no need to sync state between 2 repositories.
  - alternatively we can even build plugins into a single Docker and mount it as init container and copy the binaries into mounted volume (emptyDir). IMO no need to do that.
4. On local testing the plugins are built on change and incorporated in the Botkube image - no need to run it in 2 repos, no need to push it somewhere or mock download for local testing.

**Consequences:**
1. The core plugins are in the `botkube` repository (at least for the beginning)
3. Botkube can load plugins from file system without downloading bins from internet.

mszostok:
The additional thing that is not covered in this proposal is how we will manged the interactive messages.
IMO this can be done in separate PR.

TL;DR; 
IMO we should just add support to 3 primitives:
- plain text input
- dropdowns
- buttons

This can be achieved but including the `Message` struct in gRPC API. So executor can replay with such definition:
https://github.com/kubeshop/botkube/blob/6fa06508396eccce0766c4936fae9684e998f96d/pkg/bot/interactive/message.go#L36-L37

mszostok (approved):
All LGTM after f2f discussion and alignments! 🚀

huseyinbabal:
@mszostok I moved external repo related stuff and release versioning to main repo, and updated doc accordingly. Will update my playground projects afterwards.

docs/proposal/2022-09-28-event-manager-extensibility.md
pkosiec: I think it would be great to mention it in the doc, just to make sure we remember about it 👍

pkosiec (approved):
Looks great, awesome work! 🚀 After today's sync we're on the same page, so ✅

docs/proposal/2022-09-28-event-manager-extensibility.md
huseyinbabal: updated accordingly

docs/proposal/2022-09-28-event-manager-extensibility.md
huseyinbabal: AS we discussed, I am ok with that and updated doc accordingly.

docs/proposal/2022-09-28-event-manager-extensibility.md
huseyinbabal: As we discussed in the meeting, we can use flat data structure for now, we will be testing interactive message data structure and if it works smoothly, we can use that format.

Issue #786, opened 2022-10-03
Make event notifications actionable 


lpetkov:
Add a drop down to each Botkube notification message with a list of relevant actions that:

- [x] are supported by Botkube
- [x] are applicable to the resource type
- [x] the bot has permissions to actually run
- [x] Always return enabled verbs for a given channel regardless the resource (this is what @mszostok already did)
    - Simplified implementation, you can hardcode some verbs for specific resources
    - Populate resources dynamically and attach describe if they have `get`
    - `logs` should be handled only for Pods
    - Use`kubectl api-resources -owide` for verbs (use client-go) if that's good idea
    - Create it as a generic package - sync up with #748 owner, as it will be used there as well
    - Can be worked on in parallel
- [x] Documentation
- [x] Basic telemetry: add `eventDropdown` origin for the executed command
- [x] Integration testing out of scope (as we use legacy Slack)

### Update

- [x] ~Support filtering from the flow 2 - https://www.figma.com/proto/UcaGQ1UiBziOQbNPmSwE0T/grep?page-id=0%3A1&node-id=56%3A443&viewport=163%2C220%2C0.08&scaling=scale-down&starting-point-node-id=56%3A443&show-proto-sidebar=1 - reuse the same modal from #788~ - dropped from the scope
- [x] Add delete confirmation - will be followed-up in https://github.com/kubeshop/botkube/issues/824; for now we'll hide the option
    I'll work on a follow-up PR with a special case for delete. Any kind of minimal extra confirmation works:
    - modal dialog
    - render confirmation button
    - render button right next to it
    
    If that'll be too big effort - filter the "delete" option ouf of this dropdown.

Example:

<img width="703" alt="image" src="https://user-images.githubusercontent.com/31132490/193618055-9011f47b-fcf1-4736-b917-0b76eaf6b481.png">

This will make each notification (or most notifications) _actionable_ and therefore will make the bot easier to use.

## Out of scope

**Stretch**
If `kubectl` is not enabled, whenever we have the technical ability to do so, add a button to enable kubectl under each notification:

![image](https://user-images.githubusercontent.com/31132490/193620964-5f837777-b494-4048-bb19-84d8e1e44661.png)

mszostok:
Filtering should not be via dropdown but inlined:
![image](https://user-images.githubusercontent.com/17568639/194042172-51424dd5-eaab-4431-b239-55893a3122b0.png)

pkosiec:
[...]

Issue #748, opened 2022-09-21
Slack Common Commands


brampling:

lpetkov:
Additional acceptance criteria comments:

- In the Resource list step, let's dynamically narrow down the list of resources based on the user configuration so that they only see relevant items.
- We can allow the user to customise this page at a later point, not necessary for the time being.
- The list of resources in the prototype is not definitive, we can discuss further what other resources to include (or exclude). 
- For a start we can try to render returned _lists of things (e.g. pods)_ in drop-downs and see where it makes sense to render lists as radio button groups. 
- We can define all the kubectl commands we want to support initially and go from there to design a more generic solution if that makes sense. 

If the effort to re-create the whole `@Botkube kcc` experience as designed in the prototype is deemed too big for one release, we can first focus on making the current BotKube alerts actionable with interactive buttons and build on that for 0.16 to get to what we have in the prototype.

@brampling let us know if you have any disagreements.

lpetkov:
One more thing I added to the prototype as a requirement is to display _who_ in the channel made the bot run a command: 
<img width="506" alt="image" src="https://user-images.githubusercontent.com/31132490/192084061-530710c4-58af-442a-a27d-ba41c1341fa3.png">

mszostok:
## TODO in follow-up PRs
- ~associate verbs with resources~ will be done in https://github.com/kubeshop/botkube/issues/786
   - don't display resource types drop-down for verbs not associated with nothing e.g. `api-versions`, `cluster-info`, etc.
   - don't display resource names drop-down for verbs not associated with nothing e.g. `api-versions`, `cluster-info`, etc.
- split resource types into categories
  ![Screen Shot 2022-10-04 at 11 46 31](https://user-images.githubusercontent.com/17568639/193788973-fb17c293-4419-4e58-be29-c9e598609391.png)

- solve bug when resource name is selected and later the verb/resourceType is changed
- add the filter output input string
   ![Screen Shot 2022-10-04 at 11 44 39](https://user-images.githubusercontent.com/17568639/193788466-c2428d5e-5b0b-4269-a6e7-672475b3bbf6.png)
- print the data in the dropdown 
  - print the plaintext
     ![Screen Shot 2022-10-04 at 11 44 16](https://user-images.githubusercontent.com/17568639/193788399-940b9da2-a26e-4182-8d31-21f29c0428f0.png)

- discuss the flags approach and introduce the final design. For example, support only the `namespaces` property (and `output`).
[...]

PR #787, opened 2022-10-03
Introduce interactive `kubectl` command selection

mszostok:
## Description

- Introduce interactive `kubectl` command selection

This PR aims to add a basic support for interactive mode via Slack. I tried to enable a parallel work on rest of the items.

## Supported cases

- send it as only visible to you
- replace the original message when changing verbs/resources via dropdown
- add select dropdown support both in slack renderer and static message
- add dropdown with verbs and resources supported in a given channel (alphabetically sorted)
- fix bug with wrongly parsed args in `kubectl` executor
- populate the resource names dynamically
- add info who executed the command
  Screen Shot 2022-10-04 at 10 13 09

## In follow-ups
1. Add support for the `namespace` dropdown
2. Use a dedicated service to return resources for a given verb (when implemented in https://github.com/kubeshop/botkube/issues/786). Currently, you can "defined" not valid command e.g.  `kubectl auth deployment`.
3. Add the filter output input string field
4. Add telemetry
5. Print the dropdowns  when user types `@Botkube k|kc|kubectl` without arguments.

## Related issue(s)

- #748


pkg/bot/interactive/help.go
```diff
@@ -63,9 +63,9 @@ func Help(platform config.CommPlatformIntegration, clusterName, botName string)
 					Description: fmt.Sprintf("You can run kubectl commands directly from %s!", cases.Title(language.English).String(string(platform))),
 				},
 				Buttons: []Button{
-					btnBuilder.ForCommandWithDescCmd("Run command", "get services"),
```
huseyinbabal, line 66: what about extracting `Run command` to constant to prevent duplication?

pkg/bot/interactive/message.go
```diff
@@ -28,15 +28,30 @@ const (
 type Message struct {
 	Type MessageType
 	Base
-	Sections          []Section
+
+	Sections []Section
+
```
huseyinbabal, line 33: What is the logic behind having empty lines for some fields, bot not others?

mszostok: yep, as previously I had 
```go
// Message represents a generic message with interactive buttons.
type Message struct {
	Type MessageType
	Base

	Sections []Section
	Actions   []Actions

	OnlyVisibleForYou bool
	ReplaceOriginal   bool
}
```

But I simplified the struct, but we can remove those new lines, I don't have strong opinion here 👍

huseyinbabal:
First, it works well 🎖️ 🚀  And here are my findings which we can fix before release;

# 1. `Resources` dropdown is not removed once I select verb which does not require Resource param. To reproduce
- Verb: `get`, Resource: `configmaps`
- Verb: `api-resources` -> see `configmaps` Resource is not removed. Other examples: api-versions, cluster-info
- Click run -> you will see execution error
Screen Shot 2022-10-06 at 11 39 41
Screen Shot 2022-10-06 at 11 39 49

# 2. Since `Run command` button is not sticked, it disappears from the view after a couple of commands. To reproduce;
- Verb: `get`, Resource: `namespaces`
- Click `Run Command` 2-3 times -> see command is scrolled up and up until it disappears
 
**First Click**

Screen Shot 2022-10-06 at 11 46 49

**Second Click**

Screen Shot 2022-10-06 at 11 47 22

**Third Click**

Screen Shot 2022-10-06 at 11 47 39

# 3. There is no way to describe single resource. To reproduce;
- Verb: `get`, Resource: `namespaces` -> see namespaces are populated in Resource name dropdown
- Verb: `describe` -> See Resource name dropdown is gone and only `namespaces` Resource dropdown left
Better to keep Resource name dropdown there so that we can describe single resource.
Screen Shot 2022-10-06 at 11 51 13
Screen Shot 2022-10-06 at 11 51 29

huseyinbabal (approved):
🚀

PR #795, opened 2022-10-06
Add namespace dropdown, render other dropdowns dynamically based on context

mszostok:
## Description

- Add namespace dropdown
- Render other dropdowns dynamically based on context
- Remove "diff" and  "auth" from default kubectl verbs. They are not usable in our context.
- Add fake command guard that will be implemented in https://github.com/kubeshop/botkube/issues/786. I added it to showcase that different edge cases are supported.

## Related issue(s)

#748 

## @huseyinbabal's review status
- [x] UI test is done, see my findings [here](https://github.com/kubeshop/botkube/pull/795#issuecomment-1271291553)
- [x] Code review


huseyinbabal:
@mszostok this time, I wanted to test first to understand the flows, then will do code review :) Here are my findings.

# 1. Coloured outputs corrupts message on Slack (This is normal, maybe we can escape them.) 

To reproduce;
- Command: `cluster-info` -> See it will print color code as unicode characters as shown below.

**In Terminal**

Screen Shot 2022-10-07 at 11 16 49

**In Slack** 

Screen Shot 2022-10-07 at 11 17 02

# 2. Namespace selection is not clear enough for some cases. 

To reproduce;
- Command: `describe`, Resource: `services`  -> after this selection, it will show 2 more dropdowns. Resource name is clear, but namespace selector shows `defatul` and it is not easy to understand it is namespace selector.

Step1
Screen Shot 2022-10-07 at 11 21 06

Step2 
Screen Shot 2022-10-07 at 11 21 14

# 3. It is not possible to explain after a get or describe operation.

To reproduce;
- Select Command: `get`, Resource: `services`, Resource name: `kubernetes` -> You don't need to click Run command here
- Select Command: `explain`, Resource: `configmaps` -> see remaining dropdows are not reseted. 
- Click `Run command`

Screen Shot 2022-10-07 at 11 28 43

# 4. Confusing reshuffle on logs command

To reproduce;
- Reset all dropdowns
- Select Command: `logs`, Resource: `pods` -> See namespace selector appeared as 3 item in first row
- Select `botkube` namespace -> See namespace dropdown replaced by resource name and namespace dropdown moved to next line

Step1
Screen Shot 2022-10-07 at 11 30 21

Step2
Screen Shot 2022-10-07 at 11 30 32

pkg/execute/fake_kc_guard.go
```diff
@@ -0,0 +1,71 @@
...
+
+	var out []Resource
+	for _, name := range allConfiguredResources {
+		out = append(out, staticResourceMapping[name])
+	}
+	return out, nil
+}
+
+// GetResourceDetails returns resource details.
+func (f *FakeCommandGuard) GetResourceDetails(verb, resourceType string) Resource {
```
huseyinbabal, line 39: I see one problem here, but please ignore if @pkosiec will handle that with non-fake mapping.
- In `staticResourceMapping`, there are missing resources like `pv`, `pvc`, etc.
- Let say that, I override helm values yaml for `commands.resources` and add `pvc`
- Then I will get a nil pointer on the places where this function is used like [here](https://github.com/mszostok/botkube/blob/330c1073619368d3d44985abfe0756d39b32af54/pkg/execute/kubectl_survey.go#L212)

mszostok:
Hi, thanks for great review!

### 1. Coloured outputs corrupts message on Slack (This is normal, maybe we can escape them.)

I will try to use the `github.com/gookit/color` and their `color.ClearCode`, I used it in my other project and it should work properly.

### 2. Namespace selection is not clear enough for some cases.

The namespace selector is always "selected" as there is always a "default" namespace in which we try to find resources. The default namespace is either the one from config:
https://github.com/kubeshop/botkube/blob/main/helm/botkube/values.yaml#L264
or if not found the `default` name.

I get your point, but what you propose here? 

### 3. It is not possible to explain after a get or describe operation.

This is because I use a dummy version of command guard that will be implemented in Make bot messages actionable  #786. 

### 4. Confusing reshuffle on logs command

So the reason for that is that we cannot print disabled dropdowns. If there are no items we simply don't display a given dropdown.

The dropdowns are always printed in a given order:

<verb> <resource type> <resource name> <resource namespace>

That's the reason why you see this happening. I will talk to Lacho about that, but maybe you also have some proposition?

pkg/execute/fake_kc_guard.go
mszostok: yes, this is a dummy implementation as stated in the comment on top. I did it only to showcase how dropdowns are added/deleted based on selected options. I didn't want to spend too much time on it, so as a result it has such problems, this service is even not unit-tested. I expect that it will be replaced by Paweł's implementation at the beginning of the next week.

huseyinbabal:
> Hi, thanks for great review!
> 
> ### 1. Coloured outputs corrupts message on Slack (This is normal, maybe we can escape them.)
> I will try to use the `github.com/gookit/color` and their `color.ClearCode`, I used it in my other project and it should work properly.
> 
> ### 2. Namespace selection is not clear enough for some cases.
> The namespace selector is always "selected" as there is always a "default" namespace in which we try to find resources. The default namespace is either the one from config: https://github.com/kubeshop/botkube/blob/main/helm/botkube/values.yaml#L264 or if not found the `default` name.
> 
> I get your point, but what you propose here?
> 
> ### 3. It is not possible to explain after a get or describe operation.
> This is because I use a dummy version of command guard that will be implemented in Make bot messages actionable #786.
> 
> ### 4. Confusing reshuffle on logs command
> So the reason for that is that we cannot print disabled dropdowns. If there are no items we simply don't display a given dropdown.
> 
> The dropdowns are always printed in a given order:
> 
> That's the reason why you see this happening. I will talk to Lacho about that, but maybe you also have some proposition?

1 and 3 seems reasonable.

For 2, If there is a way to select default as you said and show `Namespaces` placeholder at the beginning, that would be nice. Here is a pure HTML example 👉 [https://jsfiddle.net/72tkh01g/show](https://jsfiddle.net/72tkh01g/show) If this will impact your current design, then you can totally skip I think, we can react based on end user feedback.

For 4, I have only one idea which is, since you already have a default namespace, maybe once I select pod, you can show both resource name and namespace dropdown, and put resources from selected namespace in resource name dropdown

huseyinbabal (approved):
There are no blocker items, only minor nice-to-have ones, well done for this great feature 🚀

mszostok:
In commit: https://github.com/kubeshop/botkube/pull/795/commits/da9b50a8fa2da73844d6bb1d932d79798843ac9e
I added:
- force initial dropdown option to fix Slack app on iOS
- remove the colors from kubectl output
- always render the resource name dropdown even if empty (no resources found in a given namespace). This solves the problem with "reshuffling" dropdowns.

PR #800, opened 2022-10-07
Support quotes for the `--cluster-name` flag

### LINKED ISSUE #799: Using --cluster-name stops any responses
## Description
```
BotKube version: v0.14.0
K8s Server Version: v1.23.10-eks-15b7512
Slack
```

We have multiple clusters running botkube, as as advised by the bot, we try to do commands in this manner:
`@botkube get pods --cluster-name=“myclustername”`
This results in no response.

If I do `@botkube get pods`, all clusters respond with the correct answer.

## Expected behavior

I should have a working mechanism to choose a call/response to a single botkube instance.

pkosiec:
## Description

- Support quotes for the `--cluster-name` flag
- Suggest in help providing cluster name without quotes (as they might be problematic in some communication platforms - e.g. replaced with different characters)

## Screenshots

Screenshot 2022-10-07 at 14 05 10

image

## Related issue(s)

Resolves #799


pkg/utils/utils.go
```diff
@@ -122,7 +123,13 @@ func GetClusterNameFromKubectlCmd(cmd string) string {
 	if len(matchedArray) >= 2 {
 		s = matchedArray[1]
 	}
-	return s
+
+	str, err := strconv.Unquote(s)
+	if err != nil {
```
ezodude, line 128: Had to think about this, works.

ezodude (approved):
👍

pkg/utils/utils.go
pkosiec: if you think I should put a comment or something, or you have better alternative, let me know 👍

ezodude: Nope all good.

PR #801, opened 2022-10-07
Print kubectl command builder on `@Botkube k|kc|kubectl`, add tokens validation

mszostok:
## Description

- Update dropdown placeholders to make them actionable
- Add telemetry
- Print the dropdowns when user types `@Botkube k|kc|kubectl` without arguments.
- Return dedicated message when kubectl is not enabled or no verbs are available in a given channel
Screen Shot 2022-10-10 at 13 51 07

- Add validation for SocketSlack and Slack tokens:
    ```
    2022/10/10 13:32:18 while loading app configuration: found critical validation errors: 3 errors occurred:
            * Key: 'Config.Communications[default-group].SocketSlack.BotToken' BotToken is a required field
            * Key: 'Config.Communications[default-group].SocketSlack.BotToken' BotToken must have the xoxb- prefix. Learn more at https://botkube.io/docs/installation/socketslack/#obtain-bot-token
            * Key: 'Config.Communications[default-group].SocketSlack.AppToken' AppToken must have the xapp- prefix. Learn more at https://botkube.io/docs/installation/socketslack/#generate-and-obtain-app-level-token
    ```

- Rename the kubectl survey into kubectl command builder.

### Enabled kubectl

1. Install Botkube from this PR
2. Start the kubectl command builder via:
   - `@Botkube k`
   - `@Botkube kc`
   - `@Botkube kubectl`

### Disabled kubectl

1. Install Botkube from this PR
2. Start the kubectl command builder via:
   - `@Botkube k`
   - `@Botkube kc`
   - `@Botkube kubectl`

## TODO

-  [x] Testing telemetry - done by @pkosiec 

## Related issue(s)

#748


mszostok:
Tests are passing locally, probably it's problem with our CI. I will discuss that offline.

pkg/execute/kubectl_cmd_builder.go
```diff
@@ -12,19 +12,23 @@ import (
 	"github.com/slack-go/slack"
```
pkosiec, line 12: We could hide `auth` command as it doesn't seem to be useful without flags:
Screenshot 2022-10-12 at 09 32 58

WDYT?

pkg/execute/executor.go
```diff
@@ -188,9 +181,20 @@ func (e *DefaultExecutor) Execute() interactive.Message {
 		return empty
 	}
 
+	if e.kubectlCmdBuilder.CanHandle(args) {
```
pkosiec, line 184: Couldn't test it fully now, but I think in current implementation of command reporting (analytics), we can't distinguish `kubectl get services/pods/deployments` clicked via Help vs `kubectl get services/pods/deployments` executed with the interactive kubectl.
I think it's time to introduce the enum now - or anything what makes us know that the origin is "interactive kubectl". For me, it's a really nice experimental feature, so we'll know if it is used or not.

pkosiec (commented):
Two small comments 🙂 I tried to test it, but I was struggling with the following error:

https://user-images.githubusercontent.com/7155799/195279306-ee630f86-174c-44a8-a680-b455347fa0a9.mov

I ran it locally outside cluster with the following executor config:
```yaml
executors:
  'kubectl-backend':
    kubectl:
      namespaces:
        include:
          - ".*"
      enabled: true
      commands:
        verbs: [ "run", "api-resources", "api-versions", "cluster-info", "describe", "diff", "explain", "get", "logs", "top", "auth", "delete" ]
        resources: [ "deployments", "pods", "namespaces", "daemonsets", "ingresses", "statefulsets", "storageclasses", "nodes", "configmaps", "services" ]
      defaultNamespace: backend
      restrictAccess: false
```

@mszostok do you know what could be the reason? Can you also reproduce it? We can sync f2f.

EDIT: Same on main 🤔

pkg/execute/kubectl_cmd_builder.go
mszostok: the `auth` was already removed from the default configuration. I also removed the `diff` as it also doesn't make sens. 
Not sure why you mentioned flags though 🤔  you can run it like `kubectl auth can-i get pods`. However, this is sth which is not true, as it's not going through our "policy engine" and as Botkube has cluster admin, it can do everything but we still have restrictions per channel.



As you can see, I introduced a Guard which I called dummy. I thought that the new Guard will handle such cases. If not, I can create a wrapper around the real Guard.

pkg/execute/executor.go
mszostok: tbh I don't see an option to do that. Let me know how you would like to implement that. For now we will have analytics that someone was using the command builder as we will report the `kc-cmd-builder`. I will also try to think about some option to distinguish those buttons clicks.

pkg/execute/kubectl_cmd_builder.go
pkosiec: I meant hiding the option to execute `auth` from interactive kubectl, as this result in the response I posted.
Screenshot 2022-10-12 at 10 23 04

We agreed Guard has two methods, and neither of them returns a list of verbs, both of them work with a selected verb:
```go
GetAllowedResourcesForVerb(verb string, allConfiguredResources []string) ([]Resource, error)
GetResourceDetails(selectedVerb, resourceType string) (Resource, error)
```

So displaying verbs is something before using guard. But we can add another method for it, which will simply filter not  supported commands by the interactive kubectl. I can do that 👍 Just wanted to hear your thoughts about hiding this option even if it is configured for executors.

pkg/execute/executor.go
pkosiec: > For now we will have analytics that someone was using the command builder as we will report the kc-cmd-builder.

Oh, I didn't see that in the Segment.io Debugger console, but you know how the console works, so... If this is reported as well, then it's fine for me 👍

mszostok:
@pkosiec The problem that you noticed was related to the dummy guard implementation. I updated that but it will be fully replaced by the proper implementation on some follow-up PR once your PR will be merged 👍 But anyway I added a few more "guards" in current code to avoid such situation in case of bug in the final guard impl👍

pkg/execute/executor.go
mszostok: I also updated the command that is sent from `kc-cmd-builder` to:
- "kc-cmd-builder --verbs"
- "kc-cmd-builder --resource-type"
- "kc-cmd-builder --resource-name"
- "kc-cmd-builder --namespace"

(depends on which dropdown was changed)

pkosiec: Can we easily strip out actual resource name and namespace?
image
image

If not, let's go back to the previous impl, as this one would leak potentially private data 👍

pkg/execute/kubectl_cmd_builder.go
```diff
@@ -12,19 +12,23 @@ import (
...
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/strings/slices"
 
 	"github.com/kubeshop/botkube/pkg/bot/interactive"
 	"github.com/kubeshop/botkube/pkg/config"
 	"github.com/kubeshop/botkube/pkg/execute/kubectl"
 )
```
pkosiec, line 20: I noticed another problem with not properly resource name cleaned up. Watch this:


https://user-images.githubusercontent.com/7155799/195316607-3a55f1a3-8d0a-4b0f-9299-e5420c26593a.mp4


Also, how to clear the resource name? Can we append an option at the top ("None") or something like this?

(This can be followed-up on different PR)

pkosiec (approved):
LGTM but please fix the stuff with telemetry and also let's follow-up on the clearing text dropdown bug 👍

pkg/execute/executor.go
mszostok: Yes, we can 👍  I will fix that too 👍

pkg/execute/kubectl_cmd_builder.go
mszostok: I also spot this bug previously, I will fix it in the next PR. Thanks! I will also talk to Lacho about an option to clear this. Maybe sth like `@None` could be introduced but it still can be misleading  😞 and we are not able to easily clear the state of only one dropdown. I can clear it but for the whole section only 😞

pkosiec (approved):
👍 Tested latest telemetry, works well

PR #805, opened 2022-10-11
Interactive output filtering

huseyinbabal:
## Description

- Introduced "Filter Output" feature to message responses that has > 16 lines
- Supports both markdown responses and file uploads due to slack message size limit.

## Testing

- Checkout BotKube to your local computer
```bash
gh pr checkout 805 --repo=https://github.com/kubeshop/botkube
```
- Prepare `comm_config.yaml` with containing slack credentials as follows
<details>
<summary>Show</summary>

```yaml
communications:
  'default-group':
    # Settings for Slack
    socketSlack:
      enabled: true
      appToken: "xapp-1..."
      botToken: "xoxb-..."
      notification:
        type: "long"
      channels:
        'default':
          name: botkube-demo
          bindings:
            executors:
              - kubectl-backend
            sources:
              - k8s-all-events
              - k8s-recommendation-events

settings:
  clusterName: gke-playground
  configWatcher: true
  lifecycleServer:
    deployment:
      name: botkube
      namespace: botkube
    port: "2113"

configWatcher:
  enabled: false
  initialSyncTimeout: 0

executors:
  'kubectl-backend':
    kubectl:
      namespaces:
        include:
          - ".*"
      enabled: true
      commands:
        verbs: [ "run", "api-resources", "api-versions", "cluster-info", "describe", "diff", "explain", "get", "logs", "top", "auth", "delete" ]
        resources: [ "deployments", "pods", "namespaces", "daemonsets", "ingresses", "statefulsets", "storageclasses", "nodes", "configmaps", "services" ]
      defaultNamespace: backend
      restrictAccess: false

rbac:
  create: true
  rules:
    - apiGroups: ["*"]
      resources: ["*"]
      verbs: ["get", "watch", "list"]
    - apiGroups: [""]
      resources: ["pods"]
      verbs: ["delete", "create"]

analytics:
  disable: true

# -- Map of sources. Source contains configuration for Kubernetes events and sending recommendations.
# The property name under `sources` object is an alias for a given configuration. You can define multiple sources configuration with different names.
# Key name is used as a binding reference.
#
## Format: sources.<alias>
sources:
  'k8s-recommendation-events':
    displayName: "Kubernetes Recommendations"
    # -- Describes Kubernetes source configuration.
    kubernetes:
      # -- Describes configuration for various recommendation insights.
      recommendations:
        # -- Recommendations for Pod Kubernetes resource.
        pod:
          # -- If true, notifies about Pod containers that use `latest` tag for images.
          noLatestImageTag: true
          # -- If true, notifies about Pod resources created without labels.
          labelsSet: true
        # -- Recommendations for Ingress Kubernetes resource.
        ingress:
          # -- If true, notifies about Ingress resources with invalid backend service reference.
          backendServiceValid: true
          # -- If true, notifies about Ingress resources with invalid TLS secret reference.
          tlsSecretValid: true

  'k8s-all-events':
    # -- Describes Kubernetes source configuration.
    kubernetes:
      # -- Describes namespaces for every Kubernetes resources you want to watch or exclude.
      # These namespaces are applied to every resource specified in the resources list.
      # However, every specified resource can override this by using its own namespaces object.
      namespaces: &k8s-events-namespaces
        # Include contains a list of allowed Namespaces.
        # It can also contain a regex expressions:
        #  `- ".*"` - to specify all Namespaces.
        include:
          - ".*"
        # Exclude contains a list of Namespaces to be ignored even if allowed by Include.
        # It can also contain a regex expressions:
        #  `- "test-.*"` - to specif all Namespaces with `test-` prefix.
        # exclude: []

      # -- Describes events for every Kubernetes resources you want to watch or exclude.
      # These events are applied to every resource specified in the resources list.
      # However, every specified resource can override this by using its own events object.
      events:
        - create
        - delete
        - error

      # -- Describes the Kubernetes resources you want to watch.
      resources:
        - name: v1/pods             # Name of the resource. Resource name must be in group/version/resource (G/V/R) format
          # resource name should be plural (e.g apps/v1/deployments, v1/pods)

        #  namespaces:             # Overrides 'source'.kubernetes.namespaces
        #    include:
        #      - ".*"
        #    exclude: []
        - name: v1/services
        - name: networking.k8s.io/v1/ingresses
        - name: v1/nodes
        - name: v1/namespaces
        - name: v1/persistentvolumes
        - name: v1/persistentvolumeclaims
        - name: v1/configmaps
        - name: rbac.authorization.k8s.io/v1/roles
        - name: rbac.authorization.k8s.io/v1/rolebindings
        - name: rbac.authorization.k8s.io/v1/clusterrolebindings
        - name: rbac.authorization.k8s.io/v1/clusterroles
        - name: apps/v1/daemonsets
          events: # Overrides 'source'.kubernetes.events
            - create
            - update
            - delete
            - error
          updateSetting:
            includeDiff: true
            fields:
              - status.numberReady
        - name: batch/v1/jobs
          events: # Overrides 'source'.kubernetes.events
            - create
            - update
            - delete
            - error
          updateSetting:
            includeDiff: true
            fields:
              - status.conditions[*].type
        - name: apps/v1/deployments
          events: # Overrides 'source'.kubernetes.events
            - create
            - update
            - delete
            - error
          updateSetting:
            includeDiff: true
            fields:
              - status.availableReplicas
        - name: apps/v1/statefulsets
          events: # Overrides 'source'.kubernetes.events
            - create
            - update
            - delete
            - error
          updateSetting:
            includeDiff: true
            fields:
              - status.readyReplicas
        ## Custom resource example
        # - name: velero.io/v1/backups
        #   namespaces:
        #     include:
        #       - ".*"
        #     exclude:
        #       -
        #   events:
        #     - create
        #     - update
        #     - delete
        #     - error
        #   updateSetting:
        #     includeDiff: true
        #     fields:
        #       - status.phase

  'k8s-err-events':
    displayName: "Kubernetes Errors"

    # -- Describes Kubernetes source configuration.
    kubernetes:
      # -- Describes namespaces for every Kubernetes resources you want to watch or exclude.
      # These namespaces are applied to every resource specified in the resources list.
      # However, every specified resource can override this by using its own namespaces object.
      namespaces: *k8s-events-namespaces

      # -- Describes events for every Kubernetes resources you want to watch or exclude.
      # These events are applied to every resource specified in the resources list.
      # However, every specified resource can override this by using its own events object.
      events:
        - error

      # -- Describes the Kubernetes resources you want to watch.
      resources:
        - name: v1/pods
        - name: v1/services
        - name: networking.k8s.io/v1/ingresses
        - name: v1/nodes
        - name: v1/namespaces
        - name: v1/persistentvolumes
        - name: v1/persistentvolumeclaims
        - name: v1/configmaps
        - name: rbac.authorization.k8s.io/v1/roles
        - name: rbac.authorization.k8s.io/v1/rolebindings
        - name: rbac.authorization.k8s.io/v1/clusterrolebindings
        - name: rbac.authorization.k8s.io/v1/clusterroles
        - name: apps/v1/deployments
        - name: apps/v1/statefulsets
        - name: apps/v1/daemonsets
        - name: batch/v1/jobs
```

</details>

- Export required variables
```bash
export BOTKUBE_SETTINGS_KUBECONFIG=$KUBE_CONFIG_PATH
export BOTKUBE_CONFIG_PATHS="$(pwd)/comm_config.yaml"
```

- Run BotKube
```bash
go run cmd/botkube/main.go
```

## Screenshots
Screen Shot 2022-10-12 at 17 20 30
Screen Shot 2022-10-12 at 17 20 55

### DIFF
--- a/pkg/bot/slack.go
+++ b/pkg/bot/slack.go
@@ -259,8 +259,6 @@ func (b *Slack) send(msg slackMessage, req string, resp interactive.Message, onl
 		return fmt.Errorf("while reading Slack response: empty response for request %q", req)
 	}
 
-	var options = []slack.MsgOption{slack.MsgOptionText(markdown, false), slack.MsgOptionAsUser(true)}
-
 	// Upload message as a file if too long
 	if len(markdown) >= slackMaxMessageSize {
 		_, err := uploadFileToSlack(msg.Channel, resp, b.client)
@@ -269,6 +267,8 @@ func (b *Slack) send(msg slackMessage, req string, resp interactive.Message, onl
 		}
 		return nil
 	}
+
+	var options = []slack.MsgOption{slack.MsgOptionText(markdown, false), slack.MsgOptionAsUser(true)}
 
 	//if the message is from thread then add an option to return the response to the thread
 	if msg.ThreadTimeStamp != "" {
--- a/pkg/bot/slack.go
+++ b/pkg/bot/slack.go
@@ -259,12 +259,16 @@ func (b *Slack) send(msg slackMessage, req string, resp interactive.Message, onl
 		return fmt.Errorf("while reading Slack response: empty response for request %q", req)
 	}
 
+	var options = []slack.MsgOption{slack.MsgOptionText(markdown, false), slack.MsgOptionAsUser(true)}
+
 	// Upload message as a file if too long
 	if len(markdown) >= slackMaxMessageSize {
-		return uploadFileToSlack(msg.Channel, resp, b.client)
+		_, err := uploadFileToSlack(msg.Channel, resp, b.client)
+		if err != nil {
+			return err
+		}
+		return nil
 	}
-
-	var options = []slack.MsgOption{slack.MsgOptionText(markdown, false), slack.MsgOptionAsUser(true)}
 
 	//if the message is from thread then add an option to return the response to the thread
 	if msg.ThreadTimeStamp != "" {
@@ -373,7 +377,7 @@ func mdHeaderFormatter(msg string) string {
 	return fmt.Sprintf("*%s*", msg)
 }
 
-func uploadFileToSlack(channel string, resp interactive.Message, client *slack.Client) error {
+func uploadFileToSlack(channel string, resp interactive.Message, client *slack.Client) (*slack.File, error) {
 	params := slack.FileUploadParameters{
 		Filename:       "Response.txt",
 		Title:          "Response.txt",
@@ -382,10 +386,10 @@ func uploadFileToSlack(channel string, resp interactive.Message, client *slack.C
 		Channels:       []string{channel},
 	}
 
-	_, err := client.UploadFile(params)
+	file, err := client.UploadFile(params)
 	if err != nil {
-		return fmt.Errorf("while uploading file: %w", err)
+		return nil, fmt.Errorf("while uploading file: %w", err)
 	}
 
-	return nil
+	return file, nil
 }
--- a/pkg/bot/socketslack.go
+++ b/pkg/bot/socketslack.go
@@ -378,7 +378,6 @@ func (b *SocketSlack) send(event socketSlackMessage, req string, resp interactiv
 				break
 			}
 		}
-
 	}
 
 	if resp.ReplaceOriginal && event.ResponseURL != "" {
--- a/pkg/bot/socketslack.go
+++ b/pkg/bot/socketslack.go
@@ -197,9 +197,8 @@ func (b *SocketSlack) Start(ctx context.Context) error {
 						b.log.Debug("Ignoring callback as its source is an active modal")
 						continue
 					}
-
 					msg := socketSlackMessage{
-						Text:                resolveBlockActionCommand(*act),
+						Text:                resolveBlockActionCommand(*act, callback.Message.Msg.Blocks.BlockSet[0].(*slack.SectionBlock).Text.Text, b.BotName()),
 						Channel:             channelID,
 						ThreadTimeStamp:     callback.MessageTs,
 						TriggerID:           callback.TriggerID,
@@ -220,7 +219,7 @@ func (b *SocketSlack) Start(ctx context.Context) error {
 							act.ActionID = actID // normalize event
 
 							msg := socketSlackMessage{
-								Text:                resolveBlockActionCommand(act),
+								Text:                resolveBlockActionCommand(act, "", b.BotName()),
 								Channel:             callback.View.PrivateMetadata,
 								User:                callback.User.ID,
 								IsButtonClickOrigin: true,
@@ -234,6 +233,7 @@ func (b *SocketSlack) Start(ctx context.Context) error {
 				default:
 					b.log.Debugf("get unhandled event %s", callback.Type)
 				}
+
 			case socketmode.EventTypeErrorBadMessage:
 				b.log.Errorf("Bad message: %+v\n", event.Data)
 			case socketmode.EventTypeIncomingError:
@@ -504,7 +504,7 @@ func (b *SocketSlack) findAndTrimBotMention(msg string) (string, bool) {
 	return b.botMentionRegex.ReplaceAllString(msg, ""), true
 }
 
-func resolveBlockActionCommand(act slack.BlockAction) string {
+func resolveBlockActionCommand(act slack.BlockAction, cmd, botName string) string {
 	command := act.Value
 	switch act.Type {
 	// currently we support only interactive.MultiSelect option
@@ -519,6 +519,9 @@ func resolveBlockActionCommand(act slack.BlockAction) string {
 		//   @Botkube kcc --verbs get
 		//   @Botkube kcc --resource-type
 		command = fmt.Sprintf("%s %s", act.ActionID, act.SelectedOption.Value)
+	case "plain_text_input":
+		cmd = strings.Split(cmd, "`")[1]
+		command = fmt.Sprintf("%s %s --filter %s", botName, cmd, command)
 	}
 
 	return command
--- a/pkg/bot/socketslack.go
+++ b/pkg/bot/socketslack.go
@@ -203,7 +203,7 @@ func (b *SocketSlack) Start(ctx context.Context) error {
 						threadTs = callback.Message.Msg.ThreadTimestamp
 					}
 					msg := socketSlackMessage{
-						Text:                b.resolveBlockActionCommand(callback),
+						Text:                b.resolveBlockActionCommand(*act),
 						Channel:             channelID,
 						ThreadTimeStamp:     threadTs,
 						TriggerID:           callback.TriggerID,
@@ -224,7 +224,7 @@ func (b *SocketSlack) Start(ctx context.Context) error {
 							act.ActionID = actID // normalize event
 
 							msg := socketSlackMessage{
-								Text:                b.resolveBlockActionCommand(callback),
+								Text:                b.resolveBlockActionCommand(act),
 								Channel:             callback.View.PrivateMetadata,
 								User:                callback.User.ID,
 								IsButtonClickOrigin: true,
@@ -350,7 +350,9 @@ func (b *SocketSlack) send(event socketSlackMessage, req string, resp interactiv
 		if err != nil {
 			return err
 		}
-		resp.IgnoreOriginalResponse = true
+		resp = interactive.Message{
+			Inputs: resp.Inputs,
+		}
 	}
 
 	// we can open modal only if we have a TriggerID (it's available when user clicks a button)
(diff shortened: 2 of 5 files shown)


pkg/bot/socketslack.go
```diff
@@ -519,6 +538,21 @@ func resolveBlockActionCommand(act slack.BlockAction) string {
...
+		// b. If the original command is located inside an input block, we use blockID since the original
+		// command is propagated via that field.
+		blockType := callback.Message.Msg.Blocks.BlockSet[0].BlockType()
+		originalCmd := ""
+		if blockType == slack.MBTSection {
+			originalCmd = callback.Message.Msg.Blocks.BlockSet[0].(*slack.SectionBlock).Text.Text
+		} else if blockType == slack.MBTInput {
+			// blockID contains previously executed command.
+			originalCmd = callback.Message.Msg.Blocks.BlockSet[0].(*slack.InputBlock).BlockID
+		}
```
mszostok, line 553: IMO this:
```go
blockType := callback.Message.Msg.Blocks.BlockSet[0].BlockType() // TODO: check blockSet size
originalCmd := ""
if blockType == slack.MBTSection {
	originalCmd = callback.Message.Msg.Blocks.BlockSet[0].(*slack.SectionBlock).Text.Text
} else if blockType == slack.MBTInput {
	// blockID contains previously executed command.
	originalCmd = callback.Message.Msg.Blocks.BlockSet[0].(*slack.InputBlock).BlockID
}
```

can be simplified to this: (it's even more type safe)
```go
block := callback.Message.Msg.Blocks.BlockSet[0] // TODO: check blockSet size
originalCmd := ""
switch b := block.(type) {
case *slack.SectionBlock:
	originalCmd = b.Text.Text
case *slack.InputBlock:
	// blockID contains previously executed command.
	originalCmd = b.BlockID
}
```

pkg/execute/executor.go
```diff
@@ -40,6 +40,8 @@ const (
 	// Currently we support only `kubectl, so we
 	// override the message to human-readable command name.
 	humanReadableCommandListName = "Available kubectl commands"
+
+	lineLimitToShowFilter = 16
```
mszostok, line 44:
```suggestion
	lineLimitToShowFilter = 15
```

pkg/bot/socketslack.go
```diff
@@ -519,6 +538,21 @@ func resolveBlockActionCommand(act slack.BlockAction) string {
...
+		cmd := strings.Split(originalCmd, "`")[1]
+		command = fmt.Sprintf("%s %s --filter %s", b.BotName(), cmd, command)
```
mszostok, line 555:
```suggestion
		command = fmt.Sprintf("%s %s --filter=%q", b.BotName(), cmd, command)
```
It will be just less error-prone and more readable if we will do `--filter="  foo bar"` instead of `--filter=  foo bar`
Screen Shot 2022-10-12 at 18 12 21

pkg/bot/socketslack.go
```diff
@@ -519,6 +538,21 @@ func resolveBlockActionCommand(act slack.BlockAction) string {
...
+		// command is propagated via that field.
+		blockType := callback.Message.Msg.Blocks.BlockSet[0].BlockType()
+		originalCmd := ""
+		if blockType == slack.MBTSection {
+			originalCmd = callback.Message.Msg.Blocks.BlockSet[0].(*slack.SectionBlock).Text.Text
+		} else if blockType == slack.MBTInput {
+			// blockID contains previously executed command.
+			originalCmd = callback.Message.Msg.Blocks.BlockSet[0].(*slack.InputBlock).BlockID
+		}
+		cmd := strings.Split(originalCmd, "`")[1]
```
mszostok, line 554: You can use the `strings.Cut` which will not panic in this case:

```go
package bot

import (
	"fmt"
	"strings"
	"testing"

	"github.com/stretchr/testify/assert"
)

func TestWrongData(t *testing.T) {
	originalCmd := "some-data"
	assert.Panics(t, func() {
		cmd := strings.Split(originalCmd, "`")[1] // this panics
		fmt.Println(cmd)
	})

	_, after, _ := strings.Cut(originalCmd, "`") // this doesn't panic
	fmt.Println(after)
}

func TestCorrectData(t *testing.T) {
	originalCmd := "`kc get po -A` on `labs`"
	cmd := strings.Split(originalCmd, "`")[1]
	fmt.Println(cmd)

	_, after, _ := strings.Cut(originalCmd, "`")
	fmt.Println(after)
}
```

pkg/bot/socketslack.go
```diff
@@ -338,8 +343,14 @@ func (b *SocketSlack) send(event socketSlackMessage, req string, resp interactiv
...
 	// Upload message as a file if too long
+	var file *slack.File
+	var err error
 	if len(markdown) >= slackMaxMessageSize {
-		return uploadFileToSlack(event.Channel, resp, b.client)
+		file, err = uploadFileToSlack(event.Channel, resp, b.client)
+		if err != nil {
+			return err
+		}
+		resp.IgnoreOriginalResponse = true
```
mszostok, line 353: Personally I don't like that we introduce the new bool property just to "talk" between 2 layers. IMO we can just do that:

```suggestion
		// The message sections were already sent, we only need to
		// take care of sending inputs it there are any.
		resp = interactive.Message{
			Inputs: resp.Inputs,
		}
```

It's simple and says already what we want to achieve. Then the renderer doesn't need to be aware of what to do on `IgnoreOriginalResponse`.

mszostok:
Please add proper labels to this PR. Thanks!

pkg/bot/slack.go
```diff
@@ -259,13 +259,17 @@ func (b *Slack) send(msg slackMessage, req string, resp interactive.Message, onl
...
 	if len(markdown) >= slackMaxMessageSize {
-		return uploadFileToSlack(msg.Channel, resp, b.client)
+		_, err := uploadFileToSlack(msg.Channel, resp, b.client)
+		if err != nil {
+			return err
+		}
+		return nil
 	}
 
-	var options = []slack.MsgOption{slack.MsgOptionText(markdown, false), slack.MsgOptionAsUser(true)}
```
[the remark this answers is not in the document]
huseyinbabal: Ok, I was using options inside the if but now I don't need that anymore, reverted to previous location.

pkg/bot/socketslack.go
```diff
@@ -360,6 +371,13 @@ func (b *SocketSlack) send(event socketSlackMessage, req string, resp interactiv
 	//if the message is from thread then add an option to return the response to the thread
 	if event.ThreadTimeStamp != "" {
 		options = append(options, slack.MsgOptionTS(event.ThreadTimeStamp))
+	} else if file != nil {
+		for _, share := range file.Shares.Public {
+			if share[0].Ts != "" {
```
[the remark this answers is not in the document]
huseyinbabal: > Please add one more check to ensure we will not panic here 🙂
> 
> adding multi `slack.MsgOptionTS(share[0].Ts)` option doesn't make sens. Only one be used (the last one), so to avoid misunderstanding, let's add only one option 👍
> 
> in the comment above, I suggested an alternative approach.

Actually, once you send a message to slack, by default it always adds share section as the original message. However, maybe for some other cases it can be nil, let me add to be on the safe side :)

pkg/bot/socketslack.go
```diff
@@ -360,6 +371,13 @@ func (b *SocketSlack) send(event socketSlackMessage, req string, resp interactiv
 	//if the message is from thread then add an option to return the response to the thread
 	if event.ThreadTimeStamp != "" {
```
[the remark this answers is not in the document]
huseyinbabal: why not, added

pkg/bot/socketslack.go
huseyinbabal: This suggestion confuses me since I already used `blockID` to propagate original command, but for only pure input actions. For the normal flow where there is no file upload, I already get blocks and use the first one. It contains original command already.

pkg/bot/socketslack.go
huseyinbabal: I am aware of that, but there are cases we have both sections and inputs. So, I cannot decide on something based on the component inside the interactive message.

pkg/bot/socketslack.go
```diff
@@ -197,11 +197,15 @@ func (b *SocketSlack) Start(ctx context.Context) error {
 						b.log.Debug("Ignoring callback as its source is an active modal")
 						continue
 					}
-
+					// Use thread's TS if interactive call triggered within thread.
```
mszostok, line 200: perfect 👌

pkg/bot/socketslack.go
mszostok: Now I understand your previous approach 👍 Thanks for clarification 👍

pkg/bot/socketslack.go
mszostok: discussed f2f :)

mszostok (approved):
LGTM 👍

PR #808, opened 2022-10-11
Output filtering

### LINKED ISSUE #742: Output Filtering
A user must be able to filter output from BotKube executor commands (e.g. kubectl) with simple string matching. This should be analogous to how users use grep when running kubectl from the command line. grep and grep-like syntax are not requirements.

**AC**

- Command output can be filtered using simple string matching
- Filtering is a CommandFlag and usable across any executor command
- Only matching output lines are returned to the user
- There is an indication that output is filtered
- There is an indication when there is output but no filter matches
- There is an indication when there is no output (i.e. no match/no output errors are differentiated)

Nice to have: more advanced filtering than simple string matching (e.g. RegEx, grep-style syntax, etc.). This should only be implemented if it comes from an existing package and we can do it with a clean UX.

## Notes
- Behavior - one of them:
    **- Dedicated flag (`@botkube kubectl logs <pod> --filter 'grep string and maybe params'`)**
    - ~Piping `@botkube kubectl logs <pod> | filter "str"`~
- Support it globally - kubectl and other commands 
- Naming: filter - we won't support full `grep` functionality
- No highlight on communication platforms
- Collect anonymous analytics
- Only `grep` for now / or instead of using binary, use some existing Go library
- Include it in integration tests
- **Interactive option - out of scope -> will be done in a separate task**

ezodude:
## Description

- Enables output filtering of executor based commands with a text output.

## Analytics Reporting

`ReportCommand(platform config.CommPlatformIntegration, command string, isButtonClickOrigin, withFilter bool) error` has been extended to include `withFilter` to indicate whether a filter was used when running the command.

## Testing

Install BotKube:
```shell
cat > /tmp/results-filter.yaml << ENDOFFILE
executors:
  'kubectl-enabled':
    kubectl:
      namespaces:
        include:
          - ".*"
        exclude: []
      enabled: true
      commands:
        verbs: ["api-resources", "api-versions", "cluster-info", "describe", "explain", "get", "logs", "top"]
        resources: ["deployments", "pods", "namespaces", "daemonsets", "statefulsets", "storageclasses", "nodes", "configmaps", "services"]
      defaultNamespace: default
      restrictAccess: false

communications:
  'default-group':
    socketSlack:
      enabled: true
      channels:
        'default':
          name: 'random'
          bindings:
            executors:
              - kubectl-enabled
            sources:
              - k8s-err-events
              - k8s-recommendation-events
      botToken: 'xoxb-...'
      appToken: 'xapp-...'
      notification:
        type: short

settings:
  clusterName: kind

  log:
    level: debug
    disableColors: false

analytics:
  disable: true

image:
  tag: '808-PR'
  pullPolicy: Always  
ENDOFFILE
```
helm install botkube ./helm/botkube -n botkube --create-namespace  /tmp/results-filter.yaml

### Run commands on botkube-test channel on Slack:
```
@BotKube get po -n kube-system --filter="proxy"
```
```
@Botkube commands list --filter=“api”
```
```
@BotKube get po -n kube-system --filter=proxy
```
```
@BotKube get po --filter="proxy" -n kube-system
```
```
@BotKube logs pods/kube-proxy-9qzj2 -n kube-system --filter="Detected node"
```

## Related issue(s)

Resolves #742


pkg/execute/executor_filter.go
```diff
@@ -0,0 +1,78 @@
...
+
+	return strings.TrimSuffix(out.String(), "\n")
+}
+
+// extractExecutorFilter extracts an executorFilter based on
+// the presence or absence of the "--filter=xxx" flag.
+// It also returns passed in executor command minus the
+// flag to be executed by downstream executors and if a filter flag was detected.
+func extractExecutorFilter(cmd string) (executorFilter, string, bool) {
+	r, _ := regexp.Compile(`--filter[=|(' ')]('.*?'|".*?"|\S+)`)
```
mszostok, line 64: we should handle the error or define it at the top of this package, which is quite common:

```go
filterFlagRegex := regexp.MustCompile(`--filter[=|(' ')]('.*?'|".*?"|\S+)`)
```

mszostok, line 64: Do you know why it's reporting this issue?
Screen Shot 2022-10-12 at 12 44 09

pkg/execute/executor_filter_test.go
```diff
@@ -0,0 +1,120 @@
...
+
+func TestExtractExecutorFilter(t *testing.T) {
+	testCases := []struct {
+		name           string
+		cmd            string
+		extractedCmd   string
+		text           string
+		filterApplied  string
+		filterDetected bool
+	}{
```
mszostok, line 69: please also handle this case:
```go
		{
			name:           "extract double quoted text filter in the middle of the command",
			cmd:            `kubectl get po  --filter="kind" --filter="test" --filter -n kube-system`,
			extractedCmd:   "kubectl get po  -n kube-system",
			text:           `etcd-control-plane                      1/1     Running   0          86m`,
			filterApplied:  "",
			filterDetected: true,
		},
```

In general:
1. What we do when user specifies the `--fliter` multiple times?
2. What we do when user specifies `--filter` (doesn't have the value)?
3. What we do when user specifies  `--filter      ` (multiple spaces)?
4. Do we want to support only such simple matching? We estimated task for 5MD, so I expected at least regex support. It was also mentioned in notes: 
   > Only grep for now / or instead of using binary, use some existing Go library

mszostok:
Please add proper labels to this PR. Thanks!

pkg/execute/executor_filter.go
```diff
@@ -0,0 +1,78 @@
...
+
+// Apply implements executorFilter to apply filtering.
+func (f *executorTextFilter) Apply(text string) string {
+	var out strings.Builder
+
+	scanner := bufio.NewScanner(strings.NewReader(text))
+	for scanner.Scan() {
+		scanned := scanner.Bytes()
+		if bytes.Contains(scanned, f.value) {
+			out.Write(bytes.TrimSpace(scanned))
```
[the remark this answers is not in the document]
ezodude: In the beginning, I didn't trim the space. But, it looked bad when viewing on Slack.

pkg/execute/executor.go
```diff
@@ -26,9 +26,9 @@ var (
 
 const (
 	unsupportedCmdMsg   = "Command not supported. Please use 'help' to see supported commands."
-	filterNameMissing   = "You forgot to pass filter name. Please pass one of the following valid filters:\n\n%s"
-	filterEnabled       = "I have enabled '%s' filter on '%s' cluster."
-	filterDisabled      = "Done. I won't run '%s' filter on '%s' cluster."
+	filterNameMissing   = "You forgot to pass Filter name. Please pass one of the following valid filters:\n\n%s"
+	filterEnabled       = "I have enabled '%s' Filter on '%s' cluster."
+	filterDisabled      = "Done. I won't run '%s' Filter on '%s' cluster."
```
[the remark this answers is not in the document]
ezodude: This should be tidied - I think a rename that touched more than it was supposed to.

pkg/execute/executor_filter.go
ezodude: Will use `MustCompile`.

Other parts of the code used this pattern `r, _ := regexp.Compile(`--filter[=|(' ')]('.*?'|".*?"|\S+)`)`, so I thought it might be good to stick with it.

ezodude: Nope.

On GoLand, `utils.go` -> `utils.GetClusterNameFromKubectlCmd` has the same issue.

pkg/execute/executor_filter_test.go
```diff
@@ -0,0 +1,120 @@
...
+
+	var txFilter executorFilter = newExecutorTextFilter("kind")
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			assert.Equal(t, tc.expected, txFilter.Apply(tc.text))
+		})
+	}
+}
+
+func TestExtractExecutorFilter(t *testing.T) {
```
mszostok, line 61: If you add one more space, I got such error:
Screen Shot 2022-10-12 at 14 33 52


and it's can be a quite popular to do that by mistake

ezodude: It makes sense to fix the edge cases before thinking about regular expressions.

ezodude:
@mszostok rebased from master and ready - that was painful!

The executor was changed a lot while I was also working on it.

mszostok (commented):
From the unit-test point of view, the behavior LGTM 👍 
However, e2e it doesn't work yet, but as discussed I will go back to testing once you will do the final round of rebasing and rechecking 👍

pkg/execute/executor.go
```diff
@@ -172,18 +150,18 @@ func (e *DefaultExecutor) Execute() interactive.Message {
...
 	if e.kubectlExecutor.CanHandle(e.conversation.ExecutorBindings, args) {
-		e.reportCommand(e.kubectlExecutor.GetCommandPrefix(args))
-		out, err := e.kubectlExecutor.Execute(e.conversation.ExecutorBindings, e.message, e.conversation.IsAuthenticated)
+		e.reportCommand(e.kubectlExecutor.GetCommandPrefix(args), execFilter.IsActive())
+		out, err := e.kubectlExecutor.Execute(e.conversation.ExecutorBindings, execFilter.FilteredCommand(), e.conversation.IsAuthenticated)
 		switch {
 		case err == nil:
 		case IsExecutionCommandError(err):
-			return response(err.Error())
+			return e.respond(err.Error(), rawCmd, execFilter.FilteredCommand(), botName, "")
```
mszostok, line 158:
```suggestion
			return e.respond(err.Error(), rawCmd, execFilter.FilteredCommand(), botName)
```

pkg/execute/executor.go
```diff
@@ -204,33 +182,33 @@ func (e *DefaultExecutor) Execute() interactive.Message {
...
 	cmds := executorsRunner{
 		"help": func() (interactive.Message, error) {
-			e.reportCommand(args[0])
+			e.reportCommand(args[0], false)
 			return interactive.NewHelpMessage(e.platform, clusterName, botName).Build(), nil
 		},
 		"ping": func() (interactive.Message, error) {
 			res := e.runVersionCommand("ping")
-			return response(fmt.Sprintf("pong\n\n%s", res)), nil
+			return e.respond(fmt.Sprintf("pong\n\n%s", res), rawCmd, execFilter.FilteredCommand(), botName, ""), nil
```
mszostok, line 190:
```suggestion
			return e.respond(fmt.Sprintf("pong\n\n%s", res), rawCmd, execFilter.FilteredCommand(), botName), nil
```

pkg/execute/executor.go
```diff
@@ -204,33 +182,33 @@ func (e *DefaultExecutor) Execute() interactive.Message {
...
 		"commands": func() (interactive.Message, error) {
-			res, err := e.runInfoCommand(args)
-			return response(res, humanReadableCommandListName), err
+			res, err := e.runInfoCommand(args, execFilter.IsActive())
+			return e.respond(execFilter.Apply(res), rawCmd, execFilter.FilteredCommand(), botName, humanReadableCommandListName), err
 		},
 		"notifier": func() (interactive.Message, error) {
 			res, err := e.notifierExecutor.Do(ctx, args, e.commGroupName, e.platform, e.conversation, clusterName, e.notifierHandler)
-			return response(res), err
+			return e.respond(res, rawCmd, execFilter.FilteredCommand(), botName, ""), err
```
mszostok, line 205:
```suggestion
			return e.respond(res, rawCmd, execFilter.FilteredCommand(), botName), err
```

pkg/execute/executor.go
```diff
@@ -239,20 +217,43 @@ func (e *DefaultExecutor) Execute() interactive.Message {
 	switch {
 	case err == nil:
 	case errors.Is(err, errInvalidCommand):
-		return response(incompleteCmdMsg)
+		return e.respond(incompleteCmdMsg, rawCmd, execFilter.FilteredCommand(), botName, "")
```
mszostok, line 220:
```suggestion
		return e.respond(incompleteCmdMsg, rawCmd, execFilter.FilteredCommand(), botName)
```

pkg/execute/executor.go
```diff
@@ -239,20 +217,43 @@ func (e *DefaultExecutor) Execute() interactive.Message {
...
 	case errors.Is(err, errUnsupportedCommand):
-		return response(unsupportedCmdMsg)
+		return e.respond(unsupportedCmdMsg, rawCmd, execFilter.FilteredCommand(), botName, "")
```
mszostok, line 222:
```suggestion
		return e.respond(unsupportedCmdMsg, rawCmd, execFilter.FilteredCommand(), botName)
```

pkg/execute/executor.go
```diff
@@ -239,20 +217,43 @@ func (e *DefaultExecutor) Execute() interactive.Message {
...
 	case IsExecutionCommandError(err):
-		return response(err.Error())
+		return e.respond(err.Error(), rawCmd, execFilter.FilteredCommand(), botName, "")
 	default:
-		e.log.Errorf("while executing command %q: %s", command, err.Error())
+		e.log.Errorf("while executing command %q: %s", execFilter.FilteredCommand(), err.Error())
 		internalErrorMsg := fmt.Sprintf(internalErrorMsgFmt, clusterName)
-		return response(internalErrorMsg)
+		return e.respond(internalErrorMsg, rawCmd, execFilter.FilteredCommand(), botName, "")
```
mszostok, line 228:
```suggestion
		return e.respond(internalErrorMsg, rawCmd, execFilter.FilteredCommand(), botName)
```

ezodude: Already sorted!

pkg/execute/executor_filter.go
```diff
@@ -0,0 +1,158 @@
...
+	filterFlagRegex, err := regexp.Compile(fmt.Sprintf(`--filter[=|(' ')]*('%s'|"%s"|%s)("|')*`,
+		escapedFilterVal,
+		escapedFilterVal,
+		escapedFilterVal))
+	if err != nil {
+		return nil, errors.New("could not extract provided filter")
+	}
+
+	matches := filterFlagRegex.FindStringSubmatch(cmd)
+	return newExecutorTextFilter(filterVal, strings.ReplaceAll(cmd, fmt.Sprintf(" %s", matches[0]), "")), nil
```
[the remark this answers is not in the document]
mszostok: to reproduce with unit-test:
```
cmd:           `kubectl get po --filter “botkube. ” -n kube-system`,
```

besides adding a len guard, do you know why it's not working 🤔 ?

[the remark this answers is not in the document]
ezodude: Cool, that's an actual issue - will investigate.

ezodude:
- Updates to ensure command names are visible for commands like `ping`.
- E2E tests are now fixed.

pkg/execute/executor_filter.go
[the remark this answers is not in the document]
ezodude: This was a rebase issue. The `filterVal` was regular expression escaped twice. (2 commits auto rebased). Now fixed. 

I added extra unit tests to cover special characters and file paths, etc..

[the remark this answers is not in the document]
ezodude: @mszostok all ready to go.

pkg/execute/executor.go
```diff
@@ -116,19 +116,20 @@ const (
...
+	empty := interactive.Message{}
+	rawCmd := utils.RemoveAnyHyperlinks(e.message)
+	rawCmd = strings.NewReplacer(`“`, `"`, `”`, `"`, `‘`, `"`, `’`, `"`).Replace(rawCmd)
+	clusterName := e.cfg.Settings.ClusterName
+	inClusterName := utils.GetClusterNameFromKubectlCmd(rawCmd)
+	botName := e.notifierHandler.BotName()
+
+	execFilter, err := extractExecutorFilter(rawCmd)
+	if err != nil {
+		return e.respond(err.Error(), rawCmd, execFilter.FilteredCommand(), botName)
```
mszostok, line 129: this one can panic as the `execFilter` is nil in case of error: e.g. `kc get po -A --filter “botkube” --filter “test”`

mszostok (approved):
I found one more panic, but besides that it LGTM 👍

pkg/execute/executor.go
ezodude: Thx! will investigate and fix.

ezodude: Done.

PR #809, opened 2022-10-12
The CONTRIBUTING.md file is outdated

### LINKED ISSUE #804: The CONTRIBUTING.md file is outdated
## Description

Instructions for installing Botkube locally are outdated in [CONTRIBUTING.md.](https://github.com/kubeshop/botkube/blob/main/CONTRIBUTING.md#build-and-install-on-kubernetes) Following them will result in such installation error:
```
Error: INSTALLATION FAILED: template: botkube/templates/service.yaml:1:42: executing "botkube/templates/service.yaml" at <include "botkube.communication.team.enabled" $>: error calling include: template: botkube/templates/_helpers.tpl:55:11: executing "botkube.communication.team.enabled" at <$val.teams.enabled>: nil pointer evaluating interface {}.enabled
```

## Acceptance Criteria
- fix the CONTRIBUTING.md to reflect the new way of installing Botkube
- if possible, link to the official installation docs which are tested each release, so we will have a single source of truth.
- describe issue with the `botkube-system` config map. It must be deleted before executing test for the second time on the same cluster.

ezodude:
## Description

- Updated Botkube helm installation instructions for contributors.
- Linked to docs as reference installation instructions when encountering problems.
- Added a new step to ensure a Slack integration is already setup.
- General tidy up.

## Related issue(s)

Resolves #804

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -78,7 +78,6 @@ This section describes how to build and run Botkube from source code.
    --set image.tag=v9.99.9-dev \
    --set image.pullPolicy=<IMAGE_PULL_POLICY e.g. Always>
    Check [values.yaml](./helm/botkube/values.yaml) for default options.
-   
 ### Build and run locally
 
 For faster development, you can also build and run Botkube outside K8s cluster.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -65,15 +65,16 @@ This section describes how to build and run Botkube from source code.
      Where `<your_account>` is Docker hub or any other registry provider account to which you can push the image.
 
 2. Install Botkube with any of communication platform configured, according to [the installation instructions](https://botkube.io/docs/installation/). During the Helm chart installation step, set the following flags:
-
-
    
    ```sh
    --set image.registry=<IMAGE_REGISTRY e.g. docker.io> \
    --set image.repository=<your_account>/botkube \
    --set image.tag=v9.99.9-dev \
    --set image.pullPolicy=<IMAGE_PULL_POLICY e.g. Always>
+   ```
+   
    Check [values.yaml](./helm/botkube/values.yaml) for default options.
+   
 ### Build and run locally
 
 For faster development, you can also build and run Botkube outside K8s cluster.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -67,9 +67,6 @@ This section describes how to build and run Botkube from source code.
 2. Install Botkube with any of communication platform configured, according to [the installation instructions](https://botkube.io/docs/installation/). During the Helm chart installation step, set the following flags:
 
 
-3. Deploy the newly created image in your cluster:
-   
-   Use [the instructions here](https://botkube.io/docs/installation/socketslack/#install-botkube-backend-in-kubernetes-cluster) to Helm install Botkube while updating the following flags:  
    
    ```sh
    --set image.registry=<IMAGE_REGISTRY e.g. docker.io> \
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -157,7 +157,23 @@ For faster development, you can also build and run Botkube outside K8s cluster.
   ```
   This will run the `golangci-lint` tool to lint the Go code.
 
-- [Run e2e tests](./test/README.md)
+### Run the e2e tests
+
+Here [are the details you need](./test/README.md) to set up and run the e2e tests.
+
+#### Clear botkube-system between test runs
+
+Botkube tracks whether the initial help message was sent or not to minimise spam. This is tracked in a `botkube-system` ConfigMap.
+
+Please ensure this ConfigMap is removed before rerunning any e2e tests.
+
+```shell
+kubectl delete cm botkube-system -n botkube # or the namespace where Botkube is installed 
+```
+
+If you don't remove the ConfigMap, any e2e tests looking to verify that a help message is displayed will error. This also stops the rest of the e2e tests from running.  
+
+### Create a Pull Request
 
 - Make sure your pull request has [good commit messages](https://chris.beams.io/posts/git-commit/):
   - Separate subject from body with a blank line
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -12,7 +12,7 @@ Follow the [botkube-docs/CONTRIBUTING.md](https://github.com/kubeshop/botkube-do
 
 ## Build and run Botkube from source code
 
-This section describes how to build and run the Botkube from the source code.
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -60,25 +62,34 @@ This section describes how to build and run the Botkube from the source code.
...
-   --set communications.slack.channel=<SLACK_CHANNEL_NAME> \
-   --set communications.slack.token=<SLACK_API_TOKEN_FOR_THE_BOT> \
-   --set settings.clusterName=<CLUSTER_NAME> \
-   --set settings.kubectl.enabled=<ALLOW_KUBECTL> \
-   --set image.registry=<image_registry e.g. docker.io> \
+   --set communications.default-group.socketSlack.enabled=true \
+   --set communications.default-group.socketSlack.channels.default.name=<SLACK_CHANNEL_NAME> \
+   --set communications.default-group.socketSlack.appToken=<SLACK_API_APP_TOKEN> \
+   --set communications.default-group.socketSlack.botToken=<SLACK_API_BOT_TOKEN> \
+   --set settings.clusterName=${CLUSTER_NAME} \
```
pkosiec, line 79: To be consistent with the rest of the flags:
```suggestion
   --set settings.clusterName=<CLUSTER_NAME> \
```

But please read my different comment as probably we could get rid of this line at all 👍

CONTRIBUTING.md
```diff
@@ -12,7 +12,7 @@ Follow the [botkube-docs/CONTRIBUTING.md](https://github.com/kubeshop/botkube-do
```
pkosiec, line 12: I think we're still missing this AC:

> describe issue with the botkube-system config map. It must be deleted before executing test for the second time on the same cluster.


See Mateusz's comment: https://botkube.slack.com/archives/CG7TTBXM2/p1665497211898899?thread_ts=1665483992.804649&cid=CG7TTBXM2

CONTRIBUTING.md
```diff
@@ -60,25 +62,34 @@ This section describes how to build and run the Botkube from the source code.
...
+   --set communications.default-group.socketSlack.enabled=true \
+   --set communications.default-group.socketSlack.channels.default.name=<SLACK_CHANNEL_NAME> \
+   --set communications.default-group.socketSlack.appToken=<SLACK_API_APP_TOKEN> \
+   --set communications.default-group.socketSlack.botToken=<SLACK_API_BOT_TOKEN> \
+   --set settings.clusterName=${CLUSTER_NAME} \
+   --set executors.kubectl-read-only.kubectl.enabled=true \
+   --set image.registry=<IMAGE_REGISTRY e.g. docker.io> \
    --set image.repository=<your_account>/botkube \
    --set image.tag=v9.99.9-dev \
+   --set image.pullPolicy=<IMAGE_PULL_POLICY e.g. Always>
```
pkosiec, line 84: What do you think to get rid of the full command (which would be hard to maintain + it works with Slack only), and instead say "hey, follow the Slack/Discord etc. installation docs as usual, but during Helm installation provide additional flags to override the image".

And there would be just the flags listed:
```
   --set image.registry=<IMAGE_REGISTRY e.g. docker.io> \
   --set image.repository=<your_account>/botkube \
   --set image.tag=v9.99.9-dev \
   --set image.pullPolicy=<IMAGE_PULL_POLICY e.g. Always>
```

pkosiec (commented):
Looks good overall, 2 small comments 🙂

CONTRIBUTING.md
ezodude: Will handle.

CONTRIBUTING.md
ezodude: That was my initial thinking too. But, the duplication here makes it a bit easier for contributors to just get up and running.

pkosiec: ...As long as the commands are up to date, and as long as the contributor wants to develop Botkube against Slack 🙂 All the options are well explained under installation docs, and the description is missing from here. So everybody still will end up navigating to the installation docs. I don't think passing additional parameters during the `helm install` is any harder than this 🤔 
While I'd still vote for the simpler option, if you don't agree it's totally alright - let's create a quick poll on Slack and select the approach the whole team likes the most 👍

CONTRIBUTING.md
ezodude: Done.

CONTRIBUTING.md
```diff
@@ -60,25 +62,25 @@ This section describes how to build and run the Botkube from the source code.
      docker tag ghcr.io/kubeshop/botkube:v9.99.9-dev-amd64 <your_account>/botkube:v9.99.9-dev
      docker push <your_account>/botkube:v9.99.9-dev
      ```
-     Where `<your_account>` is Docker hub account to which you can push the image.
+     Where `<your_account>` is Docker hub or any other registry provider account to which you can push the image.
+
+2. Ensure you have a Slack installation already setup:
```
pkosiec, line 67: I'd merge point 2 and 3 and have a single item and make it even more general - after all the Slack integration is not required for the development setup. What do you think about something like:

```suggestion
2. Install Botkube with any of communication platform configured, according to [the installation instructions](https://botkube.io/docs/installation/). During the Helm chart installation step, set the following flags:
```

+ removing sections below (suggestions incoming as well)

CONTRIBUTING.md
```diff
@@ -60,25 +62,25 @@ This section describes how to build and run the Botkube from the source code.
...
 
-2. Deploy the newly created image in your cluster:
+   Please refer to the [Slack installation instructions](https://botkube.io/docs/installation/socketslack/) to ensure you have a valid integration.
```
pkosiec, line 69: (Cannot add it as a single suggestion, but if you agree with the suggestion above, please apply this and it will automatically merge these sections)
```suggestion
```

CONTRIBUTING.md
```diff
@@ -60,25 +62,25 @@ This section describes how to build and run the Botkube from the source code.
...
 
+3. Deploy the newly created image in your cluster:
+   
+   Use [the instructions here](https://botkube.io/docs/installation/socketslack/#install-botkube-backend-in-kubernetes-cluster) to Helm install Botkube while updating the following flags:
```
pkosiec, line 73: (Cannot add it as a single suggestion, but if you agree with the suggestion above, please apply this and it will automatically merge these sections)
```suggestion

```

pkosiec (commented):
Really minor comments with suggestions ready to add to batch & commit from the GitHub UI, as long as you agree with them 🙂 That should be really quick - let me know when it's ready to approve ✅

ezodude:
Done.

pkosiec (approved):
🚀

Issue #772, opened 2022-09-28
Print in e2e message diff between expected and got message


mszostok:
## Overview

Currently, when e2e test expectation fails, we see in error logs all messages that were posted on a given channel with a message `while waiting for condition: last error: message assertion function returned false; fetched messages:`:
<details>
  <summary>Example logs, click me</summary>
  
```
=== RUN   TestSlack/Executor/Get_Pods_with_k8s_prefix_kubectl
    bots_test.go:439: 
        	Error Trace:	bots_test.go:439
        	Error:      	Received unexpected error:
        	            	while waiting for condition: last error: message assertion function returned false; fetched messages: []slack.Message{
        	            	  slack.Message{
        	            	    Msg: slack.Msg{
        	            	      Type: "message",
        	            	      User: "U03G3MQLFC2",
        	            	      Text: "`kubectl get pods` on `sample`\n```\nNo resources found in default namespace.\n```\n",
        	            	      Timestamp: "1664352221.046189",
        	            	      BotID: "B03GTEUMJQ0",
        	            	      BotProfile: &slack.BotProfile{
        	            	        AppID: "AF5DZLHPC",
        	            	        Icons: &slack.Icons{
        	            	          Image36: "https://avatars.slack-edge.com/2019-01-15/523890478611_539be8a0b4811[68](https://github.com/kubeshop/botkube/actions/runs/3141732818/jobs/5104649726#step:7:69)6d310_36.png",
        	            	          Image48: "https://avatars.slack-edge.com/2019-01-15/523890478611_539be8a0b4811686d310_48.png",
        	            	          Image[72](https://github.com/kubeshop/botkube/actions/runs/3141732818/jobs/5104649726#step:7:73): "https://avatars.slack-edge.com/2019-01-15/523890478611_539be8a0b4811686d310_72.png",
        	            	        },
        	            	        ID: "B03GTEUMJQ0",
        	            	        Name: "BotKube",
        	            	        TeamID: "T03GG4212N5",
        	            	        Updated: 16566822[76](https://github.com/kubeshop/botkube/actions/runs/3141732818/jobs/5104649726#step:7:77),
        	            	      },
        	            	      Team: "T03GG4212N5",
        	            	      Blocks: slack.Blocks{
        	            	        BlockSet: []slack.Block{
        	            	          &slack.RichTextBlock{
        	            	            Type: "rich_text",
        	            	            BlockID: "HSr",
        	            	            Elements: []slack.RichTextElement{
[...]

PR #811, opened 2022-10-12
Print diff between expected and actual messages in e2e tests

wangyangjun:
## Description
The return type of function MessageAssertion is updated from `bool` to `(bool, int, string)`.
Other corresponding changes are made base on this change. 
- The returned `int` value indicates the number of matched blocks between expect message and actual got message.
- The returned `string` value is the diff between expect message and actual got message.

This PR compares number of matched blocks and print the diff corresponding to the maximum
value of number of matched blocks.

## Related issue
Resolves: #772

### DIFF
--- a/test/e2e/discord_driver_test.go
+++ b/test/e2e/discord_driver_test.go
@@ -274,8 +274,8 @@ func (d *discordTester) WaitForMessagePostedWithAttachment(userID, channelID str
 
 	var fetchedMessages []*discordgo.Message
 	var lastErr error
-	var common int
 	var diffMessage string
+	common := -1
 
 	err := wait.Poll(pollInterval, d.cfg.MessageWaitTimeout, func() (done bool, err error) {
 		messages, err := d.cli.ChannelMessages(channelID, 1, "", "", "")
--- a/test/e2e/discord_driver_test.go
+++ b/test/e2e/discord_driver_test.go
@@ -132,20 +132,30 @@ func (d *discordTester) InviteBotToChannel(_ *testing.T, _ string) {
 }
 
 func (d *discordTester) WaitForMessagePostedRecentlyEqual(userID, channelID, expectedMsg string) error {
-	return d.WaitForMessagePosted(userID, channelID, recentMessagesLimit, func(msg string) bool {
-		return strings.EqualFold(msg, expectedMsg)
+	return d.WaitForMessagePosted(userID, channelID, recentMessagesLimit, func(msg string) (bool, int, string) {
+		if !strings.EqualFold(expectedMsg, msg) {
+			count := countMatchBlock(expectedMsg, msg)
+			msgDiff := diff(expectedMsg, msg)
+			return false, count, msgDiff
+		}
+		return true, 0, ""
 	})
 }
 
 func (d *discordTester) WaitForLastMessageContains(userID, channelID, expectedMsgSubstring string) error {
-	return d.WaitForMessagePosted(userID, channelID, 1, func(msg string) bool {
-		return strings.Contains(msg, expectedMsgSubstring)
+	return d.WaitForMessagePosted(userID, channelID, 1, func(msg string) (bool, int, string) {
+		return strings.Contains(msg, expectedMsgSubstring), 0, ""
 	})
 }
 
 func (d *discordTester) WaitForLastMessageEqual(userID, channelID, expectedMsg string) error {
-	return d.WaitForMessagePosted(userID, channelID, 1, func(msg string) bool {
-		return msg == expectedMsg
+	return d.WaitForMessagePosted(userID, channelID, 1, func(msg string) (bool, int, string) {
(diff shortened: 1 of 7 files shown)


wangyangjun:
One example of failed e2e test case result:

Screenshot 2022-10-12 at 23 21 34
Screenshot 2022-10-12 at 23 21 54

pkosiec (commented):
Great Job Reaction - XreQmk7ETCak0

I'm still finalizing testing, but that looks really good! A few small comments.

test/e2e/k8s_helpers_test.go
```diff
@@ -171,3 +172,34 @@ func updateEnv(existing []v1.EnvVar, env []v1.EnvVar, remove []string) []v1.EnvV
...
+// countMatchBlock count the number of lines matched between two strings
+func countMatchBlock(expect string, actual string) int {
+	matcher := difflib.NewMatcher(difflib.SplitLines(expect), difflib.SplitLines(actual))
+	matches := matcher.GetMatchingBlocks()
+	count := 0
+	for _, match := range matches {
+		count += match.Size
+	}
+	return count
+}
```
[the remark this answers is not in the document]
wangyangjun: Nice suggestion.

test/e2e/slack_driver_test.go
```diff
@@ -135,26 +135,41 @@ func (s *slackTester) InviteBotToChannel(t *testing.T, channelID string) {
 }
```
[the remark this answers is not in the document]
wangyangjun: Do we support diff for msg only or support diff for title, color and msg?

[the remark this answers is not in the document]
wangyangjun: From this comment, it seems we only support diff for msg -- https://github.com/kubeshop/botkube/blob/7cfcfba0ac82bdb44c6451f6de54d0a2434ddcf3/test/e2e/slack_driver_test.go#L273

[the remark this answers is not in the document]
pkosiec: @wangyangjun Oh you're right. Skip the `WaitForMessagePostedWithFileUpload` then 🙂 
And yes, I was referring to the message. Thank you!

pkosiec (approved):
Tested once again and it works well: https://github.com/pkosiec/botkube/actions/runs/3247955891/jobs/5328621321

Thank you very much for your contribution @wangyangjun!

test/e2e/discord_driver_test.go
```diff
@@ -257,6 +274,8 @@ func (d *discordTester) WaitForMessagePostedWithAttachment(userID, channelID str
 
 	var fetchedMessages []*discordgo.Message
 	var lastErr error
+	var diffMessage string
+	common := -1
```
pkosiec, line 278: Can we revert this change to the previous form? I think the behavior when initial `common` is 0 is better, as we won't print a diff when there are no common blocks. What do you think?

```suggestion
	var common int
```

test/e2e/slack_driver_test.go
```diff
@@ -245,6 +265,9 @@ func (s *slackTester) WaitForMessagePostedWithFileUpload(userID, channelID strin
 func (s *slackTester) WaitForMessagePostedWithAttachment(userID, channelID string, assertFn AttachmentAssertion) error {
 	var fetchedMessages []slack.Message
 	var lastErr error
+	var diffMessage string
+	common := -1
```
pkosiec, line 269: Same here:
```suggestion
	var common int
```

pkosiec (commented):
Hey @wangyangjun, just two quick comments as I missed the recent change to have `-1` as the initial common block count. Appreciate your patience!

I could apply the suggestion by myself and merge this, but I'd like to double check with you whether you spot some issues with that, as you changed it to -1?

test/e2e/discord_driver_test.go
wangyangjun: @pkosiec actually the original code is what you suggested. The reason I changed this in my last commit is there is only one line in the message of `AttachmentAssertion`. If there is any difference, the common will be always be `0`, then no diff will be print out. What do you think?

wangyangjun: If we don't want print diff there is only one message, feel free to apply the suggestions and merge the PR.

pkosiec: Thank you @wangyangjun, very good point. I think I found a solution - check the message limits. See this commit and tell me if that looks OK for you 👍 
https://github.com/kubeshop/botkube/pull/811/commits/e5300cf02db8cb7040d7c364574fd488fa623786

wangyangjun: @pkosiec Yes, it looks good to me.

pkosiec: Cool, I've just pushed the same changes for Discord because I forgot to do it.. Enabling auto-merge then!

PR #812, opened 2022-10-13
Add opt to append filter when building the `kubectl` command

mszostok:
## Description

- Add opt to append filter when building the `kubectl` command
- Reduce the Slack message size

## Demo

https://user-images.githubusercontent.com/17568639/195589745-aa441d0c-ce97-4278-8ac3-03b6559ebd5b.mov

## Related issue(s)

#748

### DIFF
--- a/pkg/bot/interactive/message.go
+++ b/pkg/bot/interactive/message.go
@@ -38,7 +38,7 @@ type Message struct {
 	Type MessageType
 	Base
 	Sections          []Section
-	PlaintextInputs   PlaintextInputs
+	PlaintextInputs   LabelInputs
 	OnlyVisibleForYou bool
 	ReplaceOriginal   bool
 }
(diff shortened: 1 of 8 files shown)


mszostok:
The filter option was updated according to the Lacho feedback:
- render it under the command preview (above the "run" button)
- render it only if preview is visible

New look and feel:
Screen Shot 2022-10-13 at 14 45 31

pkg/bot/interactive/message.go
```diff
@@ -128,23 +119,23 @@ type Selects struct {
...
+	DispatchInputActionOnEnter     DispatchedInputAction = "on_enter_pressed"
+	DispatchInputActionOnCharacter DispatchedInputAction = "on_character_entered"
+)
 
-// InputLabel refers to label of input element
-type InputLabel struct {
-	Type PlainTextType
-	Text string
+// PlaintextInput is used to create input elements to use in slack messages.
+type PlaintextInput struct {
```
ezodude, line 133: Would this make sense as a `LabelInput`? _A label will always be text based (similar to HTML label tags)_ 

This is just a general suggestion, not a blocker. But it could improve readability.

pkg/bot/interactive/message.go
```diff
@@ -128,23 +119,23 @@ type Selects struct {
...
+	// ID allows to carry the command that this input relates to.
+	ID               string
+	DispatchedAction DispatchedInputAction
+	Label            string
```
ezodude, line 137: If the Type is `LabelInput`, then `Label` -> `Value` or `Text`.

ezodude (approved):
👍 tested locally.

Left non blocker comment.

pkg/bot/interactive/message.go
mszostok: Thanks, suggestion applied 👍

PR #817, opened 2022-10-17
Add namespace label to default keyword, change filter placeholder

mszostok:
## Description

This PR address the Lacho feedback for the current implementation:

- Add namespace label to the `default` keyword (ONLY to it, rest should be without it):
Screen Shot 2022-10-17 at 11 27 03
Screen Shot 2022-10-17 at 11 27 37

- Change filter placeholder
Screen Shot 2022-10-17 at 11 28 33

## Related issue(s)

#748

### DIFF
--- a/pkg/execute/kubectl_cmd_builder_msg.go
+++ b/pkg/execute/kubectl_cmd_builder_msg.go
@@ -15,26 +15,27 @@ type (
 	// KubectlCmdBuilderOption defines option mutator signature.
 	KubectlCmdBuilderOption func(options *KubectlCmdBuilderOptions)
 
-	// KVItem describes the data for the dropdown item.
-	KVItem struct {
+	// dropdownItem describes the data for the dropdown item.
+	dropdownItem struct {
 		Name  string
 		Value string
 	}
 )
 
-// NewKV returns the KVItem instance.
-func NewKV(key, value string) KVItem {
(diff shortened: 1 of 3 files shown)


pkg/execute/kubectl_cmd_builder_msg.go
```diff
@@ -14,8 +14,31 @@ type (
 	}
 	// KubectlCmdBuilderOption defines option mutator signature.
 	KubectlCmdBuilderOption func(options *KubectlCmdBuilderOptions)
+
+	// KVItem describes the data for the dropdown item.
+	KVItem struct {
```
huseyinbabal, line 19: Just a minor thing, `KVItem` seems general  purpose KV storage data structure but it seems this is specifically developed for dropdown items. It is also opened to other packages. Maybe we could use more specific name like `DropdownItem`, just an idea :)

huseyinbabal (approved):
Added one minor comment, works like a charm 🚀 
Screen Shot 2022-10-18 at 11 47 22

pkg/execute/kubectl_cmd_builder_msg.go
mszostok: good point 👍  I will change to `dropdownItem` so it will be more specific and private

PR #818, opened 2022-10-17
Use real command guard and cached client, fix bug with orphan resource name

mszostok:
## Description

- Use real command guard 
- Use cached K8s client - otherwise the interactive kc builder doesn't work as we are throttled by the K8s and we cannot return the response within 3 sec which is a req from Slack server,
- Fix bug with orphan resource name, see https://user-images.githubusercontent.com/7155799/195316607-3a55f1a3-8d0a-4b0f-9299-e5420c26593a.mp4
- In case of dropdown error, just don't print them
- In case of rendering the command preview, return:
   Screen Shot 2022-10-17 at 14 23 56

## Related issue(s)

#748

### DIFF
--- a/pkg/execute/kubectl/guard.go
+++ b/pkg/execute/kubectl/guard.go
@@ -40,7 +40,8 @@ var (
 
 	// additionalResourceVerbs contains map of per-resource verbs which are not returned by K8s API, but should be supported.
 	additionalResourceVerbs = map[string][]string{
-		"nodes": {"cordon", "uncordon", "drain"},
+		"nodes": {"cordon", "uncordon", "drain", "top"},
+		"pods":  {"top"},
 	}
 
 	// additionalResourcelessVerbs contains map of per-resource verbs which are not returned by K8s API, but should be supported.
@@ -84,6 +85,7 @@ var (
 		"scale":        {},
 		"wait":         {},
 		"proxy":        {},
+		"run":          {},
 	}
 )
 
--- a/pkg/execute/kubectl/guard.go
+++ b/pkg/execute/kubectl/guard.go
@@ -135,6 +135,11 @@ func (g *CommandGuard) GetAllowedResourcesForVerb(verb string, allConfiguredReso
 
 // GetResourceDetails returns a Resource struct for a given resource type and verb.
 func (g *CommandGuard) GetResourceDetails(selectedVerb, resourceType string) (Resource, error) {
+	_, found := resourcelessVerbs[selectedVerb]
+	if found {
+		return Resource{}, nil
+	}
+
 	resMap, err := g.GetServerResourceMap()
 	if err != nil {
 		return Resource{}, err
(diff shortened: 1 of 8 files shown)


pkg/execute/kubectl/guard.go
```diff
@@ -135,6 +135,11 @@ func (g *CommandGuard) GetAllowedResourcesForVerb(verb string, allConfiguredReso
 
 // GetResourceDetails returns a Resource struct for a given resource type and verb.
 func (g *CommandGuard) GetResourceDetails(selectedVerb, resourceType string) (Resource, error) {
+	_, found := resourcelessVerbs[selectedVerb]
```
mszostok, line 138: When we know that a given verb is just resourceless instead of error we can return empty data with no error. Alternatively, I can ignore in kubectl builder logic the `ErrResourceNotFound` error. Both works for me, but from the logic perspective I found this more suitable.

pkg/execute/kubectl_cmd_builder.go
```diff
@@ -280,27 +281,40 @@ func (e *KubectlCmdBuilder) tryToGetResourceNamesSelect(botName string, bindings
...
 func (e *KubectlCmdBuilder) tryToGetNamespaceSelect(ctx context.Context, botName string, bindings []string, details stateDetails) *interactive.Select {
-	resourceDetails := e.commandGuard.GetResourceDetails(details.verb, details.resourceType)
+	log := e.log.WithFields(logrus.Fields{
+		"state":    details,
+		"bindings": bindings,
+	})
+
+	resourceDetails, err := e.commandGuard.GetResourceDetails(details.verb, details.resourceType)
+	if err != nil {
+		log.WithField("error", err.Error()).Error("Cannot fetch resource details, ignoring namespace dropdown...")
```
pkosiec, line 304: Is that proper behavior? Shouldn't we print the "internal error" message? 🤔 Looking at the guard logic once again, we can have `ErrResourceNotFound` error here which is rather impossible, and other error from K8s API. 🤔

EDIT: On the other hand even if we won't print the selectbox, the command for namespaced resource will be still valid, but executed in the  default namespace for a given channel, so...

Alright, let's keep it 👍

pkg/execute/kubectl/guard.go
pkosiec: Sounds good to me. Please just add a unit test for this test case (this will be super fast as there are table tests already) to make sure we won't accidentally break the contract 👍

pkg/execute/kubectl_cmd_builder.go
mszostok: Yep, I try to print as much as I can without printing the actual error, so great that I can keep it 👍

pkg/execute/kubectl/guard.go
```diff
@@ -135,6 +135,11 @@ func (g *CommandGuard) GetAllowedResourcesForVerb(verb string, allConfiguredReso
```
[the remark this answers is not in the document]
mszostok: I left the top, but just added it as an additional verb for pods and nodes

PR #820, opened 2022-10-17
Response thread fix

### LINKED ISSUE #816: Interactive filtering in thread does not show filter results in thread
## Description

When using interactive `kubectl` with filtering in Slack. Botkube replies with the results and starts a thread to allow for running other filters.

When using this thread to run the filter, filtered results appear in the main section of the channel and NOT in the thread. This is confusing because in the first instance you think filtering is not working.

See this screenshot:

filter_thread_bug

## Expected behavior

Further filtered messages should appear in the thread.

## Actual behavior

Further filtered messages are appearing the channel's main section.

## Steps to reproduce

- Set up Botkube in Slack.
- Run `@Butkube kc` to start interactive `kubectl`.
- Add a filter in the filter input box.
- Enter the started thread.
- Add another filter text and hit enter.
- The filter result will show in the main channel section instead of the thread.

huseyinbabal:
## Description

- File upload thread reply problem fixed

Screen Shot 2022-10-17 at 16 32 37

## Related issue(s)

Fixes #816

### DIFF
--- a/pkg/bot/slack.go
+++ b/pkg/bot/slack.go
@@ -269,7 +269,7 @@ func (b *Slack) send(msg slackMessage, req string, resp interactive.Message, onl
 
 	// Upload message as a file if too long
 	if len(markdown) >= slackMaxMessageSize {
-		_, err := uploadFileToSlack(msg.Channel, resp, b.client)
+		_, err := uploadFileToSlack(msg.Channel, resp, b.client, msg.ThreadTimeStamp)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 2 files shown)


mszostok (approved):
In general the changes LGTM.
And it works as "expected", so:
- when the messages is longer than X the filter input is attached
- if the message is to long and is sent as file then filter input is sent in thread

However when I was playing with it, it's not so easy to follow as sometimes the filter is sent and sometimes now (based on the rules that I mentioned above). For end-user it can be strange. Please see the recording.

Maybe when we sent the **filter input** in the thread, we should send it just once? and just send the filter output in it?


https://user-images.githubusercontent.com/17568639/196196650-9f225f3f-65d0-4c2c-bcfe-11a45e5c42d6.mov

But it's not sth that blocks this PR from being merged so I'm approving it 👍

huseyinbabal:
> In general the changes LGTM. And it works as "expected", so:
> 
> * when the messages is longer than X the filter input is attached
> * if the message is to long and is sent as file then filter input is sent in thread
> 
> However when I was playing with it, it's not so easy to follow as sometimes the filter is sent and sometimes now (based on the rules that I mentioned above). For end-user it can be strange. Please see the recording.
> 
> Maybe when we sent the **filter input** in the thread, we should send it just once? and just send the filter output in it?
> 
>  Screen.Recording.2022-10-17.at.15.53.28.mov 
> But it's not sth that blocks this PR from being merged so I'm approving it 👍

Current "reply in thread for filter output" is just a workaround due to slack limitation. I agree with you on revisiting this later.

PR #803, opened 2022-10-10
Make event notifications actionable

pkosiec:
## Description

Changes proposed in this pull request:

- Rework long/short notifications to Block kit for socket Slack (do not use deprecated API anymore)
- Implement Command Guard which facilitate getting commands for a given resource and verb
- Make notifications actionable
- Add empty command response message
- Add `ingresses` to default configuration of executor

## To do

On this PR:
- [x] Unit tests
- [x] Testing instruction

In follow-up PRs:
- [ ] Delete confirmation implementation
- [ ] Telemetry
- [ ] Pass real CommandGuard to kubectl survey
- [ ] Documentation

## Testing

Check out this PR and run Botkube locally:

```bash
export KUBECONFIG=...

export BOTKUBE_SETTINGS_LOG_LEVEL=info
export BOTKUBE_SETTINGS_KUBECONFIG=$KUBECONFIG
export BOTKUBE_CONFIG_PATHS="$(pwd)/comm_config.yaml"
go run cmd/botkube/main.go
```

Use the following values (NOTE: test both `short` and `long` notifications):

<details>
<summary>Show</summary>

```yaml

communications:
  'default-group':
    # Settings for Slack
    socketSlack:
      enabled: true
      appToken: "xapp-..."
      botToken: "xoxb-..."
      notification:
        type: "short" # TODO: Change to `short` during testing
      channels:
        'default':
          name: botkube-demo
          bindings:
            executors:
              - kubectl-backend
            sources:
              - k8s-all-events
              - k8s-recommendation-events

settings:
  clusterName: gke-playground
  configWatcher: true
  lifecycleServer:
    deployment:
      name: botkube
      namespace: botkube
    port: "2113"

configWatcher:
  enabled: false
  initialSyncTimeout: 0

executors:
  'kubectl-backend':
    kubectl:
      namespaces:
        include:
          - ".*"
      enabled: true
      commands:
        verbs: [ "run", "api-resources", "api-versions", "cluster-info", "describe", "diff", "explain", "get", "logs", "top", "auth", "delete" ]
        resources: [ "deployments", "pods", "namespaces", "daemonsets", "ingresses", "statefulsets", "storageclasses", "nodes", "configmaps", "services" ]
      defaultNamespace: backend
      restrictAccess: false

rbac:
  create: true
  rules:
    - apiGroups: ["*"]
      resources: ["*"]
      verbs: ["get", "watch", "list"]
    - apiGroups: [""]
      resources: ["pods"]
[...]

PR #822, opened 2022-10-18
Prevent resource deletion from the actionable notifications

pkosiec:
## Description

- Prevent resource deletion from the actionable notifications
    - This is the minimal acceptable option to avoid accidental resource deletion. As this is a rare edge case, it will be followed-up according to the community demand in https://github.com/kubeshop/botkube/issues/824.
- Unify documentation placeholder style (use `{ }` instead of `< >`
- Fix Helm chart icon and set Helm chart version on main to dev
    - The release pipeline will change it anyway, but at least it won't look odd on `main` that we have an outdated `0.13` chart. 

## Testing

Follow the same instructions from https://github.com/kubeshop/botkube/pull/803.
You won't see the `delete` option for any event notification anymore.

## Related issue(s)

Resolves https://github.com/kubeshop/botkube/issues/786

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -41,10 +41,10 @@ This section describes how to build and run Botkube from source code.
 
         ```sh
         IMAGE_PLATFORM=linux/arm64 make container-image-single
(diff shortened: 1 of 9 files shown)


huseyinbabal (approved):
Works like a charm 🚀 

Screen Shot 2022-10-18 at 18 13 03

PR #830, opened 2022-10-26
Configure action automation for a selected incoming event

### LINKED ISSUE #828: Configure action automation for a selected incoming event
## Overview

[Botkube automation](#789) allows users to execute actions (kubectl commands) based on incoming events. This requires users to add a target command to a specific event in a source.

This issue updates configuration in `values.yaml` to enable this behaviour.

## Acceptance Criteria

- Configuration settings that allow event -> action automation set up.
- Highlight the use of templates in the configured automation actions, to act on dynamic event data.
- Investigate whether the action automation should target multiple events at once to avoid repetition.

## Reason

Part of #789

ezodude:
## Description

This PR Introduces [the following configuration](https://github.com/kubeshop/botkube/blob/d1177bcf4754296876ab6f64e1980105d2f75940/helm/botkube/values.yaml#L53) to power action automations.

This is now implemented, but can be updated based on new comments.

Here are the assumptions (and some questions) on how this will work. Comments are welcome.

#### Definition

- Define your automation actions once.
- Actions can be triggered by different source types, e.g. Kubernetes, CloudEvents web hook, etc..
- Actions can target different executors, e.g. Kubectl,Helm, etc..
- Actions can be enabled/disabled.
- Actions can run templated and multi-line commands.

#### Execution

- An action is triggered by the aggregate values of resources, events, namespaces, name, annotations attached to the actions. 
- Actions can only run if there's an executor available with viable permissions to allow them to run.
- Multiple actions can run for a single event.
- Not all actions can rely on k8s events to trigger messages on a channel, e.g. some triggered actions might be informational.

### Questions to investigate (some for the actual implementation)

- What happens in case of execution errors?
- **To confirm**: the automation will target the sources of the parent event(s).
- **To confirm**: unauthorised namespaces used in the command will throw an error.

## Related issue(s)

Resolves #828


helm/botkube/values.yaml
```diff
@@ -50,6 +50,24 @@ kubeconfig:
...
   existingSecret: ""
 
+## Format: actions.{alias}
+actions:
+  ## kubectl based action
+  'show-created-resource':
+    ## Kubectl action configuration - requires that a kubectl executor with correct permission is available
+    kubectl:
+      enabled: true
+      run: "kubectl get {{.Event.ResourceType}} -n {{.Event.Namespace}} {{.Event.Name}}"
```
ezodude, line 60: Another suggestion: Instead of wiring the actions per source, @huseyinbabal asked why not add top level bindings config.

Taking his idea and re-working it slightly, we can add the attributes below to the `action` definition:
```YAML
  resources: ["deployments", "pods", "namespaces", "services", "ingresses"]
  events: ["create"]
  sources: [".*"]
```

Now there is no need to add `action` to specific sources. Giving us one place to view and configure how automations affect target resource/event/source combos.

huseyinbabal: Yep, my idea was totally a remake of prometheus configs. So prometheus has `routes` and `receivers`. Based on the criteria matching in `route` section, the command can be executed as defined in the `receivers` part. Example 👇 
```
  config:
    global:
      resolve_timeout: 5m
    route:
      routes:
      - match:
          event_type: create
        receiver: show-created-resource
    receivers:
    - name: 'show-created-resource'
      kubectl:
        enabled: true
        run: "kubectl get {{.Event.ResourceType}} -n {{.Event.Namespace}} {{.Event.Name}}"
``

ezodude: I reworked the `values.yaml` based on the discussion here.

ezodude:
@pkosiec the PR config setup has been updated as per our discussion.

pkosiec (approved):
LGTM, but please let's wait for @huseyinbabal and @mszostok (optional - if you have time) opinions. What do you think Guys?

mszostok:
I'm ok with the action syntax 👍 it solves the initial goal of the automation. Later based on the user feedback we can iterate on it but for now I don't see a reason for more complex settings. LGTM 👍

huseyinbabal (approved):

Issue #832, opened 2022-10-31
Run action automations for a selected filtered incoming event


ezodude:
## Overview

Add a mechanism to run automation actions on events using:
- A namespace filter: filters by namespace.
- A name filter: filters by event name
- An annotations filter: filters by a specific annotation configured on the event. 

## Acceptance Criteria

- Update sources config to allow the filters mentioned above.

## Reason

Part of #789 and [this comment](https://github.com/kubeshop/botkube/issues/789#issuecomment-1295160259).

Issue #831, opened 2022-10-31
Implement base action automation for a selected incoming event


ezodude:
## Overview

Implement the features enabled by setting the configuration provided by https://github.com/kubeshop/botkube/issues/828.

Investigate how this can work with the existing sources router. That is, actions added to source / resource / event combo based on actions automation config to easily trigger an action.

## Acceptance Criteria

- Configured automation is executed when a related k8s resource event is triggered.
- Can run automations with templated commands.
- A notification is sent to the appropriate channel when an automated action occurs (ideally through the k8s event system and not only through Botkube, so that the event is logged in the k8s infrastructure)
- Add any necessary unit tests.
- Add any necessary e2e tests.

## Phases

- [x] Execute actions when events are coming from bound source
- [x] Execute actions even when users don't want to see the action outcome on channel
- [x] Add E2E tests

Later, if needed:
- [x] Emit K8s event when a given command is executed - Update: Confirmed with @brampling that it is not needed

## Reason

Part of #789

PR #837, opened 2022-11-08
Wire execution of automated actions

pkosiec:
## Description

- Wire execution of automated actions
    - Action is executed only once, and the response is sent to the channels that subscribed for given source bindings. 
    - Interactive filtering is supported - try `kubectl logs` action with a big amount of logs, or command which returns more than 15 lines.
- Update automation config:
    - Add bindings for automated actions in the configuration
    - Resource name and annotation filters will be supported directly for each source
- Fix not supported source bindings for MS Teams

## Follow-up PRs

This PR contains base changes which will be followed-up with separate PRs:
- [ ] Load actions configuration and run actions based on it (even if no channel have such sources bound)
- [ ] Support resource name, annotation filtering for source configuration - #832
- [ ] Add E2E tests
- [ ] Add documentation

## Testing

Tested: Socket Slack, Discord, Teams, Mattermost

Check out this PR (`gh pr checkout 837`)
Uncomment lines from `pkg/action/provider.go`:
```go
	//cmd := "kubectl get cm -A"
	//actions = append(actions, events.Action{
	//	DisplayName: "Sample",
	//	Command:          fmt.Sprintf("%s %s", universalBotNamePlaceholder, cmd),
	//	ExecutorBindings: []string{"kubectl-read-only"},
	//})
```

Ensure you have the default executor enabled for running Botkube locally:
```yaml
executors:
  'kubectl-read-only':
    kubectl:
      namespaces:
        # -- List of allowed Kubernetes Namespaces for command execution.
        # It can also contain a regex expressions:
        #  `- ".*"` - to specify all Namespaces.
        include:
          - ".*"
        # -- List of ignored Kubernetes Namespace.
        # It can also contain a regex expressions:
        #  `- "test-.*"` - to specify all Namespaces.
        exclude: []
      # -- If true, enables `kubectl` commands execution.
      enabled: true
      ## List of allowed `kubectl` commands.
      commands:
        # -- Configures which `kubectl` methods are allowed.
        verbs: ["api-resources", "api-versions", "cluster-info", "describe", "explain", "get", "logs", "top"]
        # -- Configures which K8s resource are allowed.
        resources: ["deployments", "pods", "namespaces", "daemonsets", "statefulsets", "storageclasses", "nodes", "configmaps", "services", "ingresses"]
      # -- Configures the default Namespace for executing Botkube `kubectl` commands. If not set, uses the 'default'.
      defaultNamespace: default
      # -- If true, enables commands execution from configured channel only.
      restrictAccess: false
``` 

Run Botkube:

```
export BOTKUBE_SETTINGS_KUBECONFIG=$KUBECONFIG
export BOTKUBE_CONFIG_PATHS="$(pwd)/resource_config.yaml,$(pwd)/comm_config.yaml"
go run ./cmd/botkube/main.go
```

Trigger any event from command line, such as:

```
kubectl delete po -n kube-system -l k8s-app=metrics-server
```

See outputs:
image

Try out different commands to see that all different filtering options are supported (e.g. logs).

## Related issue(s)

https://github.com/kubeshop/botkube/issues/831


pkg/bot/discord.go
```diff
@@ -253,22 +276,21 @@ func (b *Discord) handleMessage(ctx context.Context, dm discordMessage) error {
...
 
-func (b *Discord) send(event *discordgo.MessageCreate, req string, resp interactive.Message) error {
-	b.log.Debugf("Discord incoming Request: %s", req)
+func (b *Discord) send(channelID string, resp interactive.Message) error {
 	b.log.Debugf("Discord Response: %s", resp)
 
 	markdown := interactive.RenderMessage(b.mdFormatter, resp)
 
 	if len(markdown) == 0 {
-		return fmt.Errorf("while reading Slack response: empty response for request %q", req)
```
huseyinbabal, line 271: So, we used to know which request caused an empty response, now we don't have it. It would be valuable for debug purposes, do we have it in other places and remove it here, or completely lost it?

pkosiec: The problem is that in case of automation we don't know the actual request. As I use the `send` method for it as well, I moved the debug logging for the request in the `handleMessage` method:

```go
	b.log.Debugf("Slack incoming Request: %s", request)
```

but yeah, with such approach don't have the context for a given error. 
I can bring it back though - I'll still pass `request` to the `send` method, and in case of automation I will set it to e.g.  `Automation`. WDYT?

huseyinbabal (approved):
LGTM with very minor comment. Btw, automatic action is very important to compete with alternative products, well done one more time  🥇

pkg/bot/discord.go
huseyinbabal: > The problem is that in case of automation we don't know the actual request. As I use the `send` method for it as well, I moved the debug logging for the request in the `handleMessage` method:
> 
> ```go
> 	b.log.Debugf("Slack incoming Request: %s", request)
> ```
> 
> but yeah, with such approach don't have the context for a given error. I can bring it back though - I'll still pass `request` to the `send` method, and in case of automation I will set it to e.g. `Automation`. WDYT?

Now it is clear, I think we can keep what you did 🆗 Thanks

josefkarasek (approved):
lgtm

PR #845, opened 2022-11-08
Validate bindings for communcators

### LINKED ISSUE #760: Validate bindings for a given communicator
## Description

Currently, we allow specifying unknown bindings under:
```yaml
communications:
 'default':
   {communicator}:
     channels:
       'channel':
         bindings:
           sources:
             - {sourceGroupName}  # here
             - ...
           executors:
             - {executorGroupName}  # here
             - ...

```
However, we should validate whether they exist under:
```yaml
sources:
 {sourceGroupName}: {}

executors:
 {executorGroupName}: {}
```

## Expected behavior

BotKube is not started with invalid configuration.

## Actual behavior

BotKube is started with invalid configuration but sources and executor doesn't work, and user don't have any feedback message why it's not working.

## Notes

Add dedicated validator under: https://github.com/kubeshop/botkube/blob/main/pkg/config/validator.go

josefkarasek:

* Check that bindings mentioned in Communicators exist

## Related issue(s)

Resolves #760

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -413,7 +413,7 @@ type Communications struct {
 // Slack configuration to authentication and send notifications
 type Slack struct {
 	Enabled      bool                                   `yaml:"enabled"`
-	Channels     IdentifiableMap[ChannelBindingsByName] `yaml:"channels"  validate:"required_if=Enabled true,omitempty,min=1"`
+	Channels     IdentifiableMap[ChannelBindingsByName] `yaml:"channels"  validate:"required_if=Enabled true,dive,omitempty,min=1"`
 	Notification Notification                           `yaml:"notification,omitempty"`
 	Token        string                                 `yaml:"token,omitempty"`
 }
@@ -421,7 +421,7 @@ type Slack struct {
 // SocketSlack configuration to authentication and send notifications
 type SocketSlack struct {
 	Enabled      bool                                   `yaml:"enabled"`
-	Channels     IdentifiableMap[ChannelBindingsByName] `yaml:"channels"  validate:"required_if=Enabled true,omitempty,min=1"`
+	Channels     IdentifiableMap[ChannelBindingsByName] `yaml:"channels"  validate:"required_if=Enabled true,dive,omitempty,min=1"`
 	Notification Notification                           `yaml:"notification,omitempty"`
 	BotToken     string                                 `yaml:"botToken,omitempty"`
 	AppToken     string                                 `yaml:"appToken,omitempty"`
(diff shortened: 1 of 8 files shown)


pkg/config/validator.go
```diff
@@ -99,6 +103,16 @@ func registerNamespaceValidator(validate *validator.Validate, trans ut.Translato
 	return validate.RegisterTranslation(nsIncludeTag, trans, registerFn, translateFunc)
 }
 
+func registerBindingsValidator(validate *validator.Validate, trans ut.Translator) error {
+	validate.RegisterStructValidation(bindingsStructValidator, Config{})
+
+	registerFn := func(ut ut.Translator) error {
+		return ut.Add(invalidBindingTag, "{0} {1}", false)
```
josefkarasek, line 110: not sure if this translator is needed

pkg/config/validator.go
```diff
@@ -167,6 +181,84 @@ func namespacesStructValidator(sl validator.StructLevel) {
 	}
 }
 
+func bindingsStructValidator(sl validator.StructLevel) {
+	conf, ok := sl.Current().Interface().(Config)
+	if !ok {
+		return
+	}
+	for _, comm := range conf.Communications {
+		// slack
```
josefkarasek, line 190: any idea how to make these checks more generic?

pkg/config/validator.go
```diff
@@ -14,9 +14,10 @@ import (
 )
```
pkosiec, line 14: I've just noticed a problem - please modify `Teams` in `config.go` and add `required_if=Enabled true` for BotBindings, similarly like with others. Thanks!

```go
// Teams creds for authentication with MS Teams
type Teams struct {
        // (...)
	Bindings     BotBindings  `yaml:"required_if=Enabled true bindings"`

}

```

pkosiec (commented):
A few comments 🙂

pkg/config/config.go
```diff
@@ -485,16 +485,15 @@ type Discord struct {
...
-	Channels     IdentifiableMap[ChannelBindingsByID] `yaml:"channels"  validate:"required_if=Enabled true,omitempty,min=1"`
+	Channels     IdentifiableMap[ChannelBindingsByID] `yaml:"channels"  validate:"required_if=Enabled true,dive,omitempty,min=1"`
 	Notification Notification                         `yaml:"notification,omitempty"`
 }
 
 // Webhook configuration to send notifications
 type Webhook struct {
-	Enabled bool   `yaml:"enabled"`
-	URL     string `yaml:"url"`
-	// TODO: not used yet.
```
josefkarasek, line 496: @pkosiec should the comment be still there?

pkg/config/testdata/TestLoadedConfigValidationErrors/app-token-only.yaml
```diff
@@ -12,3 +12,7 @@ communications: # req 1 elm.
               - k8s-events
 
       appToken: 'xapp-SLACK_API_TOKEN'
+executors:
+  kubectl-read-only:
```
pkosiec, line 16: IMO this would look nicer 😄 

```
kubectl-read-only: {}
```

If you agree, please change it in all places. Thanks!

pkosiec (approved):
LGTM - please see one small comment before merge though.

Issue #703, opened 2022-08-26
Unify BotKube commands naming


pkosiec:
## Overview

Currently, BotKube has the following list of commands:

```
BotKube Help

Usage:
    @BotKube ping [--cluster-name <cluster-name>]
    @BotKube commands list  (on BotKube v0.11.0+)
    @BotKube <kubectl command without kubectl prefix> [--cluster-name <cluster_name>]
    @BotKube notifier [stop|start|status|showconfig]
    @BotKube filters [list|enable|disable] [filter-name]

Description:

Cluster Status:
    - List all available Kubernetes Clusters and check connection health.
    - If flag specified, gives response from the specified cluster.

    Example:
        @BotKube ping
        @BotKube ping --cluster-name mycluster

Allowed kubectl commands list:
    - Show allowed kubectl commands as per the BotKube configuration

    Example:
        @BotKube commands list

Kubectl commands:
    - Executes kubectl commands on k8s cluster and returns output.

    Example:
        @BotKube get pods
        @BotKube logs podname -n namespace
        @BotKube get deployment --cluster-name cluster_name

    Allowed kubectl commands:
        For BotKube v0.11.0+:
            Check allowed commands with '@BotKube commands list'
        For older versions:
            Allowed commands are - get, top, cluster-info, describe, explain, logs, version, auth, api-resources, api-versions, diff

Notifier commands:
    - Commands to manage incoming notifications (Runs only on configured channel).

    Example:
        @BotKube notifier stop          Stop sending k8s event notifications to Slack
        @BotKube notifier start         Start sending k8s event notifications to Slack
        @BotKube notifier status        Show running status of event notifier
        @BotKube notifier showconfig    Show BotKube configuration for event notifier

Filters commands:
    - Command to manage filters run on K8s events (Runs only on configured channel).

    Example:
        @BotKube filters list                     Show list of available filters
        @BotKube filters disable ImageTagChecker  Disable filter
        @BotKube filters enable ImageTagChecker   Enable filter

Options:
    --cluster-name                  Get cluster specific response
```

You can see, that for example `@BotKube notifier showconfig` doesn't follow the pattern for other commands. This is something which would be great to fix.

## Acceptance Criteria

- [x] Go through the list of commands and prepare final naming convention accepted by the team @lpetkov
    - For example, `@BotKube commands list` and `@BotKube notifier showconfig` - maybe it should be just one command, e.g. `@BotKube config`?
[...]

PR #852, opened 2022-11-15
Add support for list/enable/disable actions

### LINKED ISSUE #833: Enable/Disable configured automation actions through chat commands
## Overview

Configured automation actions should be enabled/disabled dynamically through chat commands. E.g. 
- @Botkube list (automations or actions).
- @Botkube enable (automation or action).
- @Botkube disable (automation or action).

## Acceptance Criteria

- Automations can be listed, enabled/disabled dynamically through chat commands for all messaging platforms.
- Once disabled/enabled, ensure the updated configuration persists beyond Botkube restarts.
- Once disabled, ensure an automation can never run.
- Add e2e tests to cover the chat command.

## Reason

Part of #789

josefkarasek:
## Add support for list/enable/disable actions

- Allow users to list, enable and disable actions via chat messages.
- Configuration modification is persisted in runtime config map

## Related issue(s)

Resolves #833


pkg/execute/actions.go
```diff
@@ -0,0 +1,56 @@
...
+	w := tabwriter.NewWriter(buf, 5, 0, 1, ' ', 0)
+	fmt.Fprintln(w, "ACTION\tENABLED")
+	for name, enabled := range a.listActions() {
+		fmt.Fprintf(w, "%s\t%v\n", name, enabled)
+	}
+	w.Flush()
+	return buf.String()
+}
+
+func (a *ActionManager) enableAction(name string) bool {
```
pkosiec, line 42: BTW, why do we need such methods `enableAction` and `disableAction`? We don't need to write the config to the `actions map[string]bool` map, as Botkube needs to be restarted to apply the changes.

So using `PersistActionEnabled` is totally sufficient.

Remember, that filters work a bit differently - we can enable/disable them in runtime without restarting app. With actions, we'll need to restart the whole app, so we don't need to change anything in runtime.

pkg/execute/executor.go
```diff
@@ -332,6 +333,61 @@ func (e *DefaultExecutor) runFilterCommand(ctx context.Context, args []string, c
 	return "", errUnsupportedCommand
 }
 
+func (e *DefaultExecutor) runActionCommand(ctx context.Context, args []string, clusterName string) (string, error) {
```
pkosiec, line 336: As stated in the GH issue, please use new form already:
`@Botkube enable/disable/list action/actions/act <name>` instead of `@Botkube actions enable/disable/list <name>`.

This will be later unified in other commands in the task #703.

BTW I'd recommend splitting the execution part to a separate executor - see the `notifier.go` file for example. In the DefaultExecutor it would be great to have general logic (parsing args, register all the executors, and trigger commands).

cmd/botkube/main.go
```diff
@@ -128,6 +128,9 @@ func run() error {
 	// Set up the filter engine
```
pkosiec, line 128: Please also cover the new functionality with some tests. I think some unit tests will be enough. It would be also awesome to test it as a part of E2E tests, but Botkube restart could be a bit problematic 🤔

pkosiec (commented):
An early fedback - didn't test it, but it looks very promising!

cmd/botkube/main.go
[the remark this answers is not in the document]
pkosiec: Yes, point no 2 is something already done, I was referring to point no 1. Right, you're saving the values once they are changed by user, but the data won't be properly merged during Helm upgrade. We'd like to allow users overwriting the saved data from commands if they want to do it. 
How do to it? You'd need to do iterate over all actions, store `enabled: true/false` for all of them in runtime-config during Helm chart installation/upgrade, and merge state beforehand as we do it with e.g. `$mergedRuntimeCommunications`.

So, when you do `helm install`, in the runtime CM we'd need to have:
```
actions:
      'foo':
        enabled: true
      'bar':
          enabled: false
```

Ping me if it's not clear.

pkg/execute/actions.go
```diff
@@ -0,0 +1,56 @@
...
+		a[k] = v.Enabled
+	}
+	return &ActionManager{a}
+}
+
+func (a *ActionManager) listActions() map[string]bool {
+	out := make(map[string]bool)
+	for k, v := range a.actions {
+		out[k] = v
+	}
```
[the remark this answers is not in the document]
josefkarasek: Sorted

pkg/execute/actions.go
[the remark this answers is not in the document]
josefkarasek: I ended up deleting these functions

josefkarasek: I ended up deleting these functions.

pkg/execute/executor.go
josefkarasek: I tried and the way these values are wired (`arg[1]`) it's very difficult to do this one-by-one. I suggest we do all the commands at the same time.

pkosiec: Fair enough 👍 Just please mention it in PR desc 👍 Thanks!

pkosiec:
@josefkarasek I've just realized that we should also edit help so that a new section is included. I think we could reuse "Manage incoming notifications" layout:

Slack Interactive:
image

Other communication platforms:
image

Don't worry, both of them as done automatically with different renderer. It's about copy-paste, so it'll be straightforward 🙂 
CC @lpetkov

pkg/execute/executor.go
pkosiec: @josefkarasek 
On second thought, I think we cannot postpone this, and we need to go with the new command form as we initially planned.
Why? There's a chance we won't complete #703 this release, and as a result, we would do a breaking change for the newly introduced command the next release. If we already know the target form (`enable/disable/list actions/Actions/action/act, then we should support it from day one. We did that the same with `edit SourceBindings` command. 
I looked at the code and I cannot see the difficulty with implementation of it. It shouldn't be a problem, as the `enable/disable/list` verbs are not yet used anywhere else, so there won't be any conflicts. You can see how the `edit sourcebindings` command was implemented:
https://github.com/kubeshop/botkube/blob/main/pkg/execute/executor.go#L207. As you'll dig deeper, we support any kind of casing for the `SourceBindings` name (we make the user input lowercase and we compare it with https://github.com/kubeshop/botkube/blob/main/pkg/execute/edit.go#L34). Let me know how I can help with this.

I won't push too hard on how it is implemented (we can make it more beautiful in #703), but please make sure we support `enable/disable/list actions/Actions <name>` already.  We can introduce aliases like `action` (singular) or `act` later in #703. Thank you!

cmd/botkube/main.go
pkosiec: In E2E tests, can we at least list actions (`list actions`)? That will be easy to test as the output is predictable 🤔
As I wrote, enabling/disabling could be a bit harder (we'd need to wait for restart, and that will result in Botkube "welcome" message on both channels), so for me enable/disable are nice to have. The only thing we need to consider is that everything we won't cover in automated tests we'll need to test manually.

test/e2e/bots_test.go
```diff
@@ -176,6 +176,23 @@ func runBotTest(t *testing.T,
 	err = waitForDeploymentReady(deployNsCli, appCfg.Deployment.Name, appCfg.Deployment.WaitTimeout)
 	require.NoError(t, err)
 
+	cmdHeader := func(command string) string {
+		return fmt.Sprintf("`%s` on `%s`", command, appCfg.ClusterName)
+	}
+
+	t.Run("List actions stored in runtime config map", func(t *testing.T) {
```
pkosiec, line 183: I'd remove a not necessary detail:
```suggestion
	t.Run("List actions", func(t *testing.T) {
```

pkg/execute/action.go
```diff
@@ -0,0 +1,88 @@
...
+	var cmdVerb = args[0]
+	defer func() {
+		cmdToReport := fmt.Sprintf("%s %s", args[1], cmdVerb)
+		e.reportCommand(cmdToReport, false)
+	}()
+
+	actions, err := e.cfgManager.ListActions(ctx)
+	if err != nil {
+		return "Failed to list actions", err
+	}
```
pkosiec, line 37: Actually a dedicated call to get ConfigMap data is not needed - we can read the actions directly from the config. We reload Botkube anytime the config changes, so we can assume it is always up to date (as it is the one loaded into the Botkube app) 👍 

This is the same as the edit executor: https://github.com/kubeshop/botkube/blob/main/pkg/execute/edit.go - it also takes a config, processes it and based on it it displays a modal with all sources configured.

pkg/execute/action.go
```diff
@@ -0,0 +1,88 @@
...
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"text/tabwriter"
+
+	"github.com/kubeshop/botkube/pkg/config"
+)
+
+func (e *DefaultExecutor) runActionCommand(ctx context.Context, args []string, clusterName string) (string, error) {
```
pkosiec, line 12: Great, that it is defined in a separate file, but can you also define a separate struct, same as the Notifier executor? https://github.com/kubeshop/botkube/blob/main/pkg/execute/notifier.go#L40

In that way you'll be able to pass the actions config there and read it when doing `actions list` 👍

pkg/config/manager.go
```diff
@@ -179,3 +180,41 @@ func (m *PersistenceManager) PersistFilterEnabled(ctx context.Context, name stri
...
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
+	// return results
+	actions := make(map[string]Action)
+	for _, name := range keys {
+		actions[name] = state.Actions[name]
+	}
+	return actions, nil
+}
```
pkosiec, line 220: As the previous comment said, this will be not needed, so you can remove it 👍

pkg/execute/action.go
```diff
@@ -0,0 +1,88 @@
...
+	validSubCmd := false
+	for _, cmdName := range []string{"action", "actions", "act"} {
+		if args[1] == cmdName {
+			validSubCmd = true
+			break
+		}
+	}
+	if !validSubCmd {
+		return fmt.Sprintf("'%s' is not a valid subcommand", args[1]), nil
+	}
```
pkosiec, line 26: 1. Didn't test it yet, but looks like `@Botkube list Actions` wouldn't work
2. Using map will be more efficient and also simpler

You can replace it with:

below imports define such variable:
```go
var actionResourceNames = map[string]struct{} {
	"action": {},
	"act": {},
	"actions": {},
}
```

replace the selected snippet with:

```suggestion
	cmdRes := strings.ToLower(args[1])
	_, ok := actionResourceNames[cmdRes]
	if !ok {
		return fmt.Sprintf("'%s' is not a valid resource", args[1]), nil
	}
```

pkg/execute/action.go
```diff
@@ -0,0 +1,88 @@
...
+
+	var cmdVerb = args[0]
+	defer func() {
+		cmdToReport := fmt.Sprintf("%s %s", args[1], cmdVerb)
```
pkosiec, line 30: It's inverted - if you apply the previous snippet, then you can use:
```suggestion
		cmdToReport := fmt.Sprintf("%s %s", cmdVerb, cmdRes)
```

pkosiec (commented):
Didn't test it yet, but here are a few small comments after review. Looks good overall, good work! 🚀

pkosiec (approved):
Tested, works well 👌 Good job!

Issue #838, opened 2022-11-08
Add plugin manager to Botkube


huseyinbabal:
## Overview

Plugin manager is responsible for handling plugin registration based on the configuration of each plugin. We need to  implement that in Botkube as described in AC

## Acceptance Criteria
-  Plugin manager can accept plugin configuration with name and version.

- Plugin manager can accept plugin configuration with name and version
- It should download the plugin executable by using the name and version to a specified local folder before running the Botkube application
- There should be a caching mechanism not to download same version again
- You can see a working example [here](https://github.com/huseyinbabal/botkube-plugins-playground/blob/main/plugin/manager.go)

mszostok:
- [ ] Add executor plugin manager with example executor
- [ ] Add source plugin manager with example source
- [ ] Add validation according to [Validating Plugin configuration](https://github.com/kubeshop/botkube/blob/main/docs/proposal/2022-11-22-botkube-plugins-integration.md#validating-plugin-configuration)
- [ ] Add logger for go-plugin client
   - We need to create a wrapper for [plugin binary output](https://github.com/hashicorp/go-plugin/blob/b60690f8999f79dae652a67e7b81a6923b43cff8/client.go#L161-L166) and the [plugin client](https://github.com/hashicorp/go-plugin/blob/b60690f8999f79dae652a67e7b81a6923b43cff8/client.go#L181-L183) itself
- [ ]

PR #857, opened 2022-11-21
Add plugin manager for sources

mszostok:
## Description

- Add plugin manager for sources
- Update contrib doc and provide an easy way to develop Botkube core plugins
- Support serving source via fake plugin HTTP server
- Add e2e tests

## Related issue(s)

- #838


test/fake/plugin_server.go
```diff
@@ -1,4 +1,4 @@
-package e2e
+package fake
```
mszostok, line 1: I cannot put the `fake` pkg inside `e2e` as then logs streaming for test doesn't work because of this issue: https://github.com/golang/go/issues/46959

CONTRIBUTING.md
```diff
@@ -130,6 +130,44 @@ For faster development, you can also build and run Botkube outside K8s cluster.
...
+
+1. Start fake plugins server to serve binaries from [`dist`](dist) folder:
+
+   ```bash
+   go run test/helpers/plugin_server.go
+   ```
+
+	 > **Note**
+	 > If Botkube runs inside the k3d cluster, export the `PLUGIN_SERVER_HOST=http://host.k3d.internal` environment variable.
+
```
mszostok, line 150:
```suggestion
   > **Note**
   > If Botkube runs inside the k3d cluster, export the `PLUGIN_SERVER_HOST=http://host.k3d.internal` environment variable.

```

cmd/source/cm-watcher/main.go
```diff
@@ -0,0 +1,106 @@
...
+	// Object holds information about object to watch.
+	Object struct {
+		Name      string          `yaml:"name"`
+		Namespace string          `yaml:"namespace"`
+		Event     watch.EventType `yaml:"event"`
+	}
+)
+
+// CMWatcher implements Botkube source plugin.
+type CMWatcher struct{}
```
josefkarasek, line 39: Code style consistency - `type` block above combined with single line declaration.

internal/source/source.go
```diff
@@ -0,0 +1,184 @@
...
+// Start starts all sources and dispatch received events.
+func (d *Dispatcher) Start(ctx context.Context) error {
+	// startProcesses holds information about started unique plugin processes
+	// We start a new plugin process each time we see a new order of source bindings.
+	// We do that because we pass the array of configs to each `Stream` method and
+	// the merging strategy for configs can depend on the order.
+	// As a result our key is e.g. ['source-name1;source-name2']
+	startProcesses := map[string]struct{}{}
+
+	startPlugin := func(bindSources []string) error {
```
josefkarasek, line 51: Isn't this function complex enough to be named / defined separately?

josefkarasek (approved):
lgtm :+1:

cmd/source/cm-watcher/main.go
mszostok: Actually, So I use the `type` block to "group" related entities, so the code block about groups types related to the configuration. This one is about the "service" that implements the source plugin functionality. 

Are you OK with such approach?

internal/source/source.go
mszostok: yes, you are right. I changed it to become a dedicated method on this struct 👍 

In the next PR, I will change it even more to make it more straight forward when it comes to the responsibilities. Because I see that it is to complex to maintain and test all those pieces together.

mszostok:
Hi @josefkarasek!

Thanks for your review 🙇 
I addressed the feedback and also merge the PR as it was a dependency for my second PR. You were also assigned there 🙂 so we will be able to follow up there with some changes that you will see as required. Thanks!

Issue #844, opened 2022-11-08
Botkube Plugin System


huseyinbabal:
This is the umbrella epic for [Event Extensibility](https://github.com/kubeshop/botkube/issues/650) and [Executor Extensibility ](https://github.com/kubeshop/botkube/issues/649)design epics. You can see the stories to implement a plugin system for Botkube to support both Source and Executor extensions.

### Design Documentation
https://github.com/kubeshop/botkube/blob/main/docs/proposal/2022-09-28-botkube-plugin-system.md

### Stories
 - [x] https://github.com/kubeshop/botkube/issues/838
 - [x] https://github.com/kubeshop/botkube/issues/843
 - [x] https://github.com/kubeshop/botkube/issues/839
 - [x] https://github.com/kubeshop/botkube/issues/925
 - [x] https://github.com/kubeshop/botkube/issues/901
 - [x] https://github.com/kubeshop/botkube/issues/899
 - [x] https://github.com/kubeshop/botkube/issues/870

 - [ ] https://github.com/kubeshop/botkube/issues/840
 - [x] https://github.com/kubeshop/botkube/issues/841

## Later
 - [ ] https://github.com/kubeshop/botkube/issues/842

 - [ ] https://github.com/kubeshop/botkube/issues/878

PR #859, opened 2022-11-22
Describe how plugins can be consumed within Botkube core

mszostok:
## Description

- Describe how plugins can be consumed within Botkube core

## Related issue(s)

- #844


docs/proposal/2022-11-22-botkube-plugins-integration.md
```diff
@@ -0,0 +1,498 @@
...
+      - https://github.com/kubeshop/botkube/releases/download/v0.17.0/source_kubernetes-linux-arm64
+```
+
+The index file can be generated automatically just by scanning directory where plugins binaries are stored.
+
+## Implementation details
+
+### E2E testing
+
+Botkube fetches the plugins index and binaries when it's started. However, if they are already present, Botkube doesn't take any action. We could leverage that and simply built all plugins binaries and include them in Docker image with in a proper directories.
```
mszostok, line 391: I tested that approach both locally and on CI. It works properly, see: https://github.com/mszostok/botkube/actions/runs/3523896348/jobs/5908734163

docs/proposal/2022-11-22-botkube-plugins-integration.md
```diff
@@ -0,0 +1,498 @@
...
+To enable a given plugin you need to specify it under the `executors` or `sources` property. In the first phase, we only allow to enable and disable plugins using Botkube configuration. Later we can introduce `@Botkube enable [executor|source] NAME` and `@Botkube disable [executor|source] NAME`. Disabling and enabling plugins should be possible in the Botkube runtime without restarting it.
+
+**Syntax:**
+
+```yaml
+executors:
+  'plugin-based':
+    botkube/kubectl:     # <repo>/<plugin> is syntax for plugin based executors
+      enabled: true      # if not enabled we don't download and start a given plugin
+      version: v1.0.0    # if empty, the latest version is used.
```
pkosiec, line 94: This version somehow doesn't fit here to me, as it is more about the plugin metadata, like `{repo}/{plugin-name}` 🤔 What do you think to have something like

```yaml
executors:
  'plugin-based':
    botkube/kubectl@v1.0.0:     # <repo>/<plugin> is syntax for plugin based executors
      enabled: true      # if not enabled we don't download and start a given plugin
      config: {}
```

?

docs/proposal/2022-11-22-botkube-plugins-integration.md
```diff
@@ -0,0 +1,498 @@
...
+     'plugin-based':
+       botkube/kubectl:
+         enabled: true
+         config:
+           namespaces:
+             include: ["botkube", "default","ambassador"]
+           commands:
+             verbs: ["get","logs","top"]
+             resources: ["pods","deployments","nodes","configmap"]
+       mszostok/kubectl:  # not allowed as it is conflicting with already registered 'kubectl' from the Botkube repository.
```
pkosiec, line 182: In this config I understand there is a conflict, but if I change the aliases I can have these two executors, right? If not, why?

If we listed all the options inside the repository index (`kubectl`, `kc`, `k`) then I would be able to override them for `mszostok/kubectl` - e.g. define `aliases: [ "mkubectl", "mk", "mkc" ]`. WDYT?

In that way options no 3 and 4 would be no longer valid. I think that would be a nice feature, and also very convenient for development/testing purposes.

docs/proposal/2022-11-22-botkube-plugins-integration.md
```diff
@@ -0,0 +1,498 @@
...
+
+Currently, the aliases are built-in into Botkube core logic. One option is to define them in the index file for a given executor:
+
+```yaml
+entries:
+  - name: "kubectl"
+    type: "executor"
+    description: "Kubectl executor plugin."
+    version: "v1.0.0"
+    aliases: [ "kc", "k" ]
```
pkosiec, line 122: Please see my previous comment (posted for line 182). Shouldn't we also list here `kubectl`? I think it would be better to list all the possibilities here and I don't see any obstacles to not treat all the 3 options the same way 🤔 WDYT?

docs/proposal/2022-11-22-botkube-plugins-integration.md
```diff
@@ -0,0 +1,498 @@
...
+- all enabled plugins binaries
+
+We use the `emptyDir` so the data is not removed as long as the Pod is not rescheduled to a different Node. To force download, you need to run:
+
+```bash
+kubectl exec -it $(kubectl get po -l app=botkube -n botkube -oname) -- rm -rf /tmp
+kubectl delete po -l app=botkube -n botkube
+```
+
+Later we can provide a dedicated `@Botkube` command to simplify refreshing downloaded plugins. However, in the happy path scenarios no one should replace already release binaries.
```
pkosiec, line 320:
```suggestion
Later we can provide a dedicated `@Botkube` command to simplify refreshing downloaded plugins. However, in the happy path scenarios no one should replace already released binaries.
```

Alternatively: "properly versioned binaries"?

docs/proposal/2022-11-22-botkube-plugins-integration.md
```diff
@@ -0,0 +1,498 @@
...
+
+Botkube uses the `runtime.GOOS` and `runtime.GOARCH` variables to determine on which system Botkube is running. It downloads respected binary by searching a proper link in plugin index file:
+
+```yaml
+entries:
+  - name: "kubectl"
+    type: "executor"
+    description: "Kubectl executor plugin."
+    version: "v1.0.0"
+    links:
```
pkosiec, line 339: BTW (minor) I like `urls` more - this is what Helm has 😛 Or `binaries`?

Ignore this comment, if you don't like both options 😄

docs/proposal/2022-11-22-botkube-plugins-integration.md
```diff
@@ -0,0 +1,498 @@
...
+### Supporting multi OS and architectures
+
+We want to run plugins on different Kubernetes distros and also locally for development purposes. As a result, we need to support different operating systems and different architectures.
+
+We enforce that plugin binaries have a suffix with a given pattern: `{os}_{arch}[.exe]`. For example:
+
+- `botkube_executor_helm_linux_amd64`
+- `botkube_source_prometheus_darwin_amd64`
+
+Botkube uses the `runtime.GOOS` and `runtime.GOARCH` variables to determine on which system Botkube is running. It downloads respected binary by searching a proper link in plugin index file:
```
pkosiec, line 331: > It downloads respected binary by searching a proper link in plugin index file

So it is based on the link itself - looking for specific pattern. Mixed feelings there - we won't be able to support some arbitrary links without such pattern.

WDYT about having it explicitly inside the repository like:

```yaml
entries:
  - name: "kubectl"
    type: "executor"
    description: "Kubectl executor plugin."
    version: "v1.0.0"
    urls: # or `binaries`?
      - url: https://github.com/kubeshop/botkube/releases/download/v0.17.0/executor_kubectl-darwin-amd64
        platform:
          architecture: amd64
          os: linux
     # (...)
```

? This is very similar what's in the OCI image spec: https://github.com/opencontainers/image-spec/blob/main/image-index.md#example-image-index

We can skip `platform` property though and flat it if you like it. But having an object for each `url`/`binary` will be more future-proof if we needed to extend it at some point.

Alternatively, we could have some map, like:

```yaml
entries:
  - name: "kubectl"
    type: "executor"
    description: "Kubectl executor plugin."
    version: "v1.0.0"
    binaries:
      darwin/amd64: 
          url: https://github.com/kubeshop/botkube/releases/download/v0.17.0/executor_kubectl-darwin-amd64
     linux/arm64: { url: "..." }
```

But I like the list more I think 🤔

docs/proposal/2022-11-22-botkube-plugins-integration.md
```diff
@@ -0,0 +1,498 @@
...
+Before downloading index file or plugin binaries, Botkube checks whether a given file already exist. If yes, no action is taken.
+
+### Do we want to decouple the interfaces between plugin Go implementation and gRPC?
+
+I don't have a strong opinion here. For now, I reuse the generated struct from the Protocol Buffers.
+
+### Do we want to have a separate go mod for each executor/source?
+
+I don't have a strong opinion here. For now, just for the sake of simplicity I would stay with shared dependencies as long as we don't extract full the `kubectl` executor and `kubernetes` source as external plugins.
+Later we can revisit this decision.
```
pkosiec, line 415: Agree with both. If the whole team agrees, then let's rephrase the sections to highlight this is a team decision 👍

docs/proposal/2022-11-22-botkube-plugins-integration.md
```diff
@@ -0,0 +1,498 @@
...
+              "title": "Kubernetes version",
+              "description": "Kubernetes version",
+              "default": "",
+              "examples": [
+                  "1.19.0"
+              ]
+            }
+          },
+          "additionalProperties": true
+        }
```
pkosiec, line 286: image

😄

docs/proposal/2022-11-22-botkube-plugins-integration.md
```diff
@@ -0,0 +1,498 @@
...
+```
+
+This will be used by Botkube to validate that configuration defined by user is valid before even starting a given plugin.
```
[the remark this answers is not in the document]
mszostok: yes 👍

docs/proposal/2022-11-22-botkube-plugins-integration.md
mszostok: >So it is based on the link itself - looking for specific pattern. Mixed feelings there - we won't be able to support some arbitrary links without such pattern.

I was also thinking about that and I checked how `gh` CLI works and they are doing exactly the same thing. The question is whether there is a real use-case for that. However, I need to agree that the contract is not visible OOTB.

However, I decided to just have it as simple as possible. 

Let's wait also for @huseyinbabal opinion and decide which one to pick 👍

docs/proposal/2022-11-22-botkube-plugins-integration.md
mszostok: I like it 👍

docs/proposal/2022-11-22-botkube-plugins-integration.md
mszostok: Personally, I have two concerns:
- is it a real use case? How you would use it for testing?
- if we will specify also the "original" name then IMO it doesn't fit into `aliases` definition 🤔

pkosiec: > is there a real use case for it? How you would use it for testing?

1. For example, I'd be able to fetch your forked kubectl executor to test your changes from PR (if they don't affect the whole Botkube) without touching the core which I have installed on Oracle Cloud.
2. Also, what if there are two different plugins, but with the same name? Like `https://github.com/travis-ci/gimme` and mszostok/gimme` (I know, stupid example, but that was the first binary name that came to my mind 😄) which do different things, but share the same name? There won't be a way to install and use them both, even if they come from different repos (!). I think this is a serious issue.
3. I would also like to invert the question: why two separate binaries could run into a conflict which we cannot resolve? If I have two versions of kubectl, I can rename the binary / make an alias and run it. Same with Go / Node.js versions. I just don't see any reason why we would need to block downloading two plugins with the same name. If there's any obstacle to do it, then of course we should reevaluate this 👍 But enforcing the community having unique names across all repositories is a huge downside.

> if we will specify also the "original" name then IMO it doesn't fit into aliases definition 🤔

yep, we can come up with different name for the `aliases` property if you don't like it - `commands`, `prefixes`, ...?
But I think `aliases` is rather OK. IMHO name is just... a name, which can be different from the actual command. For example, I have `mongo-seeding` project, but CLI is called `seed`, because `mongo-seeding` is not really a great name for a binary.

mszostok: Let's discuss it f2f:

1.  This still won't be possible 🤔 because we don't publish such binaries. And event if that will be the case, then probably even easier will be to just change `botkube/kubectl` to `mszostok/kubectl` 🤔 
2. Yeah, but is it real thing, when sth like that will happen 😄 ?
3. I'm trying to introduce as little as possible when it comes to the new features. 
4. Even if we will have what you describe, then validation is still needed as user still can define the same `aliases` in two places.

TL;DR;
1. By default, we use the plugin name, e.g. `botkube/kubectl` resolves to `kubectl`
2. If user defines in the index file the`aliases/commands` property, then we use only them.
    ```yaml
      - name: "kubectl"
        type: "executor"
        description: "Kubectl executor plugin."
        version: "v1.0.0"
        aliases: [ "kc", "k" ]  # overrides under which names the plugin will be available
    ```
3. If users want, they can also override it once enabling:
    ```yaml
    executors:
      'plugin-based':
        botkube/kubectl@v1.0.0:
          enabled: true
          aliases: ["kubectl", "kc", "k"]   # overrides under which names the plugin will be available
          config: {}
    ```

We would need to have some mechanism to tell the end-user under which names the plugin will be available.

Consider such case:
1. I enable `botkube/kubectl@v1.0.0`
2. In the `botkube` repo index the `kubectl@v1.0.0` has the `aliases: [ "kc", "k" ]` property.

In this way, the enabled plugin will be available only under `kc` and `k`. We would need to somehow tell about that our users.

pkosiec: What if admin is always required to define aliases? And the repo index won't contain them at all? Users will be in full control what prefixes they want to use. Why would a plugin author define aliases actually? 🤔

And yes, I didn't say we don't need validation for it 👍 it's more about name conflicts. I think if we can easily avoid them, let's do it 👍

mszostok: >What if admin is always required to define aliases? And the repo index won't contain them at all? Users will be in full control what prefixes they want to use. Why would a plugin author define aliases actually? 🤔

The only reason is that there are some well known aliases, e.g. for `kubectl`. Defining them in the index file solves one problem of redefining them each time you define configuration.

You can already see that we have multiple configuration for `kubectl` executor. If we do not have aliases defined in index file. Then each time we will need to remember to do that.

docs/proposal/2022-11-22-botkube-plugins-integration.md
```diff
@@ -0,0 +1,498 @@
...
+
+This section describes the necessary changes in the syntax. **It's backward compatible.**
+
+1. A new `plugins` property. It allows to specify a list of repository from where the plugins can be downloaded.
+   ```yaml
+   plugins:
+     cacheDir: "/tmp"
+     repositories:
+       botkube: https://plugins.botkube.io/botkube.yaml
+       huseyinbabal: https://raw.githubusercontent.com/huseyinbabal/botkube-plugins/main/index.json
```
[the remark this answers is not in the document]
huseyinbabal: I agree with @pkosiec and more concrete example would be Helm's dependency system. So you basically define your dependencies (it is repo indexes in our case) as follows,

```
# Chart.yaml
dependencies:
- name: kubectl
  version: "1.2.3"
  repository: "https://botkube.io/repo.yaml"
- name: kubernetes
  version: "3.2.1"
  repository: "https://another.example.com/index.yaml"
```

Also you showed, you are using repo in your plugin configuration like <repo>/<plugin>, but we cannot use multiple instance of a specific plugin from different repos right? If we use Helm's dependency structure, we can also eliminate the usage `huseyin/botkube` and just use `botkube`.

docs/proposal/2022-11-22-botkube-plugins-integration.md
```diff
@@ -0,0 +1,498 @@
...
+   ```
+
+2. The `executors` definition now can refer to the executor plugins specified in a given repository.
+   ```yaml
+   executors:
+     'plugin-based':
```
huseyinbabal, line 51: I didn't get this field, it is for back ground compatibility to not corrupt our confs under executors?

docs/proposal/2022-11-22-botkube-plugins-integration.md
```diff
@@ -0,0 +1,498 @@
...
+       botkube/kubectl:     # <repo>/<plugin> is syntax for plugin based executors
```
huseyinbabal, line 52: If we use Helm's dep system, we can use `kubectl` only.

docs/proposal/2022-11-22-botkube-plugins-integration.md
huseyinbabal: When I see version, I expect there might be 2 versions of same plugin at the same time. Could you please elaborate on why we don't  trust the version in dependencies section only and eliminate from here?

docs/proposal/2022-11-22-botkube-plugins-integration.md
huseyinbabal: what @pkosiec suggested is close to [docker's manifest system](https://docs.docker.com/engine/reference/commandline/manifest/#inspect-an-images-manifest-and-get-the-osarch-info). If we try to use more standardized format, it will help us to detect customer's os/arch with go and map it to exact manifest.

huseyinbabal:
Let's recap tomorrow, we can easily have a consensus and unblock you :)

docs/proposal/2022-11-22-botkube-plugins-integration.md
mszostok: FYI: accepted format:
```
  - name: "kubernetes"
    type: "source"
    description: "Kubernetes source plugin."
    version: "v1.0.0"
    urls:
      - url: https://github.com/kubeshop/botkube/releases/download/v0.17.0/source_kubernetes-darwin-amd64
        platform:
          architecture: amd64
          os: darwin
```

docs/proposal/2022-11-22-botkube-plugins-integration.md
```diff
@@ -0,0 +1,542 @@
...
+Currently, the aliases are built-in into Botkube core logic. We can allow to specify them when enabling a given plugin:
+
+```yaml
+executors:
+  'plugin-based':
+    botkube/kubectl@v1.0.0:
+      enabled: true
+      aliases: [ "kc", "k" ]
+```
+
```
pkosiec, line 121: I think one additional sentence will be helpful:

```suggestion

With such configuration, the `kubectl` executor plugin can be used with the following command prefixes: `kubectl`, `kc` and `k`.
```

Or something similar 🙂

docs/proposal/2022-11-22-botkube-plugins-integration.md
```diff
@@ -0,0 +1,542 @@
...
+						 include: ["botkube", "default","ambassador"]
+					 commands:
+						 verbs: ["get","logs","top"]
+						 resources: ["pods","deployments","nodes","configmap"]
+			 botkube/kubectl@v1.2.0:  # not allowed as it is conflicting with already registered 'kubectl' from the Botkube repository in different version.
+				 enabled: true
+				 config:
+					 commands:
+						 verbs: ["get","logs","top"]
+	 ```
```
pkosiec, line 191: Something's wrong with formatting, please fix it

docs/proposal/2022-11-22-botkube-plugins-integration.md
```diff
@@ -0,0 +1,542 @@
...
+This will be used by Botkube to validate that configuration defined by user is valid before even starting a given plugin.
+
+### Passing configuration to a given plugin
+
+Current Botkube implementation allows you to specify different executor and sources configuration and bind them to a single channel. Later they are merged together based on the given business logic.
+
+To support the same experience for plugins, we need to pass a list of configuration each time we run the:
+
+- `Execute` method for executor plugins
+- `Source` method for source plugins
```
pkosiec, line 275: It would be good to clarify when these methods are run. I think this will be accurate:

```suggestion
- `Execute` method for executor plugins, invoked when user runs a given command,
- `Source` method for source plugins, invoked during Botkube startup.
```

docs/proposal/2022-11-22-botkube-plugins-integration.md
pkosiec: This comment is still not resolved. Please replace the phrases, e.g.:

- Remove "I don't have a strong opinion here"
- Replace "I reuse" to "We'll reuse"
- Replace "I would stay" to "We'll keep the shared dependencies"

Also, replace the questions for the section headers to something like:

- "Decoupling interfaces between Go implementation and gRPC"
- "Separate Go modules for each plugin"

docs/proposal/2022-11-22-botkube-plugins-integration.md
```diff
@@ -0,0 +1,542 @@
...
+│  ├── api      # Botkube public API that can be imported also by some 3rd party libs/apps.
+│  │  ├── executor
+│  │  └── source
+│  └── ...
+└── proto        # here are the Protocol buffers API that can be used to generate clients in different languages. It's on the root as it's Botkube not Go specific files. We generate Go client/server into `pkg/api/{executor|source}`, so it can be use by other 3rd plugins.
+```
+
+## Alternatives
+
+Other approaches that I consider but were introducing too many complications.
```
pkosiec, line 454:
```suggestion
Other approaches that were considered but rejected because of the complexity.
```

docs/proposal/2022-11-22-botkube-plugins-integration.md
```diff
@@ -0,0 +1,542 @@
...
+          architecture: amd64
+          os: linux
+      - url: https://github.com/kubeshop/botkube/releases/download/v0.17.0/executor_kubectl-linux-arm64
+        platform:
+          architecture: arm64
+          os: linux
+```
+
+Optionally
+</details>
```
pkosiec, line 542: I think it will be good to mention also the alternative idea with defining all the command prefixes by the user in a single field (`aliases/prefixes: [kubectl, k, kc]`) we were discussing recently 👍

pkosiec (approved):
LGTM - just a few minor comments 🚀

docs/proposal/2022-11-22-botkube-plugins-integration.md
mszostok: added

Issue #802, opened 2022-10-10
Deprecated support for typing commands without `k|kc|kubectl` prefix


mszostok:
## Overview

We are going to introduce a required kubectl prefix (one of `k`, `kc`, `kubectl`) for commands typed in communicators. Instead of
`@Botkube get pod` user will need to type one of:
- `@Botkube k get pod`
- `@Botkube kc get pod`
- `@Botkube kubectl get pod`

**Deprecation steps:**
1. ~In 0.14:~
     - we already print the warning that `k|kc|kubectl` will be required in the future versions (see: https://github.com/kubeshop/botkube-docs/pull/150)
2. ~In 0.15:~
      - we change [the warning](https://botkube.io/docs/usage/#debugging-deployments-using-botkube) and say that `k|kc|kubectl` will be required in the 0.17 version
      - we replace all text examples and add required prefix e.g. change command from `@Botkube get po` to `@Botkube kubectl get po` in the Help message.
announce that in the release notes
3. ~In 0.16:~
      - we try to replace all screenshot examples and add required prefix e.g. change command from `@Botkube get po` to `@Botkube kubectl get po`
    - We print in the help message that is sent by Botkube on a given channel, warning that the `k|kc|kubectl` prefix will be required in the next version (0.17)

4. In 0.17:
   -  prefix is required, and all commands without prefix are processed as the Botkube ones, e.g. as we have for `@Botkube notifier stop`
    - ~if user executes command without the prefix print a warning message that it is not supported anymore.~

## Reason

We want to drive a clear line between the executor commands and Botkube built-in commands. This separation simplifies introducing more executors such as `helm`, `argo`, etc.

PR #861, opened 2022-11-23
Add warning about kubectl to the help message

mszostok:
## Description

- Add warning about kubectl to the help message. Examples:
  - interactive Screen Shot 2022-11-23 at 14 40 30
  - legacy Screen Shot 2022-11-23 at 14 40 38
  - discord Screen Shot 2022-11-23 at 14 40 43

## Related issue(s)

- https://github.com/kubeshop/botkube/issues/802


pkosiec:
Open question: Do you think we should also print such message for each kubectl command execution which doesn't have the required prefix? 🤔

pkosiec:
Annoying, but it could teach users to include the prefix. If someone doesn't check help, then probably it could be missed 🤔 But no strong opinion there, so I'll rely on @lpetkov opinion 😄

lpetkov:
I'd say we need to print this message to users who are on 0.17+ and still try to run kubectl commands the old-fashioned way. For 0.16 it's fine the way it is as it works.

mszostok:
OK 👍 I will not that in the issue for 0.17 👍

pkg/bot/interactive/help.go
```diff
@@ -156,12 +156,18 @@ func (h *HelpMessage) actionSections() []Section {
 }
 
 func (h *HelpMessage) kubectlSections() []Section {
+	// TODO(https://github.com/kubeshop/botkube/issues/802): remove this warning in after releasing 0.17.
+	warn := ":warning: Botkube 0.17 and above will require a prefix (`k`, `kc`, `kubectl`) when running kubectl commands through the bot.\n\n`@Botkube k get pods` instead of `@Botkube get pods`\n"
```
pkosiec, line 160: If you use ⚠️, please add it to the MS Teams emoji mapping. Thanks!

pkosiec (approved):
LGTM with small minor comment

PR #875, opened 2022-12-01
Pass configuration to source and executor plugins

### LINKED ISSUE #872: Passing configuration to a given plugin
## Overview

Botkube support multiple bindings to the same executor/source. The configuration is later merged based on a given business logic. For example, in case of `kubectl` executor, here is a [merging strategy](https://docs.botkube.io/configuration/executor#merging-strategy). 

To support the same experience for plugins, we need to pass a list of configuration each time we run the:

- `Execute` method for executor plugins, invoked when user runs a given command,
- `Source` method for source plugins, invoked during Botkube startup.
 
We pass an array of strings, so later they can be unmarshalled by a given plugin and merged based on custom business logic.

## Acceptance Criteria

- Update the source and executor gRPC API
- Update example source and executor to handle specified configuration
- Resolve `TODO(configure plugin)`  For example:
   - https://github.com/kubeshop/botkube/blob/525c737956ff820a09321879284037da8bf5d647/pkg/execute/plugin_executor.go#L72-L76

mszostok:
## Description

- Pass configuration to source and executor plugins

## Related issue(s)

Fix #872


internal/source/dispatcher.go
```diff
@@ -0,0 +1,88 @@
+package source
```
mszostok, line 1: I just split the Dispatcher service into two:
- `Scheduler` analyzes the provided configuration and based on that schedules plugin sources.
- and `Dispatcher` provides functionality to starts a given plugin, watches for incoming events and calling all notifiers to dispatch received event.
-

pkg/api/source/grpc_adapter.go
```diff
@@ -7,14 +7,13 @@ import (
...
 	"google.golang.org/grpc"
-	"google.golang.org/protobuf/types/known/emptypb"
 
 	"github.com/kubeshop/botkube/pkg/api"
 )
 
 // Source defines the Botkube source plugin functionality.
 type Source interface {
-	Stream(ctx context.Context) (StreamOutput, error)
+	Stream(ctx context.Context, configs [][]byte) (StreamOutput, error)
```
mszostok, line 16: as the `[][]byte` is used if a few places, I was thinking whether to introduce a dedicated type like:
- `type Configuration []byte` so in func signature: `Stream(ctx context.Context, configs []plugin.Configuration)`
- or - `type Configurations [][]byte` so in func signature: `Stream(ctx context.Context, configs plugin.Configurations)`

WDYT? I would require to stop using the `Request` `Response` on Executor plugin. Is it worth it? On the other hand, it will unify the approach and make the API more self-descriptive.

josefkarasek (approved):
lgtm

proto/source.proto
```diff
@@ -11,6 +9,11 @@ message StreamResponse {
 	bytes output = 1;
 }
 
+message StreamRequest {
+	// Holds list of config YAML definitions in order specified by user
```
josefkarasek, line 13: What does this comment mean?

pkg/api/source/grpc_adapter.go
josefkarasek: Would sending `repeated Config` over network instead of `repeated bytes` also make sense?

proto/source.proto
mszostok: It means that the `configs` property holds the list of config in YAML format, I also wanted to express that the order is preserved. 

The Go models are later auto-generated and this comment is also above given property so when you use this struct you know what to expect from it.

pkg/api/source/grpc_adapter.go
mszostok: Not sure if I understand that correctly. In my comment I only mentioned the Go implementation where we can do simply type aliasing, I don't think that is also supported by protobufs.

mszostok: After a f2f discussion:
```
message Config {
   RawYAML bytes = 1
}

repeated Config
```

so in Go: `config.RawYAML`.

mszostok:
Hi @josefkarasek! 
Thanks for valuable feedback 🙂 In [this commit](https://github.com/kubeshop/botkube/pull/875/commits/3e9622d327f2f4b22c64e616bbf03bca43ed7f34), I updated the gRPC according to our discussion.
I spot and updated the old plugin repository definition too.

PR #883, opened 2022-12-12
Make the integration jobs independent

mszostok:
## Description

- Currently, when one of a job in matrix fails, the other one is canceled. This is problematic when we want to debug the root cause of the failure. If we make those jobs independent then we can see which one really fails and if it is a generic problem or only with one of our platform integration.

Changes are introduced based on the GitHub action documentation: https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#handling-failures

Example where Discord test failed and as a result canceled the Slack testing: https://github.com/kubeshop/botkube/actions/runs/3677219708

### DIFF
--- a/.github/workflows/branch-build.yml
+++ b/.github/workflows/branch-build.yml
@@ -60,6 +60,9 @@ jobs:
       packages: read
 
     strategy:
+      # make the jobs independent
+      fail-fast: false
+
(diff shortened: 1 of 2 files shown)


josefkarasek (approved):
Good job :+1:

PR #882, opened 2022-12-12
Add release process for Botkube plugins

mszostok:
## Introduction

https://youtu.be/b06cZRzlJgM

## Description

Merge after: https://github.com/kubeshop/botkube/pull/884

Changes proposed in this pull request:
**Flow:**

1. Plugins binaries are build via GoReleaser
   - the `version` field is automatically set via ldflags
   - binary name format: `{executor|source}_{pluginName}_{os}_{arch}`
2. The index file is built via Go service based on directory with plugins binaries
    - type, plugin name, os and arch parameters are parsed from the binary name
    - description and version are taken by calling the `Metadata` method on each plugin

 
Benefits of implemented approach with auto-generation:
- because we use the Go code, we use always up-to-date index entry definition. If we change/add a property, it will be automatically reflected in the generated index file
- the e2e tests and release job utilize the same service for generating the index file, so on each PR we test that generate index can be properly consumed by Botkube
- because we introduced the `Metadata` method for each plugin, we can:
   - have a strong typing and validation for required metadata fields
   - we define the metadata close to the plugin implementation, no need for additional YAML/Markdown file.

**Fixed bugs:**
- fix env setting in CI workflows
- fix archive section in GoReleaser

## Testing

I changed the e2e tests workflow, so it won't work on this PR. However, I tested that on my fork:

The RC release:
- CI job: https://github.com/mszostok/botkube/actions/runs/3685786750/jobs/6237163704
- GitHub release: https://github.com/mszostok/botkube/releases/tag/untagged-2936565ff27e5ba54c8c

The final release:
- CI job: https://github.com/mszostok/botkube/actions/runs/3685975503/jobs/6237589515
- GitHub release: https://github.com/mszostok/botkube/releases/tag/v0.66.0

### Running integration tests

Install Botkube:
```
export SLACK_BOT_TOKEN="xoxb-"

helm upgrade botkube --install --namespace botkube ./helm/botkube --wait --create-namespace \
  -f ./helm/botkube/e2e-test-values.yaml \
  --set image.repository=kubeshop/pr/botkube \
  --set image.tag=882-PR \
  --set image.pullPolicy="Always" \
  --set communications.default-group.slack.token="${SLACK_BOT_TOKEN}"
```

Build plugins:
```
make build-plugins
```

Run tests:
```
export DISCORD_TESTER_APP_TOKEN="dummy"
export DISCORD_GUILD_ID="dummy"
export DISCORD_TESTER_NAME="dummy"

export KUBECONFIG="<config>"
[...]

Issue #871, opened 2022-11-29
Add Botkube plugin release job


mszostok:
### Releasing Botkube plugins

As a part of 0.17 release, such binaries should be added to the GitHub release assets:

- `plugins-index.yaml`
- `source_kubernetes-darwin-amd64`
- `source_kubernetes-linux-amd64`
- `executor_kubectl-darwin-amd64`
- `executor_kubectl-linux-amd64`

Where the `plugins-index.yaml` is defined as follows:

```yaml
entries:
  - name: "kubectl"
    type: "executor"
    description: "Kubectl executor plugin."
    version: "v1.0.0"
    urls:
      - url: https://github.com/kubeshop/botkube/releases/download/v0.17.0/executor_kubectl-darwin-amd64
        platform:
          architecture: amd64
          os: darwin
      - url: https://github.com/kubeshop/botkube/releases/download/v0.17.0/executor_kubectl-darwin-arm64
        platform:
          architecture: arm64
          os: darwin
      - url: https://github.com/kubeshop/botkube/releases/download/v0.17.0/executor_kubectl-linux-amd64
        platform:
          architecture: amd64
          os: linux
      - url: https://github.com/kubeshop/botkube/releases/download/v0.17.0/executor_kubectl-linux-arm64
        platform:
          architecture: arm64
          os: linux
  - name: "kubernetes"
    type: "source"
    description: "Kubernetes source plugin."
    version: "v1.0.0"
    urls:
      - url: https://github.com/kubeshop/botkube/releases/download/v0.17.0/source_kubernetes-darwin-amd64
        platform:
          architecture: amd64
          os: darwin
      - url: https://github.com/kubeshop/botkube/releases/download/v0.17.0/source_kubernetes-darwin-arm64
        platform:
          architecture: arm64
          os: darwin
      - url: https://github.com/kubeshop/botkube/releases/download/v0.17.0/source_kubernetes-linux-amd64
        platform:
          architecture: amd64
          os: linux
      - url: https://github.com/kubeshop/botkube/releases/download/v0.17.0/source_kubernetes-linux-arm64
        platform:
          architecture: arm64
          os: linux
```

The index file can be generated automatically just by scanning the directory where plugins binaries are stored.

## Acceptance Criteria

- Release all available plugins as a part of automated release process
  - source binaries
  - executor binaries
  - index.yaml
- Support patch releases too
- Provide an example of successful release done on your fork
- Update release documentation: https://docs.botkube.io/community/contribute/release

PR #884, opened 2022-12-13
Add automated way of generating plugin index file

mszostok:
## Description

I extracted the Go service that generates the index YAML from the CI release job that uses it. It allows as to review a smaller part and also use the current CI to prove that it works. On the https://github.com/kubeshop/botkube/pull/882 is not possible to make the e2e test pass as I had to change the CI workflow definitions. 

- Add automated way of generating plugin index file
  - Introduce the `Metadata` RPC both to executors and sources.
  - Implement plugin index builder based on directory with plugins binaries

For more details, see: https://github.com/kubeshop/botkube/pull/882 where you can check how it comes together.

## Related issue(s)

- https://github.com/kubeshop/botkube/issues/871


josefkarasek (approved):
lgtm :+1:

PR #885, opened 2022-12-13
Dynamically generate help messages

josefkarasek:
## Dynamically generate help message when subcommand is missing

For commands that support subcommands/resources, when such subcommand is missing, return a dynamically
generated table of all subcommands supported by the command.

## Related issue(s)

#703

### DIFF
--- a/pkg/execute/executor.go
+++ b/pkg/execute/executor.go
@@ -23,49 +23,13 @@ const (
 	internalErrorMsgFmt = "Sorry, an internal error occurred while executing your command for the '%s' cluster :( See the logs for more details."
 	emptyResponseMsg    = ".... empty response _*<cricket sounds>*_ :cricket: :cricket: :cricket:"
 
-	// incompleteCmdMsg incomplete command response message
-	incompleteCmdMsg = "You missed to pass options for the command. Please use 'help' to see command options."
-
 	anonymizedInvalidVerb = "{invalid verb}"
 
 	lineLimitToShowFilter = 16
-
-	// TODO: build the help msg dynamically
-	helpMessageList = `@Botkube list [resource]
-
-Available resources:
-actions  | action  | act          list available automations
-filters  | filter  | flr          list available filters
-commands | command | cmds | cmd   list enabled executors`
-
-	helpMessageEdit = `@Botkube edit [resource]
-
-Available resources:
-sourcebindings | actsourcebinding   edit source bindings`
-
-	helpMessageEnable = `@Botkube enable [resource]
-
-Available resources:
-actions | action | act    enable available automations
-filters | filter | fil    enable available filters`
-
-	helpMessageDisable = `@Botkube disable [resource]
-
-Available resources:
-actions | action | act    disable available automations
-filters | filter | fil    disable available filters`
 )
 
 var (
-	// noResourceNames is used for commands that have no resources defined
-	noResourceNames       = []string{""}
-	clusterNameFlagRegex  = regexp.MustCompile(`--cluster-name[=|\s]+\S+`)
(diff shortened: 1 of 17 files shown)


pkg/execute/executor.go
```diff
@@ -267,19 +193,18 @@ func (e *DefaultExecutor) Execute(ctx context.Context) interactive.Message {
...
+	if !foundRes {
 		e.reportCommand(anonymizedInvalidVerb, false)
 		e.log.Infof("received unsupported command: %q", execFilter.FilteredCommand())
 		return respond(unsupportedCmdMsg, cmdCtx)
 	}
 
-	fn, found := resources[cmdRes]
-	if !found {
+	if !foundFn {
 		e.reportCommand(fmt.Sprintf("%s {invalid resource}", cmdVerb), false)
```
mszostok, line 204: **nit**
```suggestion
		e.reportCommand(fmt.Sprintf("%s {invalid feature}", cmdVerb), false)
```

we forgot to update that on the previous PR.

pkg/execute/executor.go
```diff
@@ -267,19 +193,18 @@ func (e *DefaultExecutor) Execute(ctx context.Context) interactive.Message {
...
 		e.log.Infof("received unsupported resource: %q", execFilter.FilteredCommand())
 		msg := incompleteCmdMsg
-		if helpMessage, ok := availableHelpMessages[cmdVerb]; ok {
+		if helpMessage := e.cmdsMapping.HelpMessageForVerb(cmdVerb, botName); helpMessage != "" {
 			msg = helpMessage
 		}
```
mszostok, line 209: maybe we can move the `incompleteCmdMsg` to `HelpMessageForVerb` and return it in case there is no other help message. As a result, we make this flow more straight.
```go
msg := e.cmdsMapping.HelpMessageForVerb(cmdVerb, botName) // it's up to the help provider to resolve the final message
```

pkg/execute/mapping.go
```diff
@@ -0,0 +1,150 @@
...
+				}
+				cmdsMapping[verb][featureName] = cmdFn
+			}
+		}
+	}
+	mapping := &CommandMapping{
+		commands: cmdsMapping,
+		help:     helpMapping,
+	}
+	return mapping, mappingsErrs.ErrorOrNil()
```
mszostok, line 104: **nit** In Go there is a pattern to return the zero value for other values when err occurs. It's also more readable as you see the err paths directly. It's also consistent with other principle to align the happy path to the left edge.

```suggestion
	if err := mappingsErrs.ErrorOrNil(); err != nil {
		return nil, err
	}

	return &CommandMapping{
		commands: cmdsMapping,
		help:     helpMapping,
	}, nil
```

pkg/execute/mapping.go
```diff
@@ -0,0 +1,150 @@
...
+		return nil, true, false
+	}
+	return fn, true, true
+}
+
+// HelpMessageForVerb dynamically builds help message for given CommandVerb, or empty string
+func (m *CommandMapping) HelpMessageForVerb(verb CommandVerb, botName string) string {
+	cmd, ok := m.help[verb]
+	if !ok {
+		return ""
```
mszostok, line 124: here we can just return `incompleteCmdMsg` as I mentioned in the previous comment. This will simply the code and also group it in a single place. For example, right now the `HelpMessageForVerb` func is used in 2 places, so we will make sure that each time a proper help message is returned.

mszostok (approved):
It's perfect, good job 🚀 

I left some minor comments that will be good to address before merging, however I already approve this PR as I don't have any objections to the main logic.

PR #895, opened 2022-12-19
Fix notifier verbs

josefkarasek:
## Fix notifier verbs

Based on #703 notifier verbs are `enable, disable` and not `start, stop`.

## Related issue(s)

#703


mszostok (approved):
LGTM

mszostok:
Can we also fix this comment on this PR: https://github.com/kubeshop/botkube/issues/703#issuecomment-1359960622 ?

pkosiec:
I think the help (`@Botkube help`) should be also updated.

Also, I think we miss another thing:

> `@Botkube` - same as `@Botkube help`

Source: https://github.com/kubeshop/botkube/issues/703#issuecomment-1308613253

Does it work this way? Looking at the code I don't think so 🤔 If not, please also add it here or in a separate PR, whatever it's easier for you 👍

PR #896, opened 2022-12-19
config source provider is added

huseyinbabal:
## Description

- Added new config source provider to fetch config from


internal/config/gql_provider.go
```diff
@@ -0,0 +1,45 @@
...
+	gqlClient *GqlClient
+}
+
+// NewGqlProvider initializes new GraphQL config source provider
+func NewGqlProvider(gql *GqlClient) *GqlProvider {
+	return &GqlProvider{gqlClient: gql}
+}
+
+// Configs returns list of config files
+func (g *GqlProvider) Configs(ctx context.Context) ([]string, error) {
```
pkosiec, line 23: Do we need to create file? I know that koanf supports loading config from raw bytes: https://github.com/knadh/koanf#reading-raw-bytes

That would simplify the code 👍

pkg/config/config.go
```diff
@@ -667,15 +672,18 @@ func LoadWithDefaults(getCfgPaths PathsGetter) (*Config, LoadWithDefaultsDetails
...
-		return strings.Split(envCfgs, ",")
+func FromProvider(gql *config.GqlClient) ([]string, error) {
+	var provider config.Provider
+	if os.Getenv("CONFIG_SOURCE_IDENTIFIER") != "" {
+		provider = config.NewGqlProvider(gql)
+	} else if os.Getenv("BOTKUBE_CONFIG_PATHS") != "" {
+		provider = config.NewEnvProvider()
+	} else {
+		provider = config.NewStaticProvider(configPathsFlag)
 	}
```
pkosiec, line 685: I suggest refactoring the code a bit to reduce indentations:

```go
if ... {
    return ...
}

if ... {
    return ...
}
```

BTW you can already return `(Config, error)` struct instead of file paths. That is related to my previous comment about reading raw bytes directly by koanf 👍

pkosiec (commented):
This is not a full review as @josefkarasek will do it, but just throwing two small comments:

internal/config/gql_provider.go
huseyinbabal: I didn't wanted to do so much change, as you know the config struct used in koanf is dedicated to yaml, where I am using json notation for botkubeConfig. That is why I have written that in file and provided as application parameters. So, if I use koanf to unmarshall to our config, I need to add `json` tag to all struct fields.
Instead I have saved the config and reused botkube environment vars to prevent a huge change.

pkg/config/config.go
huseyinbabal: I would but I am using `provider` variable at the end of the function.

internal/config/gql_provider.go
pkosiec: What about unmarshalling it by yourself and using Go struct provider https://github.com/knadh/koanf/blob/master/providers/structs/structs.go#L22?

internal/config/gql_provider.go
```diff
@@ -0,0 +1,45 @@
...
+	d := os.Getenv("CONFIG_SOURCE_IDENTIFIER")
+	if d == "" {
+		return nil, nil
```
pkosiec, line 26: Maybe we could return here an error? WDYT? 🤔

internal/config/gql_provider.go
huseyinbabal: Yesterday I paired with Mateusz and restructured it, now it is not generating file, it returns set of config contents instead.

internal/config/gql_provider.go
huseyinbabal: This part is also outdated, please check new content

cmd/botkube/main.go
```diff
@@ -28,6 +29,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/manager/signals"
 
 	"github.com/kubeshop/botkube/internal/analytics"
+	intConfig "github.com/kubeshop/botkube/internal/config"
```
josefkarasek, line 32: I wouldn't use came case here.

[go.dev](https://go.dev/blog/package-names) says _"They are lower case, with no under_scores or mixedCaps."_

pkg/config/config.go
```diff
@@ -667,15 +661,18 @@ func LoadWithDefaults(getCfgPaths PathsGetter) (*Config, LoadWithDefaultsDetails
...
+	if os.Getenv("CONFIG_SOURCE_IDENTIFIER") != "" {
+		provider = config.NewGqlProvider(*gql)
+	} else if os.Getenv("BOTKUBE_CONFIG_PATHS") != "" {
+		provider = config.NewEnvProvider()
+	} else {
+		provider = config.NewFileSystemProvider(configPathsFlag)
 	}
-
-	return configPathsFlag
+	return provider.Configs(context.Background())
```
josefkarasek, line 675: Shouldn't context come from the calling function?

josefkarasek (commented):
Good work @huseyinbabal :+1: ! Just want to make sure I understand some of your choices, will approve then.

pkg/config/config.go
huseyinbabal: calling function does not have context, and I needed to pass context since there is a such contract

josefkarasek: Can we take this context? https://github.com/kubeshop/botkube/blob/3aa76fade5ba174b6829942a0bc095d46eb21558/cmd/botkube/main.go#L105

huseyinbabal: But it is set after config load operation.Even analytics initialization is done without context.
Screen Shot 2022-12-23 at 15 01 19

josefkarasek: Just a suggestion. The ctx could me moved.

josefkarasek (approved):
lgtm

pkg/config/config.go
huseyinbabal: I created a follow up task for that https://github.com/kubeshop/botkube/issues/908

Issue #899, opened 2022-12-20
Add plugin dependency management into Botkube


mszostok:
## Overview

Add plugin dependency management into Botkube. The `MetadataOutput` should support specifying a list of binaries (in different os/arch) that are a dependency for a given plugin. This should be included in the generated plugin index YAML file.

```yaml
entries:
  - name: "helm"
    type: "executor"
    description: "Helm executor plugin."
    version: "v1.0.0"
    urls:
      - url: https://github.com/kubeshop/botkube/releases/download/v0.17.0/executor_helm-darwin-amd64
        dependencies:
           helm:
              url: https://github.com/kubeshop/botkube/releases/download/v0.17.0/helm-darwin-amd64
        platform:
          os: darwin
          architecture: amd64
      - url: https://github.com/kubeshop/botkube/releases/download/v0.17.0/executor_helm-darwin-arm64
        dependencies:
          helm: https://github.com/kubeshop/botkube/releases/download/v0.17.0/helm-darwin-amd64
        platform:
          os: darwin
          architecture: arm64
      - url: https://github.com/kubeshop/botkube/releases/download/v0.17.0/executor_helm-linux-amd64
        dependencies:
          helm: https://github.com/kubeshop/botkube/releases/download/v0.17.0/helm-darwin-amd64
        platform:
          os: linux
          architecture: amd64
      - url: https://github.com/kubeshop/botkube/releases/download/v0.17.0/executor_helm-linux-arm64
        dependencies:
          helm: https://github.com/kubeshop/botkube/releases/download/v0.17.0/helm-darwin-amd64
        platform:
          os: linux
          architecture: arm64
```

Botkube should collect all dependencies of enabled plugins and install all required binaries.

Issue #843, opened 2022-11-08
Helm as alternative executor plugin


huseyinbabal:
## Overview

Botkube has existing support for `kubectl`, now we will be adding helm support to execute helm commands through Botkube.

## Acceptance Criteria

- [x] Helm executor plugin should have the same experience (commands, flags) as Helm CLI
- [x] Implement Helm plugin for Botkube
    - interactivity is out of scope
    - install, uninstall, upgrade, rollback, list, version, test, status
       - `--repo` - custom flag which adds the repo for a short period of time
       - support `--set` for now
       - ensure multiline commands work properly
- [x] Design a configuration structure for the helm executor plugin
- [x] Document it in plugins folder under the main Botkube repository.

Stretch:

- [ ] add `helm history` command
- [ ] add `helm get [all|manifest|hooks|notes]` commands
## Reason

Helm support was heavily requested by the community. That's why this would be one of the first new executors.

brampling:
We don't need to support all of the commands that are used for creating Helm charts. We should support the commands used for working with software installed using Helm (show, install, uninstall, upgrade, rollback, etc.).

mszostok:
@brampling @lpetkov 

Do we want to add a Helm related section to our Help message? _(visible only when Helm plugin is enabled)_
Something similar to what we have for `kubectl`?
<img width="818" alt="Screen Shot 2022-12-21 at 23 29 10" src="https://user-images.githubusercontent.com/17568639/209014582-54e6bfe2-1ed5-412d-9890-764e60198aea.png">

mszostok:
Stretch:
- add `history` command
- add `helm get [all|manifest|hooks|notes]` commands

brampling:
Yes, this would be nice to add for now. If we eventually get lots of plugins we should make it generic, but in this version it's useful to get users to try the new feature.

PR #900, opened 2022-12-20
Add Helm plugin that supports `helm install`

mszostok:
## Description

- Support all requested helm commands
- Validation:
  - Detect unknown flags
  - Detect known but not supported flags
  - Detect wrong syntax
  - For validation a real flag parser `go-arg` lib is used, to do not play with regexp as we do in Botkube, e.g. https://github.com/kubeshop/botkube/blob/03c742d3cacd133839bd54687ffd4ee0b5f4951a/pkg/execute/params.go#L25)
- Print dedicated helm message for each command
- Implemented in a way that it is extendable in the future, we can add custom logic at each level and also use helm flag values if needed.

### Issues
- `--filter` flag is globally taken by Botkube
- `--filter` it actually executes the command one more time, it's quite bad in case of `install`, `delete`, etc.
- the `go-arg` lib:
   - handle `--version` flag internally, which is problematic for us
   - don't support aliases for subcommands
   For all the above I implemented a simple workaround but will try to contribute to this library to remove our hacks. If you know a better library that we can use, let me know 👍

## Notes

- For now, the `helm` binary is embedded into Botkube Docker image. It will be solved once we will implement https://github.com/kubeshop/botkube/issues/899
- I planned to only implement the `install` command first. But finally, I implement all in this PR, however only `install` is covered with unit-test. Personally, I don't think if there is a value in those tests. Please let me know what do you think about it and if we should add coverage for other methods, I will do that in the follow-up PR.

## Testing 

https://user-images.githubusercontent.com/17568639/209023645-98e8de18-df26-4921-b4df-bf11c30d8369.mov

Basic integration is tested by e2e test (`helm install --help` command). The below script shows how to run Botkube locally to test all commands.

### Install Botkube 

1. Create `/tmp/test-values.yaml`:
    ```yaml
    communications:
      default-group:
        socketSlack:
          enabled: true
          channels:
            default:
              name: random
              bindings:
                executors:
                  - 'plugin-based'
          appToken: "xapp-1-"
          botToken: "xoxb-"

    executors:
      'plugin-based':
        botkube/helm:
          enabled: true
          config: {}
    
    extraEnv:
      - name: LOG_LEVEL_EXECUTOR_BOTKUBE_HELM
        value: "debug"
    
    plugins:
      cacheDir: "/tmp/plugins"
      repositories:
        botkube:
          url: http://host.k3d.internal:3000/botkube.yaml

    rbac:
      create: true
      rules:
        - apiGroups: ["*"]
          resources: ["*"]
          verbs: ["get", "watch", "list", "create", "delete"]
    ```
3. Start k3d cluster 

    ```bash
    k3d cluster create labs --image=rancher/k3s:v1.25.0-k3s1
    ```

4. Build plugins:
    ```bash
    make build-plugins
    ```
5. Start plugin server:
    ```bash
    env PLUGIN_SERVER_HOST=http://host.k3d.internal go run test/helpers/plugin_server.go
    ```
6. Install Botkube
    ```bash
    helm upgrade botkube --install --namespace botkube ./helm/botkube --wait --create-namespace \
      -f /tmp/test-values.yaml \
      --set image.tag=900-PR \
      --set image.pullPolicy="Always"
    ```

### Steps
1. Version
	 ```bash
	 @Botkube helm help
	 @Botkube helm --help
	 ```
1. Version
	 ```bash
	 @Botkube helm version -h
	 @Botkube helm version
	 ```
1. Install
	 ```bash
	 @Botkube helm install -h
	 
	 # By absolute URL:
	 @Botkube helm install --repo https://charts.bitnami.com/bitnami psql postgresql
	 # By chart reference:
	 @Botkube helm install https://charts.bitnami.com/bitnami/postgresql-12.1.0.tgz --create-namespace -n test2 --generate-name --set clusterDomain='testing.local'
	 ```
1. List
	 ```bash
	 @Botkube helm list -h
	 
	 @Botkube helm list -A
	 @Botkube helm list -f 'p' -A
	 ```
1. Status
	 ```bash
	 @Botkube helm status -h
	 @Botkube helm status psql
	 ```
1. Upgrade
	 ```bash
	 @Botkube helm upgrade -h
	 @Botkube helm upgrade --repo https://charts.bitnami.com/bitnami psql postgresql --set clusterDomain='testing.local'
	 ```
1. Rollback
	 ```bash
	 @Botkube helm rollback -h
	 @Botkube helm rollback psql
	 ```
1. Test
	 ```bash
	 @Botkube helm test -h
	 @Botkube helm test psql
	 ```
1. uninstall
	 ```bash
	 @Botkube helm uninstall psql
	 ```
1. Unknown flag
	 ```bash
	 @Botkube helm uninstall psql --random-flag
	 ```
1. Known but not supported flag
	 ```bash
	 @Botkube helm install --repo https://charts.bitnami.com/bitnami psql postgresql --wait
	 ```

## Related issue(s)

- #843


cmd/executor/helm/README.md
```diff
@@ -0,0 +1,13 @@
...
+Helm is the Botkube executor plugin that allows you to run the Helm CLI commands directly from any communication platform.
+
+## Configuration parameters
+
+The configuration should be specified in the YAML format. Such parameters are supported:
+
+```yaml
+helmDriver: "secret", # Allowed values are configmap, secret, memory.
+helmCacheDir: "/tmp/helm/.cache",
+helmConfigDir: "/tmp/helm/",
```
mszostok, line 12:
```suggestion
helmDriver: "secret" # Allowed values are configmap, secret, memory.
helmCacheDir: "/tmp/helm/.cache"
helmConfigDir: "/tmp/helm/"
```

build/Dockerfile
```diff
@@ -15,6 +15,15 @@ RUN MACH=$(uname -m); if [[ ${MACH} == "aarch64" ]]; then ARCH=arm64; \
...
     chmod +x /usr/local/bin/kubectl
+# Download the latest helm in the appropriate architecture. Currently handles aarch64 (arm64) and x86_64 (amd64).
+RUN MACH=$(uname -m); if [[ ${MACH} == "aarch64" ]]; then ARCH=arm64; \
+    elif [[ ${MACH} == "x86_64" ]]; then ARCH=amd64; \
+    elif [[ ${MACH} == "armv7l" ]]; then ARCH=arm; \
+    else echo "Unsupported arch: ${MACH}"; ARCH=${MACH}; fi; \
+    wget https://get.helm.sh/helm-v3.10.3-linux-${ARCH}.tar.gz && \
+    tar -zxvf helm-v3.10.3-linux-${ARCH}.tar.gz && \
+    mv linux-${ARCH}/helm /usr/local/bin/helm && \
+    chmod +x /usr/local/bin/helm
```
josefkarasek, line 26: I believe if you delete the `.tar.gz ` in the same `RUN` statement, it will actually be deleted - same image layer.

Or you can pipe wget into tar, something like:
```
RUN mkdir -p /usr/src/things \
    && curl -SL https://example.com/big.tar.xz \
    | tar -xJC /usr/src/things \
    && make -C /usr/src/things all
```

[docker docs](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#add-or-copy)

internal/executor/helm/commands.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+
+	return helmCmd, args, nil
+}
+
+// The go-arg library is handling the `--version` flag internally, see:
+// https://github.com/alexflint/go-arg/blob/727f8533acca70ca429dce4bfea729a6af75c3f7/parse.go#L610
+//
+// In case of Helm the `--version` flag has a different purpose, so we just remove it for now.
+func renameVersionFlag(args []string) []string {
```
josefkarasek, line 64: Command `helm --version` causes the plugin to crash.

image

josefkarasek (changes requested):
Overall good :+1: 

`--version` should not crash the plugin - only one change requested.

internal/executor/helm/commands.go
mszostok: Fixed, I also added a new uni-test to make sure it won't happen again 👍

build/Dockerfile
mszostok: Piping is sth that I would avoid here, but removing the `tar.gz` make sens 👍

mszostok:
Thanks @josefkarasek for the review 👍 

I applied changes in a single commit to make the second review easier: https://github.com/kubeshop/botkube/pull/900/commits/5b17ae90d6a16e588c2d31f43f94e920ff586f43

josefkarasek (approved):
lgtm

PR #909, opened 2023-01-02
Prepare PoC for RBAC execution support

pkosiec:
## Description

Changes proposed in this pull request:

- Prepare PoC for RBAC execution support

This PR contains all source code changes related to RBAC. For instructions, see PR https://github.com/kubeshop/botkube/pull/910.

## Video

https://user-images.githubusercontent.com/7155799/210326564-64210b6e-22d2-4200-88bb-1d6b342a404c.mp4

## Related issue(s)

#873

PR #910, opened 2023-01-02
Describe RBAC design proposal and add instruction for RBAC PoC

### LINKED ISSUE #873: RBAC Design
This task is to create a design for [Role-Based Access Control](https://github.com/kubeshop/botkube/issues/712)

**AC**

- A design for RBAC mapping between communication and K8s users/roles has been created and approved by the team and community (as applicable)
- A process for mapping communication users to K8s users/roles has been defined from the cluster admin and the users' perspectives 
- The executor authorization process (i.e. when a command is run from Botkube, how is it authorized and run from the K8s side) has been documented
- A rough implementation plan has been created and estimated
- Discuss the implementation approach with team
- See what we could use for user ID - email? Can it be used for all comm platforms?
- Ensure executor plugins take into account the mapping
- Nice to have: PoC

**Summary:** channel user -> cluster user / group mapping
**Out of scope:**
- sources

This design should be shared with the community for feedback as it is very security-sensitive.

pkosiec:
## Description

- Describe RBAC design proposal
- Add instruction for RBAC PoC

Source code changes for PoC are available in https://github.com/kubeshop/botkube/pull/909.

## Related issue(s)

Resolves #873

### DIFF
--- a/docs/investigation/rbac/README.md
+++ b/docs/investigation/rbac/README.md
@@ -112,7 +112,7 @@ You can change permissions for Roles and ClusterRoles in runtime and they will b
 
 ## Process isolation
 
-We have a full control over how a given plugin is run. That means we can use e.g. `chroot` to isolate the plugin process from the rest of the system, in order to avoid reading sensitive credentials or other Kubeconfigs with different permissions.
+We have a full control over how a given plugin is run. That means we can use e.g. `chroot` to isolate the plugin process from the rest of the system, in order to avoid reading sensitive data such as Slack credentials.
 
 I tested it successfully with a simple separate Go app on my machine. However, when I modified the code in Botkube codebase, the plugin's gRPC server exits with an error. 
 
--- a/docs/investigation/rbac/README.md
+++ b/docs/investigation/rbac/README.md
@@ -1,0 +1,119 @@
+# RBAC Proof of concept
+
+This PoC is a part of the [RBAC proposal](../../proposal/2022-12-23-rbac.md). It shows the way how we can use user-provided Roles and ClusterRoles to execute `kubectl` commands. While this PoC bases on built-in `kubectl` executor, the proposal itself describes also the API for plugins. The implementation will be very similar, but we need to make such modifications in the Default Executors instead of modifying the `kubectl` executor.
+
+## Usage
+
+1. Create `rbac-test` and `rbac-test2` channels in your Slack workspace you'll use for testing.
+1. Set up a Kubernetes cluster, e.g. with `k3d` or `colima`.
+1. Invite Botkube bot to these channels.
+1. Check out PoC pull request with GH CLI:
+
+   ```bash
+   gh pr checkout 909
+   ```
+
+1. Deploy Botkube with the following values using local chart from the PR:
+
+   ```yaml
+   image:
+     repository: kubeshop/pr/botkube
+     tag: 909-PR
+   communications:
+     "default-group":
+       socketSlack:
+         enabled: true
+         appToken: "xapp-..." # TODO: Set app token
+         botToken: "xoxb-..." # TODO: Set bot token
+         notification:
+           type: "short"
+         channels:
+           "default":
+             name: rbac-test
+             bindings:
+               executors:
+                 - kubectl-read-only
+               sources:
+                 - k8s-err-events
+           "second":
+             name: rbac-test2
+             bindings:
+               executors:
+                 - kubectl-read-only
+               sources:
+                 - k8s-err-events
+   executors:
+     kubectl-read-only:
+       kubectl:
+         enabled: true
+   settings:
+     clusterName: dev
+
+   analytics:
+     disable: true
+   ```
+
+   ```bash
+   helm install botkube --namespace botkube ./helm/botkube --wait --create-namespace -f ~/rbac-poc-values.yaml
+   ```
+
+1. Apply Role and ClusterRole resources:
+
+   ```bash
+   kubectl apply -f ./docs/investigation/rbac/assets
+   ```
+
+1. Test getting pods in the `rbac-test` channel:
+
+   ```
+   @Botkube get po
+   ```
+
+   ```
+   @Botkube get po -n botkube
+   ```
+
+   Do the same in the `rbac-test2` channel. You should get an error:
+
+   ```
+   Error from server (Forbidden): services is forbidden: User "{your-email}" cannot list resource "pods" in API group "" in the namespace "default"
+   exit status 1
+   ```
+
+1. Test getting services on the first channel:
+
+   This will work:
+
+   ```
+   @Botkube get svc
+   ```
+
+   But this command won't:
+
+   ```
+   @Botkube get svc -n botkube
+   ```
+
+   Same with other resources:
+
+   ```
+   @Botkube get ingress
+   ```
+
+1. In the second channel (`rbac-test2`), get all deployments:
+
+   ```
+   @Botkube get deploy -A
+   ```
+
+   This command won't work in the first channel.
+
+You can change permissions for Roles and ClusterRoles in runtime and they will be taken into account on the next command execution.
+
+## Process isolation
+
+We decided to come back to this topic during actual implementation.
+
+## References
+
+See the pull request [#909](https://github.com/kubeshop/botkube/pull/909) for source code changes needed for the basic RBAC PoC.
--- a/docs/investigation/rbac/README.md
+++ b/docs/investigation/rbac/README.md
@@ -7,7 +7,7 @@ This PoC is a part of the [RBAC proposal](../../proposal/2022-12-23-rbac.md). It
 1. Create `rbac-test` and `rbac-test2` channels in your Slack workspace you'll use for testing.
 1. Set up a Kubernetes cluster, e.g. with `k3d` or `colima`.
 1. Invite Botkube bot to these channels.
-1. Check out PoC pull request with GH CLI:
+1. Clone this repository and check out the PoC pull request with GH CLI:
 
    ```bash
    gh pr checkout 909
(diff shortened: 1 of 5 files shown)


docs/proposal/2022-12-23-rbac.md
```diff
@@ -0,0 +1,399 @@
...
+  'kubectl-deploy-rw':
+    botkube/kubectl@v1.0.0:
+      enabled: true
+      context:
+        kubeconfig: &kubeconfigCtx # no need to specify it again if it's the same as above
+```
+
+A few remarks:
+
+- Merging different mapping configuration won't not supported and will result in an error. The configuration will be validated during Botkube startup.
```
huseyinbabal, line 218: ... `won't be supported` ...

docs/proposal/2022-12-23-rbac.md
```diff
@@ -0,0 +1,399 @@
...
+executors:
+  'kubectl-read-only':
+    botkube/kubectl:
+      enabled: true
+      # New field "context" which will be used by Botkube Default Executor
+      context:
+        kubeconfig: &kubeconfigCtx
+
+            # Option 1: user impersonation using only email address
+            mapping:
```
huseyinbabal, line 163: - What about moving this to same indentation with `kubeconfig`? Not all of the executors may need kubeconfig, so that we can use exactly same mapping strategy for all plugins.
- `mapping` seems very broad term, what about using `rbac` or something like that?
- `context` seems good, especially isolating rbac stuff from other plugin parameters 👌

docs/proposal/2022-12-23-rbac.md
```diff
@@ -0,0 +1,399 @@
...
+  ```
+
+MS Teams:
+
+- We need to do additional API calls to get the user name email. Firstly, we need to [retrieve user ID and/or channel ID](https://learn.microsoft.com/en-us/microsoftteams/platform/bots/how-to/conversations/channel-and-group-conversations?tabs=json#retrieve-mentions), then [get user details](https://learn.microsoft.com/en-us/microsoftteams/platform/bots/how-to/get-teams-context?tabs=json#get-single-member-details).
+- There are no user groups or roles in MS Teams we can use.
+
+### Configure mapping
+
+User creates Roles and ClusterRoles, as well as the RoleBindings and ClusterRoleBindings. For example:
```
huseyinbabal, line 124: This means, Botkube Helm will be creating that automatically right?

docs/investigation/rbac/README.md
```diff
@@ -0,0 +1,149 @@
...
+
+   This command won't work in the first channel.
+
+You can change permissions for Roles and ClusterRoles in runtime and they will be taken into account on the next command execution.
+
+## Process isolation
+
+We have a full control over how a given plugin is run. That means we can use e.g. `chroot` to isolate the plugin process from the rest of the system, in order to avoid reading sensitive data such as Slack credentials.
+
+I tested it successfully with a simple separate Go app on my machine. However, when I modified the code in Botkube codebase, the plugin's gRPC server exits with an error.
```
huseyinbabal, line 117: Most probably Hashicorp's Go plugin system has system-wide dependency and fails due to chroot.

huseyinbabal (approved):
LGTM with minor comments

docs/proposal/2022-12-23-rbac.md
pkosiec: Right 😄 good catch!
```suggestion
- Merging different mapping configuration won't be supported and will result in an error. The configuration will be validated during Botkube startup.
```

docs/proposal/2022-12-23-rbac.md
pkosiec: For the default configuration yes, we will enable static mapping by default:

From consequences:

>  - Update defaults to use static mapping.
>  - Enable creating optional ClusterRole for all users to keep previous behavior.

Users (Cluster admins) will be able to create custom (cluster)role + (cluster)rolebindings per channel and that would be up to the users, unless the configuration is managed and then we can facilitate it further.

docs/investigation/rbac/README.md
pkosiec: That's true, I also suspect that 👍

docs/proposal/2022-12-23-rbac.md
pkosiec: 1. Do you mean to get rid of the second level, that is instead of `executors[name][pluginName].context.kubeconfig.nmapping` have `executors[name][pluginName].context.kubeconfig` or (`executors[name][pluginName].context.rbac`)? I think this mapping will be used only for Kubernetes cluster 🤔 Do you see any possibilities to reuse it in a different way? I would like to understand other cases, as I thought about this RBAC as something 100% related to Kubernetes, which could be wrong thinking indeed. BTW I'm totally fine to decrease the nesting level and/or rename the properties 👍 
3. You're right
4. 👌

docs/proposal/2022-12-23-rbac.md
huseyinbabal: You can freely commit your suggestion 😃

docs/proposal/2022-12-23-rbac.md
huseyinbabal: I meant something like this

```
context:
  kubeconfig: ....
  mapping: ...
``` 

Remember Mateusz showed sample plugin in one of his demos, github plugin that creates github issues. I would like to say, "the user with email foo@example.com in channel #dev can create github issues". If this is totally out of context of this proposal, then freely ignore 1. item.

pkosiec: Ah, alright, so you mean to unify the RBAC mapping config across all the plugins? Sure, I think that's great idea, as long as we are sure they will be unified 👍 Then the `context.kubeconfig` would be removed, as `mapping` is currently the only property there, so that's why there won't be a need to have something like this 👍  So yeah, let's remmove `context.kubeconfig` and move `mapping` to the top of the context, and rename it to e.g. `rbac` or something similar 👍

docs/proposal/2022-12-23-rbac.md
```diff
@@ -0,0 +1,399 @@
...
+- delete the temporary subdirectory after command execution.
+
+This is possible as we have full control over how the command is run:
+
+```go
+	for key, path := range bins {
+    // ...
+
+		//nolint:gosec // warns us about 'Subprocess launching with variable', but we are the one that created that variable.
+		cmd := exec.Command(path)
```
mszostok, line 266: I would also set the `KUBECONFIG={path}` automatically.

docs/proposal/2022-12-23-rbac.md
mszostok: I have more or less the same feedback as @huseyinbabal. The current structure is a bit too complex, but proposed simplification is ok with me 👍 

Probably later it would be good to pass the `userInfo` object to each executor.

docs/proposal/2022-12-23-rbac.md
```diff
@@ -0,0 +1,399 @@
...
+
+Once the proposal is accepted, the following changes will be made:
+
+1. Update configuration:
+
+   - Introduce `executors[name][pluginname]context` field.
+   - Support initially:
+
+     - static mapping for both users and groups,
+     - email-based user mapping,
```
mszostok, line 382: I would also push that into the follow-up task. It would be good to know if our clients want to give us a `users:read` and `users:read.email` for their workspace.

Additionally, this will be platform-specific and would require more work.

docs/proposal/2022-12-23-rbac.md
```diff
@@ -0,0 +1,399 @@
...
+
+We can support just a few options initially, and later add more based on user feedback.
+
+To get user email, channel name or user group name, we would need to do additional API calls in bot logic. However, we can cache the fetched values.
+
+#### Implementation details
+
+Slack:
+
+- We need to add two new scopes to the Slack app: `users:read` and `users:read.email`
```
mszostok, line 76: I know that for interactive messages we have the username and user email already in the request, so we can extract it and use it for a given command execution flow.

Probably the same should be also for normal messages 🤔

mszostok (approved):
LGTM 👍 

I didn't check the code, but the proposal LGTM 👍 but please see my comments. Mostly I would only suggest simplifying the proposed syntax and covering the source case too.

docs/proposal/2022-12-23-rbac.md
pkosiec: Please take a look at the PoC doc - for "normal" messages we don't have the email address, just user ID. Checked Slack, Discord, and Mattermost with verbose logging of incoming events, and for all of them this is the case. Teams docs also mention a need of additional query.

docs/proposal/2022-12-23-rbac.md
pkosiec: So for users you would suggest just static mapping? That sounds good to me, as long as Blair agrees with that 👍

docs/proposal/2022-12-23-rbac.md
pkosiec: > Probably later it would be good to pass the userInfo object to each executor.

Do you mean allowing cluster admin to define whole userInfo, including `uid` and `extra`? Not sure if there will be demand for that, but yes, this config is extendable for future, so 👍 At some point we can even extract `extra` and `uid` from user profiles if necessary, but probably that would need some argumentation and concrete use cases 😄

docs/proposal/2022-12-23-rbac.md
pkosiec: While I did that in PoC, we cannot do it here, as executor is a long-running subprocess, and we want to have a specific kubeconfig per a given execution. So the only option I see is the one described in the proposal - passing path as a part of the gRPC API. Even if we had some `envs` in the context, plugin couldn't set them they might process multiple commands at the same time and there could be a problem with using proper access rights.

docs/proposal/2022-12-23-rbac.md
```diff
@@ -0,0 +1,399 @@
...
+
+### Executing commands with proper access rights
+
+After identifying the user and/or group, instead of introspection, we could run executor as a short-living separate Pod with a dedicated ServiceAccount and (Cluster)RoleBinding. This way we could use Kubernetes RBAC to restrict access to the cluster, with the isolation provided by Kubernetes. However, this is not possible as our plugin system is not compatible:
+
+> While the plugin system is over RPC, it is currently only designed to work over a local reliable network. Plugins over a real network are not supported and will lead to unexpected behavior.
+
+**Source:** https://github.com/hashicorp/go-plugin
+
+## Consequences
```
mszostok, line 372: What about `kubectl` executor?

maybe we can finally get rid of this custom [`commands`](https://github.com/kubeshop/botkube/blob/main/helm/botkube/values.yaml#L389) property and remove our custom "policy engine"?

docs/proposal/2022-12-23-rbac.md
```diff
@@ -0,0 +1,399 @@
...
+
+- Stop using ClusterRole assigned to the Botkube pod for executing commands.
+- Implement ability to configure command execution permissions for different user groups.
+- Add ability to map users from the communication platforms to Kubernetes permissions.
+- Ensure that the proposed approach can be used with external identity providers.
+- Support plugins.
+
+### Non-Goals
+
+- Changing RBAC approach for Sources.
```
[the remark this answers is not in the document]
pkosiec: That's a very good idea, I will address it 👍

docs/proposal/2022-12-23-rbac.md
pkosiec: You're totally right, I forgot about it 👍

docs/proposal/2022-12-23-rbac.md
pkosiec: Ah, I think I understand now - so you mean getting k8s `userInfo` using  `SelfSubjectReview` API to get data about a given user and pass it to the plugin, correct? Or do you mean some other arbitrary object with data got from config and comm platform APIs? Either way, that's totally doable, but let's see first if that will be useful 👍I'll add a new sentence:

> This path may be used by the plugin to access Kubernetes cluster in a restricted way. Later we can introduce more fields to the context, such as user details.

docs/proposal/2022-12-23-rbac.md
```diff
@@ -0,0 +1,476 @@
...
+  'kubectl-deploy-rw':
+    botkube/kubectl@v1.0.0:
+      enabled: true
+      context:
+        rbac: &rbacCtx # no need to specify it again if it's the same as above
+```
+
+A few remarks:
+
+- Merging different mapping configuration won't besupported and will result in an error. The configuration will be validated during Botkube startup.
```
mszostok, line 213:
```suggestion
- Merging different mapping configuration won't be supported and will result in an error. The configuration will be validated during Botkube startup.
```

mszostok (approved):
With introduced changes, it LGTM 🚀

brampling (approved):
LGTM

PR #911, opened 2023-01-02
Add plugins configuration to make release possible

mszostok:
## Description

The Botkube release process is automated. We need to make sure that the default `values.yaml` configuration that is released for a given Helm chart is valid also for plugins.

- Add default configuration for Botkube repository index
- Add default configuration for Helm plugin

## Related issue(s)

- #843


helm/botkube/values.yaml
```diff
@@ -843,3 +856,13 @@ configWatcher:
...
+
+# -- Configuration for Botkube executors and sources plugins.
+plugins:
+  # -- Directory, where downloaded plugins are cached.
+  cacheDir: "/tmp"
+  # -- List of plugins repositories.
+  repositories:
+    # -- This repository serves officially supported Botkube plugins.
+    botkube:
+      url: https://github.com/kubeshop/botkube/releases/download/v9.99.9-dev/plugins-index.yaml
```
mszostok, line 868: The `v9.99.9-dev` tag is used to make sure that the `sed` command will override it during the release process to a given Botkube release version. 

As a follow-up, we can consider adding a CI job to do the `v9.99.9-dev` GitHub release each time a new commit is merged to `main`. (This release will contains built Botkube plugins and generated plugins index). We will need to make sure that jobs are run in sequence or the previous ones are cancelled.

josefkarasek (approved):
lgtm

Issue #891, opened 2022-12-15
Remove `list commands` and introduce `list executor/sources` instead


mszostok:
## Overview

Remove `list commands` and introduce `list executor/sources` instead

Highlights about `@Botkube list commands` issues:
- we have 2 different terms for the same thing - commands and executors . It could be misleading for our users. The `list commands` command actually returns the list of enabled executor configurations on a given channel 
![image](https://user-images.githubusercontent.com/17568639/207889799-eb254d20-6279-4e2e-8480-b3de1e245ed2.png)

   so it's actually not listing the available commands. User needs to figure it out based on a given configuration, so it’s not so useful.
- minor: the `list commands` command has hard-coded "Available kubectl commands" title which doesn’t reflect other plugins that can be enabled

## Acceptance criteria

- Remove list commands. Currently, it returns the enabled executor configuration, which is not so helpful anyway.
-  Introduce list executors and list sources commands that has output similar to filters and actions list. For example:
    ```bash
    ACTION                    ENABLED  DISPLAY NAME
    describe-created-resource false    Describe created resource
    show-logs-on-error        false    Show logs on error
    ```

PR #913, opened 2023-01-04
Replace removed 'list commands', print multi-cluster info only for supported platforms

mszostok:
## Description

- Replace removed 'list commands'
- Print multi-cluster info only for supported platforms

Here is the updated Slack help message:
   Screenshot 2023-01-04 at 11 38 32

## Related issue(s)

Fix https://github.com/kubeshop/botkube/issues/891


josefkarasek (approved):
Thanks for spotting this @mszostok!!

Issue #881, opened 2022-12-07
Missing readiness probe for GKE/Ingress integration


mirkosertic:
## Description

We want to use BotKube with MS Teams integration. We do have an existing K8s cluster running on GKE, and we want to expose the BotKube API over an existing GKE Ingress instance. However, GKE Ingress seems to require an HTTP Probe returning 200 for Ingress backend readiness check. Unfortunately, we cannot find a documentation or way to use the BotKube container for that purpose,

## Expected behavior

The BotKube container provides a readiness probe that can also be used for Ingress backend healthchecks.

## Actual behavior

There seems to be no probe in the current BotKube container image,

mszostok:
Hi @mirkosertic 

You are right. I checked and we don't have any healthcheck endpoint exposed from Botkube container.


I see two options:
- add global healthcheck e.g. `{host}:{botkube_port}/healthz` so we can also utilize that on `liveness` probe for Botkube itself.
- add a dedicated MS Teams healthcheck e.g. `{host}:{ms_teams_port}/bots/teams/heathcheck`
@pkosiec any thoughs?

We will try to incorporate that into the current release to make it available as soon as possible.

mszostok:
Summary after dev-team meeting:

- Introduce Botkube `healthz` endpoint that returns the 200 (OK). 
- Later we will maybe return JSON with information about Botkube core health
  - and enabled plugins (after https://github.com/kubeshop/botkube/issues/878)
  - and communications platform status (is it possible?)

PR #916, opened 2023-01-04
Add health endpoints

josefkarasek:
## Add health probes

New http server is started with two handlers: ready and live. Both endpoints right now always return `200 ok`.

## Related issue(s)

https://github.com/kubeshop/botkube/issues/881
https://cloud.google.com/kubernetes-engine/docs/concepts/ingress#interpreted_hc

### DIFF
--- a/cmd/botkube/main.go
+++ b/cmd/botkube/main.go
@@ -51,11 +51,13 @@ import (
 )
 
 const (
-	componentLogFieldKey = "component"
-	botLogFieldKey       = "bot"
-	sinkLogFieldKey      = "sink"
-	commGroupFieldKey    = "commGroup"
-	printAPIKeyCharCount = 3
+	componentLogFieldKey  = "component"
+	botLogFieldKey        = "bot"
+	sinkLogFieldKey       = "sink"
+	commGroupFieldKey     = "commGroup"
+	readinessEndpointName = "/healthz"
+	livenessEndpointName  = "/readyz"
+	printAPIKeyCharCount  = 3
 )
 
 func main() {
@@ -138,6 +140,12 @@ func run() error {
 	if err != nil {
 		return reportFatalError("while registering current identity", err)
 	}
+
+	probesServer := newHealthProbesServer(logger.WithField(componentLogFieldKey, "Health server"), conf.Settings.HealthPort)
+	errGroup.Go(func() error {
+		defer analytics.ReportPanicIfOccurs(logger, reporter)
+		return probesServer.Serve(ctx)
+	})
 
 	// Prometheus metrics
 	metricsSrv := newMetricsServer(logger.WithField(componentLogFieldKey, "Metrics server"), conf.Settings.MetricsPort)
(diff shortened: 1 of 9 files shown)


cmd/botkube/main.go
```diff
@@ -386,6 +394,19 @@ func newMetricsServer(log logrus.FieldLogger, metricsPort string) *httpsrv.Serve
...
 }
 
+type healthChecker struct{}
+
+func (healthChecker) ServeHTTP(resp http.ResponseWriter, req *http.Request) {
+	fmt.Fprint(resp, "ok")
+}
+
+func newHealthProbesServer(log logrus.FieldLogger, port string) *httpsrv.Server {
+	mux := mux.NewRouter()
```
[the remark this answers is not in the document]
huseyinbabal: I don't think creating a separate mux router is a good idea, what about having as follows

```
func newMetricsServer(log logrus.FieldLogger, metricsPort string) *httpsrv.Server {
	addr := fmt.Sprintf(":%s", metricsPort)
	router := mux.NewRouter()
	router.Handle("/metrics", promhttp.Handler())
        router.Handle(readinessEndpointName, healthChecker{})
        router.Handle(livenessEndpointName, healthChecker{})
	return httpsrv.New(log, addr, router)
}
```

huseyinbabal:
@josefkarasek to make this pr work, we need to add some default values to our helm chart. Otherwise, even we have health endpoint, it will not be functional

pkg/config/config_test.go
```diff
@@ -26,6 +26,7 @@ func TestLoadConfigSuccess(t *testing.T) {
 	t.Setenv("BOTKUBE_SETTINGS_CLUSTER__NAME", "cluster-name-from-env")
 	t.Setenv("BOTKUBE_SETTINGS_KUBECONFIG", "kubeconfig-from-env")
 	t.Setenv("BOTKUBE_SETTINGS_METRICS__PORT", "1313")
+	t.Setenv("BOTKUBE_SETTINGS_HEALTH__PORT", "1414")
```
huseyinbabal, line 29: Once we add health endpoint under same router, we won't need dedicated port for that

pkg/config/testdata/TestLoadConfigSuccess/config.golden.yaml
```diff
@@ -444,6 +444,7 @@ settings:
             configMap:
                 name: runtime-config
     metricsPort: "1313"
+    healthPort: "1414"
```
huseyinbabal, line 447: same as above

pkg/execute/config_test.go
```diff
@@ -60,6 +60,7 @@ func TestConfigExecutorShowConfig(t *testing.T) {
 						            fileName: ""
 						            configMap: {}
 						    metricsPort: ""
+						    healthPort: ""
```
huseyinbabal, line 63: same as above

pkg/config/config_test.go
josefkarasek: Any suggestion how to do the change?

1. Keep `BOTKUBE_SETTINGS_METRICS__PORT` and document it also covers health port configuration
2. Deprecate `BOTKUBE_SETTINGS_METRICS__PORT` and introduce more generic envar, eg `BOTKUBE_SETTINGS_OBSERV__PORT`

pkosiec: 🤔 That would be a breaking change, not sure if this is needed. Honestly I would stick to a separate server for now 👍 Yes, later we can have a single port for all the endpoints (GraphQL, metrics, health ones).

huseyinbabal (approved):

Issue #901, opened 2022-12-21
Support defining JSON Schema in source and executor plugins


mszostok:
## Overview

To enable dynamic configuration, each plugin should define the own JSON Schema.

## Acceptance Criteria

- MetadataOutput accepts required JSON Schema input
- MetadataOutput accepts optional [ui schema](https://react-jsonschema-form.readthedocs.io/en/latest/api-reference/uiSchema/)
- Generate Index Service includes new fields in generated index YAML file.
- Make sure that JSON schema input property is extensible, so in the future we can accept not only inline JSON schema but also external refs. 
  - it will allow us to put size restriction on inlined JSON Schema, so bigger will be hosted externally.
- Create initial schemas for supported plugins (Kubernetes, kubectl, Helm)
   - Kubernetes: defaults: all events?
- Ensure it is consumed by our GraphQL service

## Reason

Provide a flexible solution that allows us to enable and configure plugins without hard-coding the input parameters on end-user side.

PR #932, opened 2023-01-11
Add JSON schema to plugins

josefkarasek:
## Description

- JSON schema defined for configuration options of plugins: helm, echo, cm-watcher

## Testing

`make gen-plugins-index` - file `plugin-index.yaml` should contain JSON schema for plugins

## Related issue(s)
#901


internal/executor/helm/executor.go
```diff
@@ -150,3 +153,79 @@ func (e *Executor) handleHelmCommand(ctx context.Context, cmd command, cfg Confi
...
+						"notes": {
+							"description": "",
+							"type": "boolean"
+						}
+					}
+				}
+			},
+			"required": []
+		}`, description))
+}
```
josefkarasek, line 231: Or alternatively something like this?
```json
"command": {
    "description": "",
    "type": "string",
    "enum": ["install", "uninstall", "list", "version", "status", "test", "rollback", "upgrade", "history", "get"]
},
"subcommand": {
    "description": "",
    "type": "string",
    "enum": ["all", "manifest", "hooks", "notes"]
},
"help": {
    "description": "",
    "type": "string",
    "enum": ["help", "--help", "-h"]
},
"install": {
    "description": "",
    "type": "boolean"
},
```

pkg/api/plugin.go
```diff
@@ -20,6 +20,8 @@ type MetadataOutput struct {
 	Version string
 	// Descriptions is a description of a given plugin.
 	Description string
+	// JSONSchema is a JSON schema for a given plugin.
+	JSONSchema string
```
huseyinbabal, line 24: Can we remove this if it is not used?

pkg/api/plugin.go
```diff
@@ -33,6 +35,10 @@ func (m MetadataOutput) Validate() error {
 		issues = append(issues, "version field cannot be empty")
 	}
 
+	// if m.JSONSchema == "" {
```
huseyinbabal, line 38: Can we remove this if it is not necessary?

huseyinbabal (approved):
LGTM with minor comment

cmd/executor/echo/main.go
```diff
@@ -29,7 +34,8 @@ type EchoExecutor struct{}
 func (EchoExecutor) Metadata(context.Context) (api.MetadataOutput, error) {
 	return api.MetadataOutput{
 		Version:     version,
-		Description: "Echo is an example Botkube executor plugin used during e2e tests. It's not meant for production usage.",
+		Description: description,
+		JSONSchema:  jsonSchema(),
```
mszostok, line 38: as stated in the AC in task:
> - Make sure that JSON schema input property is extensible, so in the future we can accept not only inline JSON schema but also external refs.
>    - it will allow us to put size restriction on inlined JSON Schema, so bigger will be hosted externally.

with such approach is not possible, better is to do:
```
JSONSchema:
  Value:  "string"
```
so later we can have:
```
JSONSchema:
  RefURL:  "http://example.com/schema.json"
```

cmd/executor/echo/main.go
```diff
@@ -71,3 +77,21 @@ func main() {
 		},
 	})
 }
+
+func jsonSchema() string {
+	return heredoc.Doc(
+		fmt.Sprintf(`{
+			"$schema": "http://json-schema.org/draft-04/schema#",
```
mszostok, line 84: **nit**
```suggestion
	return heredoc.Docf`({
			"$schema": "http://json-schema.org/draft-04/schema#",
```

cmd/executor/echo/main.go
```diff
@@ -71,3 +77,21 @@ func main() {
...
+			"title": "botkube/echo",
+			"description": "%s",
+			"pluginType": "executor",
```
mszostok, line 87: this is not a valid use of JSONSchema. The `title`, `description` properties are officially supported. 

For GraphQL service, the PluginType you can take from the index YAML.

proto/executor.proto
```diff
@@ -25,6 +25,7 @@ message ExecuteResponse {
 message MetadataResponse {
 	string version = 1;
 	string description = 2;
+	string jsonschema = 3;
```
mszostok, line 28:
```suggestion
	string json_schema = 3;
```

https://developers.google.com/protocol-buffers/docs/style#message_and_field_names

proto/source.proto
```diff
@@ -27,6 +27,8 @@ message MetadataResponse {
 	string version = 1;
 	// Descriptions is a description of a given plugin.
 	string description = 2;
+	// JSONSchema is a JSON schema of a given plugin.
+	string jsonschema = 3;
```
mszostok, line 31:
```suggestion
	string json_schema = 3;
```

https://developers.google.com/protocol-buffers/docs/style#message_and_field_names

pkg/api/plugin.go
mszostok: The JSON Schema shouldn't be required, IMO. If you have some test/dev plugins, you don't always want to have it. Additionally, not always you have input parameters.

internal/plugin/index.go
```diff
@@ -50,6 +50,12 @@ type (
 		Description string     `yaml:"description"`
 		Version     string     `yaml:"version"`
 		URLs        []IndexURL `yaml:"urls"`
+		JSONSchema  JSONSchema `yaml:"JSONschema"`
```
mszostok, line 53:
```suggestion
		JSONSchema  JSONSchema `yaml:"JSONSchema"`
```

mszostok (approved):
Please adjust the `yaml` tag, besides that it LGTM 🚀

Issue #849, opened 2022-11-14
Prometheus as alternative source plugin


brampling:
A user must be able to source events from Prometheus in order to receive notifications about important cluster state issues and other alerts triggered in Prometheus.

**AC**
- Botkube can source events from Prometheus via Alertmanager
- Botkube shows important alerts (e.g. kube-state-metrics node CPU/memory alarms) collected by Prometheus/Alertmanager

**Reason**
Botkube needs to be capable of showing events from within a K8s cluster beyond those in the K8s event stream. These include alerts on node health (CPU/memory usage, etc.), alerts from kube-state-metrics, and others.

huseyinbabal:
@brampling I implemented initial working plugin of Prometheus, and need some clarifications about following item.
- I am using alert manager API, and there is no date filter there but I have a filtering mechanism on plugin side that shows only new alarms on Botkube, and does not send same alarm to Botkube twice. So, do we want to show only negative statuses or do you want to see also alarms that converted from firing -> to pending, pending -> to success?

huseyinbabal:
> @huseyinbabal Ideally we would make that configurable. Is it a lot of additional work?

I will more details then :) What kind of config will it be, list of status to be reported?

huseyinbabal:
Ok I implemented this,
![Screen Shot 2023-01-17 at 12 32 16](https://user-images.githubusercontent.com/1237982/212861493-e7164f15-e2bc-4e82-a548-9a120040a3a1.png)

PR #942, opened 2023-01-16
Prometheus source plugin

huseyinbabal:
## Description

- prometheus source plugin is added.

### Checkout branch
`gh pr checkout 942`

### Build plugins
`make build-plugins`

### Generate plugin index
`PLUGIN_DOWNLOAD_URL_BASE_PATH=http://localhost:8080/dist make gen-plugins-index`

### Run local plugin repository
`npx serve --listen 8080`

### Install Prometheus stack in k8s
```
helm repo add prometheus-community https://prometheus-community.github.io/helm-charts
helm repo update
helm install prom prometheus-community/kube-prometheus-stack
```

### Export prometheus endpoint
kubectl port-forward prometheus-botkube-kube-prometheus-st-prometheus-0 9090

### Prepare botkube config
```yaml
## /tmp/botkubecfg.yaml
communications:
  'default-group':
    # Settings for Slack
    socketSlack:
      enabled: true
      appToken: "xapp...."
      botToken: "xoxb..."
      notification:
        type: "long" # TODO: Change to `short` during testing
      channels:
        'default':
          name: botkube-demo
          bindings:
            sources:
              - plugin-based

settings:
  clusterName: gke-playground
  configWatcher: true
  lifecycleServer:
    deployment:
      name: botkube
      namespace: botkube
    port: "2113"

configWatcher:
  enabled: false
  initialSyncTimeout: 0

sources:
  'plugin-based':
    botkube/prometheus:
      enabled: true
      config:
        url: http://localhost:9090
        ignoreOldAlerts: false
        alertStates: ["pending"]

plugins:
  repositories:
    botkube:
```

### Run botkube
```
export BOTKUBE_CONFIG_PATHS="./helm/botkube/values.yaml,/tmp/botkubecfg.yaml"
export BOTKUBE_SETTINGS_KUBECONFIG="/path/to/.kube/config"

go run cmd/botkube/main.go
```

### Deploy problematic nginx
`kubectl run nginx --image nginx:1.23_err -n default`

## Related issue(s)

#849


internal/source/prometheus/client.go
```diff
@@ -0,0 +1,75 @@
...
+// NewClient initializes prometheus client
+func NewClient(url string) *Client {
+	c, _ := promClient.NewClient(promClient.Config{
+		Address: url,
+	})
+
+	newAPI := promApi.NewAPI(c)
+	return &Client{
+		Api: newAPI,
+	}
```
mszostok, line 53:
```suggestion
	return &Client{
		Api: promApi.NewAPI(c),
	}
```

internal/source/prometheus/config.go
```diff
@@ -0,0 +1,43 @@
...
+)
+
+// Config prometheus configuration
+type Config struct {
+	URL             string               `yaml:"url"`
+	AlertStates     []promApi.AlertState `yaml:"alertStates"`
+	IgnoreOldAlerts *bool                `yaml:"ignoreOldAlerts"`
+}
+
+func MergeConfigs(configs []*source.Config) (Config, error) {
```
mszostok, line 18:
```suggestion
// MergeConfigs merges all input configuration.
func MergeConfigs(configs []*source.Config) (Config, error) {
```

internal/source/prometheus/config.go
```diff
@@ -0,0 +1,43 @@
...
+	out := Config{
+		URL:             "http://localhost:9090",
+		AlertStates:     []promApi.AlertState{promApi.AlertStateFiring, promApi.AlertStatePending, promApi.AlertStateInactive},
+		IgnoreOldAlerts: ptr.Bool(true),
+	}
+	for _, rawCfg := range configs {
+		var cfg Config
+		err := yaml.Unmarshal(rawCfg.RawYAML, &cfg)
+		if err != nil {
+			return Config{}, err
```
mszostok, line 28:
```suggestion
			return Config{}, fmt.Errorf("while unmarshalling YAML config: %w", err)
```

internal/source/prometheus/config.go
```diff
@@ -0,0 +1,43 @@
...
+		}
+
+		if cfg.URL != "" {
+			out.URL = cfg.URL
+		}
+		if len(cfg.AlertStates) > 0 {
+			out.AlertStates = cfg.AlertStates
```
mszostok, line 35: WDYT about appending them instead of resetting 🤔 ? 
you can have a `map[string]struct{}` to make sure that you have only unique `AlertStates`.

In general, I'm fine with both approaches, but I'm curious why you picked overriding instead of merging 🙂

internal/source/prometheus/source.go
```diff
@@ -0,0 +1,75 @@
...
+
+// Source prometheus source plugin data structure
+type Source struct {
+	prometheus    *Client
+	l             sync.Mutex
+	pluginVersion string
+	startedAt     time.Time
+}
+
+func NewSource(version string) *Source {
```
mszostok, line 26:
```suggestion
// NewSource returns a new instance of Source.
func NewSource(version string) *Source {
```

Just for the sake of having comments on public methods 🙂

internal/source/prometheus/source.go
```diff
@@ -0,0 +1,75 @@
...
+	config, err := MergeConfigs(input.Configs)
+	if err != nil {
+		return source.StreamOutput{}, fmt.Errorf("while merging input configs: %w", err)
+	}
+	go p.consumeAlerts(ctx, config, out.Output)
+	return out, nil
+}
+
+// Metadata returns metadata of prometheus configuration
+func (p *Source) Metadata(_ context.Context) (api.MetadataOutput, error) {
```
mszostok, line 45: Please define the `JSONSchema` field.

internal/source/prometheus/source.go
```diff
@@ -0,0 +1,75 @@
...
+	return api.MetadataOutput{
+		Version:     p.pluginVersion,
+		Description: "Prometheus polls alerts from Prometheus alert manager to report in Botkube",
```
mszostok, line 48:
```suggestion
		Description: "Prometheus plugin polls alerts from configured Prometheus Alertmanager.",
```

From what I see in the official doc, it should be written as `Alertmanager` instead of `alert manager`.

internal/source/prometheus/source.go
```diff
@@ -0,0 +1,75 @@
...
+	}, nil
+}
+
+func (p *Source) consumeAlerts(ctx context.Context, config Config, ch chan<- []byte) {
+	prometheus := p.getPrometheusClient(config.URL)
+	for {
+		alerts, _ := prometheus.Alerts(ctx, GetAlertsRequest{
```
mszostok, line 55: Please log that error, so it will be visible when the `debug` mode will be enabled.

internal/source/prometheus/source.go
```diff
@@ -0,0 +1,75 @@
...
+			IgnoreOldAlerts: *config.IgnoreOldAlerts,
+			MinAlertTime:    p.startedAt,
+			AlertStates:     config.AlertStates,
+		})
+		for _, alert := range alerts {
+			msg := fmt.Sprintf("[%s][%s][%s] %s", PluginName, alert.Labels["alertname"], alert.State, alert.Annotations["description"])
+			ch <- []byte(msg)
+		}
+		time.Sleep(time.Second * 5)
```
mszostok, line 64: I will appreciate a comment explaining that sleep call. Thanks!

internal/source/prometheus/source.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+}
+
+func (p *Source) getPrometheusClient(url string) *Client {
+	p.l.Lock()
+	defer p.l.Unlock()
+	if p.prometheus == nil {
+		return NewClient(url)
+	}
+	return p.prometheus
```
mszostok, line 74:
```suggestion
	if p.prometheus == nil {
		p.prometheus = NewClient(url)
	}
	return p.prometheus
```

Otherwise, it doesn't make sens 🤔

internal/source/prometheus/config.go
huseyinbabal: Because, think about we have defaults "pending", "firing" in our values yaml. Then customer wants to see only "inactive". In this case, if I append instead of reset, then we will be showing alerts that have state that customer does not want.

internal/source/prometheus/source.go
huseyinbabal: Good catch 👌

internal/source/prometheus/source.go
```diff
@@ -0,0 +1,146 @@
...
+		Description: description,
+		JSONSchema:  jsonSchema(),
+	}, nil
+}
+
+func (p *Source) consumeAlerts(ctx context.Context, config Config, ch chan<- []byte) {
+	log := loggerx.New(loggerx.Config{
+		Level: config.Log.Level,
+	})
+	prometheus, err := p.getPrometheusClient(config.URL)
```
mszostok, line 66: **nit**

we can move it to `Stream` method, same as you did with `MergeConfigs`, so you will return an error there

```go
	prometheus, err := p.getPrometheusClient(config.URL)
	if err != nil {
		return source.StreamOutput{}, fmt.Errorf("while creating Prometheus client: %w", err)
	}
```

or move that into `for` loop, and just do the `log.Errorf` only, so client creation will be "retried" in case of error.

mszostok (approved):
LGTM 🚀

internal/source/prometheus/source.go
huseyinbabal: ok I kept it in consumeAlerts function, and removed mutex

PR #944, opened 2023-01-18
Upload plugins to GCS

### LINKED ISSUE #943: Public hosting of latest plugin binaries & index
## Overview

Currently plugin binaries and plugin index are released together with Botkube.
This is inflexible for consumers of individual plugins, which receive changes in different
cadence than Botkube.

To split these two, a new place for plugin binaries and index has to be created together
with automation that will publish latest build.

The underlying place for the plugins will be a public GCS bucket and the automation
will be handled by github actions.

## Acceptance Criteria

- [x] GCS bucket for plugins and index is created and configured
- [x] github action that moves plugins and index to the GCS bucket is created
- [x] new plugin and index release is made public

## Reason

1. decoupling of Botkube and individual plugin release cycles
2. easier testing and development - no more need for running local plugin provider

josefkarasek:
## Description

- Github action that builds and uploads plugins to GCP

Resolves #943

### DIFF
--- a/.github/workflows/branch-build.yml
+++ b/.github/workflows/branch-build.yml
@@ -115,7 +115,7 @@ jobs:
       - name: Build all plugins into dist directory
         env:
           # we hardcode plugins version, so it's predictable in e2e tests
-          GORELEASER_CURRENT_TAG: "v1.0.0-devel"
+          GORELEASER_CURRENT_TAG: "v0.0.0-latest"
         run: |
           make build-plugins
 
--- a/.github/workflows/pr-build.yaml
+++ b/.github/workflows/pr-build.yaml
@@ -207,7 +207,7 @@ jobs:
       - name: Build all plugins into dist directory
         env:
           # we hardcode plugins version, so it's predictable in e2e tests
-          GORELEASER_CURRENT_TAG: "v1.0.0-devel"
+          GORELEASER_CURRENT_TAG: v0.0.0-latest"
         run: |
           make build-plugins
 
--- a/.github/workflows/upload-plugins.yaml
+++ b/.github/workflows/upload-plugins.yaml
@@ -20,6 +20,8 @@ jobs:
       uses: 'google-github-actions/auth@v1'
       with:
         credentials_json: ${{ secrets.PLUGINS_BUCKET_CREDENTIALS }}
+    - name: 'Set up Cloud SDK'
+      uses: 'google-github-actions/setup-gcloud@v1'
     - name: Set up QEMU
       uses: docker/setup-qemu-action@v1
     - name: Setup Go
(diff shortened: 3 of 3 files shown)


pkosiec:
Hey @josefkarasek, please post the links to sample runs of this Job on your fork (as a proof this works as expected). Thanks!

josefkarasek:
Workflow here: https://github.com/kubeshop/botkube/actions/workflows/test-push-plugins.yaml

Plugins here: https://console.cloud.google.com/storage/browser/botkube-plugins-latest

.github/workflows/test-push-plugins.yaml
```diff
@@ -0,0 +1,54 @@
...
+    branches:
+    - main
+    paths:
+    - 'cmd/executor/**'
+    - 'cmd/source/**'
+jobs:
+  publish_plugins:
+    permissions:
+      contents: 'read'
+      id-token: 'write'
```
pkosiec, line 13: Is this needed? 🤔

.github/workflows/test-push-plugins.yaml
```diff
@@ -0,0 +1,54 @@
...
+        go-version-file: 'go.mod'
+        cache: true
+    - name: Install GoReleaser
+      uses: goreleaser/goreleaser-action@v2
+      with:
+        install-only: true
+        version: latest
+    - name: Build all plugins into dist directory
+      env:
+        GORELEASER_CURRENT_TAG: "v1.0.0-devel"
```
pkosiec, line 41: 🤔 I know we used this for integration tests, but this is different case, because the version is public. Can we change this `v9.99.9-dev`? It's not that I like `v9.99.9-dev` tag, but just for consistency reason it would be better than 1.0.0. Somebody can think we're going to release 1.0.0 soon 😄 

Other ideas: `v0.0.0-latest`, `v0.0.0-dev`... but probably consistency is better.

What do you think?

PS you can also change the version for the integration tests.

pkosiec (commented):
Generally LGTM, good work! 👍 Just minor comments.

Also, I know we're also building these plugins for integration test purposes, so theoretically we could extend that workflow, but building plugins is fast and I think a separate workflow like you did is much easier to maintain 👍

.github/workflows/test-push-plugins.yaml
josefkarasek: I there any place in the test where the version is also hard-coded?

pkosiec: Yes, branch build, but you already changed 👍

pkosiec (approved):
Before merge, can you run this job once again just to make sure it works properly? I think the `content: read` permission is necessary (I wanted to remove `id_token`: write). But let's make sure it works before merge 👍 You can post a link to successful job here. Thanks!

PR #948, opened 2023-01-19
Support Help() in plugins

### LINKED ISSUE #893: Provide an option to get plugin related help
## Overview

Introduce the `@Botkube {executor_name} help` command, e.g. `@Botkube kubectl help`. It returns the syntax of a given executor e.g. similar output as `kubectl help`. However, it should be adjusted, for example we don't support all verbs or flags, so we should strip them before showing the help message to the user. It’s a plugin responsibility to return a proper help message. 

## Acceptance Criteria

-  provide an option for our end-users to get plugin related help. 
  - it should contain information what is possible with a given executor and what is the syntax of it.

## Reason

Unification of showing help for all plugins

## Notes
- extend gRPC API with a dedicated method for getting Help
    - Return help as the `interactive.Message` (plugin needs to return it)
    - accept empty message too for some internal/dev/testing tools (present the "cricket" message to users)

josefkarasek:
## Provide an option to get plugin related help #893

- Add `Help()` function to the executor interface

## Test

1. Rebuild plugins and plugin index `make gen-plugins-index`
2. Delete local plugin cache, for example `rm -rf /tmp/plugin*`
3. Run local plugin server `env PLUGIN_SERVER_HOST=http://localhost go run test/helpers/plugin_server.go`
4. Reference this plugin server in your config file
5. Enable executors
```
executors:
  'echo':
    botkube/echo:
      enabled: true
  'helm':
    botkube/helm:
      enabled: true
  'gh':
    botkube/gh:
      enabled: true
```
7. Run botkube locally: `go run cmd/botkube/main.go`
8. See help output for plugins `gh`, `helm`, `echo`

## Related issue(s)

Resolves #893


internal/executor/helm/executor.go
```diff
@@ -131,6 +131,12 @@ func (e *Executor) Execute(ctx context.Context, in executor.ExecuteInput) (execu
 	}
 }
 
+func (Executor) Help() executor.HelpResponse {
+	return executor.HelpResponse{
+		Help: (&HelpCommand{}).Help(),
```
josefkarasek, line 136: What should happen with existing Helm plugin `help` sub-command?

internal/executor/helm/executor.go
```diff
@@ -131,6 +132,16 @@ func (e *Executor) Execute(ctx context.Context, in executor.ExecuteInput) (execu
 	}
 }
 
+func (Executor) Help(ctx context.Context) (interactive.Message, error) {
+	return interactive.Message{
+		Base: interactive.Base{
+			Body: interactive.Body{
+				CodeBlock: (&HelpCommand{}).Help(),
```
pkosiec, line 139: The `HelpCommand` struct is no longer needed if you'll get rid of the unused logic. So then you can simplify it - either keep `Help() string` as a simple function, or embed it directly here in the codeblock - both option work for me 👍

pkg/execute/executor.go
```diff
@@ -169,6 +169,14 @@ func (e *DefaultExecutor) Execute(ctx context.Context) interactive.Message {
...
 
 	if isPluginCmd {
+		if isHelpCmd(cmdCtx.Args) {
+			msg, err := e.pluginExecutor.Help(ctx, e.conversation.ExecutorBindings, cmdCtx.Args, cmdCtx.CleanCmd)
+			if err != nil {
+				e.log.Errorf("while executing help command %q: %s", cmdCtx.CleanCmd, err.Error())
+				return empty
+			}
+			return msg
+		}
```
pkosiec, line 179: 1. We should report the help command usage as well
2. Once you add the command reporting, then I think the code can look even more complicated - consider extracting this help usage to a separate small method.

pkosiec (commented):
Here are some comments from the code review. I'll test it after fixes 🙂 

You can also provide some small instruction in the PR desc how did you test it (executor with help like helm, or the one with empty help, like echo).

internal/executor/helm/executor.go
```diff
@@ -9,6 +9,7 @@ import (
```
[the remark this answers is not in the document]
pkosiec: This is still valid comment - look for `case arg.ErrHelp:` or for usage for `Help()`:
image

cmd/executor/gh/main.go
```diff
@@ -98,6 +99,11 @@ func (e *GHExecutor) Execute(ctx context.Context, in executor.ExecuteInput) (exe
 	}, nil
 }
 
+// Help returns help message
+func (*GHExecutor) Help(ctx context.Context) (interactive.Message, error) {
+	return interactive.Message{}, nil
```
pkosiec, line 104: Please add help based on https://blog-post.botkube-docs.pages.dev/blog/custom-executors/

image

This is a low-hanging fruit, so.. 🙂 


```
Usage: `gh create issue KIND/NAME [-n, --namespace]`
```
Will be totally fine 👍

test/e2e/bots_test.go
```diff
@@ -264,6 +264,19 @@ func runBotTest(t *testing.T,
...
+			expectedMessage := ".... empty response _*&lt;cricket sounds&gt;*_ :cricket: :cricket: :cricket:"
+			if botDriver.Type() == DiscordBot {
+				expectedMessage = ".... empty response _*<cricket sounds>*_ :cricket: :cricket: :cricket:"
+			}
+
+			botDriver.PostMessageToBot(t, botDriver.Channel().Identifier(), command)
+			err := botDriver.WaitForLastMessageEqual(botDriver.BotUserID(), botDriver.Channel().ID(), expectedMessage)
+			assert.NoError(t, err)
+		})
+
```
pkosiec, line 279: WDYT about adding also `helm help` to the E2E tests? In that way we'd test both cases (empty help + provided help)  🤔

pkosiec (commented):
In general LGTM, just a final batch of small comments 👍

internal/executor/helm/executor.go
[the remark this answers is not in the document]
josefkarasek: Can you check you have the latest code? I removed `HelpCommand.Help()`. It should not be there at all.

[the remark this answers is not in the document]
pkosiec: You're right - I'm sorry, that was an outdated screenshot. I forgot that help still needs to be handled for underlying commands. All good 👍

pkosiec (approved):
LGTM 👍 E2E tests are failing because of the formatting issue of the Helm help, but basically everything works well and all comments were addressed 👌 Thanks!

PR #950, opened 2023-01-23
Support fetching plugin dependencies

pkosiec:
## Description

- Support fetching plugin dependencies
  - Download plugins and its dependencies with `go-getter`
  - Pass `PLUGIN_DEPENDENCY_DIR` env variable to the plugins to use fetched dependencies
  - Fix plugin index validation
  - Fix inconsistencies in gRPC API field naming

## Testing

No need to test it manually - the current integration tests prove the functionality works as expected. Helm binary is no longer there in the Botkube Docker image, but the integration tests still pass.

However, if you'd like to, then follow the tutorial here: https://docs.botkube.io/plugin/local-testing

I used the following file as the `/tmp/config-values.yaml`:

```yaml
plugins:
  repositories:
    botkube:

communications:
  "default-group":
    socketSlack:
      enabled: true
      appToken: "xapp-..."
      botToken: "xoxb-..."
      notification:
        type: "short"
      channels:
        "default":
          name: botkube-demo
          bindings:
            executors:

              - helm
            sources: []

executors:
  helm:
    botkube/helm:
      config:
        helmCacheDir: /tmp/helm/.cache
        helmConfigDir: /tmp/helm/
        helmDriver: secret
      enabled: true
```

In `botkube` repo, run:
```bash
npx serve --listen 8080
```

Then in a new terminal run:
```bash
export BOTKUBE_CONFIG_PATHS="$(pwd)/resource_config.yaml,/tmp/config-values.yaml"
PLUGIN_DOWNLOAD_URL_BASE_PATH="http://localhost:8080/plugin-dist" make gen-plugins-index; rm -rf /tmp/plugins; go run cmd/botkube/main.go
```

I hope you have different Helm version installed on your system - if not, upgrade it locally to anything different than v3.6.3 (e.g. 3.10+) 😄 
Once Botkube is run, run `@Botkube helm version` on the `botkube-demo` channel to see the Helm version used. The output should be equal to:

image

## Related PRs:

https://github.com/kubeshop/botkube-docs/pull/203

## Related issue(s)

#899


internal/plugin/index.go
```diff
@@ -101,12 +121,11 @@ func (in Index) Validate() error {
 		if entry.Type != "" && entry.Name != "" && entry.Version != "" {
 			key := strings.Join([]string{string(entry.Type), entry.Name, entry.Version}, ";")
 			firstEntryIdx, found := entriesByKey[key]
-			if !found {
-				entriesByKey[key] = currentIdx
-				continue
+			if found {
```
mszostok, line 124: **nit** hm.. so the logic is not consistent with the approach in 110 line, tbh having `if found` and then don't have the `continue` was a bit misleading for me as I was thinking why on error we want to proceed further, and then I realized that there is a map, so it's not a problem at the end however was not readable for me at first look.

internal/plugin/index_test.go
```diff
@@ -27,14 +29,18 @@ func TestValidateIndexErrors(t *testing.T) {
 			* entries[6]: 1 error occurred:
 				* field type cannot be empty
 			* entries[7]: 1 error occurred:
-				* field name cannot be empty`)
+				* field name cannot be empty
+			* entries[8]: 1 error occurred:
+				* field type is not valid, allowed values are [source executor]
+			* entries[9]: 1 error occurred:
+				* dependency URL for key "kubectl" in the URL entry "https://github.com/kubeshop/botkube/releases/download/v0.1.0/linux-_arm64_source_kubernetes" cannot be empty`)
```
mszostok, line 36: maybe we should use the `platform` and `arch` properties instead of URL? WDYT? 

as the URL in some edge cases can be the same (https://en.wikipedia.org/wiki/Fat_binary)

pkg/api/plugin.go
```diff
@@ -33,6 +36,19 @@ type JSONSchema struct {
...
+	URLs URLs `yaml:"urls"`
+}
+
+type URLs map[string]string
+
+func (u URLs) For(os, arch string) (string, bool) {
+	val, exists := u[fmt.Sprintf("%s/%s", os, arch)]
+	return val, exists
+}
+
```
mszostok, line 51: please add comment

proto/source.proto
```diff
@@ -21,21 +21,27 @@ message StreamResponse {
...
+	// json_schema is a JSON schema of a given plugin.
 	JSONSchema json_schema = 3;
+	// dependencies is a list of dependencies of a given plugin.
+	map<string, Dependency> dependencies = 4;
 }
 
 message JSONSchema {
 	// value is the string value of the JSON schema.
 	string value = 1;
 	// refURL is the remote reference of the JSON schema.
```
mszostok, line 38:
```suggestion
	// ref_url is the remote reference of the JSON schema.
```

internal/plugin/manager.go
```diff
@@ -372,57 +369,127 @@ func createGRPCClients[C any](logger logrus.FieldLogger, bins map[string]string,
...
+		if err != nil {
+			return fmt.Errorf("while downloading dependency from URL %q: %w", url, err)
+		}
 	}
 
-	m.log.WithFields(logrus.Fields{
-		"url":     url,
+	// Ensure all dependencies are downloaded
+	log.WithFields(logrus.Fields{
 		"binPath": binPath,
```
mszostok, line 417: **nit** it's already defined at line 390

mszostok (approved):
I have only minor comments, overall it LGTM

internal/executor/helm/bin_runner.go
```diff
@@ -8,9 +8,19 @@ import (
 )
 
 func runHelmCLIBinary(ctx context.Context, cfg Config, args []string) (string, error) {
-	cmd := exec.CommandContext(ctx, helmBinaryName, args...)
-	cmd.Env = os.Environ()
-	cmd.Env = append(cmd.Env,
+	// Use full path if the PLUGIN_DEPENDENCY_DIR env variable is found.
+	// If not, Go will get $PATH for current process and try to look up the binary.
+	//
+	// Unfortunately, we cannot override PATH env variable for a go-plugin process as Hashicorp overrides envs during plugin startup.
```
mszostok, line 14:
```suggestion
	// Unfortunately, we cannot override PATH env variable for a go-plugin process as Hashicorp overrides envs during plugin startup.
```

If you know the internal code then I can understand this comment, but without that it's strange that we cannot override env PATH as go-plugin overrides envs during startup, but somehow the `PLUGIN_DEPENDENCY_DIR` is available 🤔  😄 

would be good to rephrase it

internal/plugin/index.go
pkosiec: Yes, the problem is that because of the consistent approach there was a bug 😄 we need to assign this in both cases:
```go
entriesByKey[key] = currentIdx
``` 

I hope this is clearer now:

```go

		if entry.Type != "" && entry.Name != "" && entry.Version != "" {
			// check if we have a duplicate entry
			key := strings.Join([]string{string(entry.Type), entry.Name, entry.Version}, ";")
			firstEntryIdx, alreadyExist := entriesByKey[key]
			if alreadyExist {
				// duplicate, return error
				entryIssues = multierror.Append(entryIssues, fmt.Errorf("conflicts with the %s entry as both have the same type, name, and version", humanize.Ordinal(firstEntryIdx)))
			}
			
			entriesByKey[key] = currentIdx
		}
```

There are two alternative options (copying the assignment, or moving it to the top, before checking `if found`), but honestly I don't like them, so hope the comments and var name change improves the readability

internal/executor/helm/bin_runner.go
pkosiec: Right 😄 Will do!

internal/plugin/index.go
pkosiec: Yes, the problem is that because of the consistent approach there was a bug 😄 we need to assign this in both cases:
```go
entriesByKey[key] = currentIdx
``` 

If you go back to the previous form, then when running the index_test.go, you'll get 5 errors instead of 8, as these lines are skipped:

```go
err := entryIssues.ErrorOrNil()
		if err != nil {
			issues = multierror.Append(issues, fmt.Errorf("entries[%d]: %s", currentIdx, stringx.IndentAfterLine(err.Error(), 1, "\t")))
		}
```

I hope this is clearer now:

```go

		if entry.Type != "" && entry.Name != "" && entry.Version != "" {
			// check if we have a duplicate entry
			key := strings.Join([]string{string(entry.Type), entry.Name, entry.Version}, ";")
			firstEntryIdx, alreadyExist := entriesByKey[key]
			if alreadyExist {
				// duplicate, return error
				entryIssues = multierror.Append(entryIssues, fmt.Errorf("conflicts with the %s entry as both have the same type, name, and version", humanize.Ordinal(firstEntryIdx)))
			}
			
			entriesByKey[key] = currentIdx
		}
```

There are two alternative options (copying the assignment, or moving it to the top, before checking `if found`), but honestly I don't like them, so hope the comments and var name change improves the readability

Issue #925, opened 2023-01-09
Create getting started with plugins blog post


brampling:
### Overview
Write and post a blog post that provides a "getting started" guide to creating a new Botkube plugin.

### Acceptance Criteria

- Blog post created
  - Contains a brief overview of how the plugin system works
  - Guides the user on how to create and publish a new plugin (on external repository)
- Blog post published
- Post shared on social media

## Notes

- Figure out the idea for a nice executor
- Focus on executor plugin
- GitHub repo template for external executor
   - Something what @mszostok already did https://github.com/mszostok/botkube-plugins-template - maybe we should transfer this repo?

PR #951, opened 2023-01-23
Add plugin helper functions

mszostok:
## Description

- Add plugin helper functions

## Related issue(s)

#925


pkg/pluginx/command.go
```diff
@@ -0,0 +1,66 @@
...
+	envs, args, err := shellwords.ParseWithEnvs(rawCmd)
+	if err != nil {
+		return "", err
+	}
+
+	if len(args) < 1 {
+		return "", fmt.Errorf("invalid raw command: %q", rawCmd)
+	}
+
+	if err := os.Setenv("PATH", fmt.Sprintf(`%s:%s`, pluginsDir, os.Getenv("PATH"))); err != nil {
```
pkosiec, line 47: `pluginsDir` is hardcoded to `/tmp/plugins`, but the `cacheDir` for plugin manager is configurable, so we cannot do it here. Should be passed from top or ignored.

pkosiec, line 47: I don't like modifying `PATH` - remember that this will be modified also for other executors  like the built-in `kubectl` one. Also, the change is long-living, and for each occurrence you add the same path over and over again.

pkosiec (commented):
The `ParseCommand` + `MergeConfigs` helpers could be useful, but I think the other ones don't belong here. They doesn't seem to be generic enough, and also use some shortucts like harcoded paths or modifying envs for current process. The `DownloadDependencies` will be not necessary anymore on Feburary 6th so I wouldn't keep it as official helper.

I think the solutions you provided are totally fine for the example executor purpose, so let's move most of the code there. In `pluginx` I would put something we actually use for our plugins.

We can discuss it offline, but please see my comments first 👍

pkg/pluginx/bin.go
```diff
@@ -0,0 +1,78 @@
...
+//	   			"darwin/amd64": "https://github.com/cli/cli/releases/download/v2.21.2/gh_2.21.2_macOS_amd64.tar.gz//gh_2.21.2_macOS_amd64/bin",
+//	   			"linux/amd64":  "https://github.com/cli/cli/releases/download/v2.21.2/gh_2.21.2_linux_amd64.tar.gz//gh_2.21.2_linux_amd64/bin",
+//	   		},
+//	   	},
+//	   }
+//
+//	however, instead of calling download method:
+//	  pluginx.DownloadDependencies(depsDownloadLinks)
+//
+//	move the `depsDownloadLinks` property under the MetadataOutput object:
```
mszostok, line 40:
```suggestion
//	move the `depsDownloadLinks` object under `MetadataOutput`:
```

mszostok:
I applied all fixed discussed offline. Once this PR will be merged, I will update our tutorials

pkg/pluginx/command.go
```diff
@@ -0,0 +1,118 @@
+package pluginx
```
pkosiec, line 1: General comment: not sure what about error wrapping (if we used Go-native wrapping, then it would be easy to get the root error without an external library), personally I would add it at some point. But I think it's fine how it is for now 👍

pkosiec (approved):
LGTM 👌 Good work!

PR #952, opened 2023-01-24
Fix Prometheus source JSON schema and generated resources CI job

pkosiec:
## Description

- Fix Prometheus source JSON schema
- Fix `check-generated-resources` job. Changes will be applied after merge.

## Testing

Checkout this PR and run:
```bash
PLUGIN_DOWNLOAD_URL_BASE_PATH="http://localhost:8080/plugin-dist" make gen-plugins-index
```

Then verify the output index file to see if the schema is valid. You can use https://www.jsonschemavalidator.net/ to validate both previous and current (from this PR) schemas.

### DIFF
--- a/.github/workflows/pr-build.yaml
+++ b/.github/workflows/pr-build.yaml
@@ -126,7 +126,8 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
         with:
-          fetch-depth: 0
+          ref: ${{ github.event.pull_request.head.sha }}
+          persist-credentials: false
 
       - name: Run generating gRPC resources
         run: make gen-grpc-resources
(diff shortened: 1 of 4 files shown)


pkosiec:
I'm not sure why I'm getting error https://github.com/kubeshop/botkube/actions/runs/3997391552/jobs/6858659306 if I already updated the files. And why I didn't get such error earlier when I updated gRPC API? Is it because this is `pull_request_target`? 🤔 

Oh right...

```yaml
  check-generated-resources:
    runs-on: ubuntu-latest
    steps:
      - name: Checkout
        uses: actions/checkout@v3
        with:
          fetch-depth: 0
```

Will fix that here as well.

Issue #908, opened 2022-12-23
Move context setup to the beginning of run method


huseyinbabal:
If we move context setup upside from https://github.com/kubeshop/botkube/blob/3aa76fade5ba174b6829942a0bc095d46eb21558/cmd/botkube/main.go#L105, we can pass existing context to downstream objects like analytics, config loader etc

Issue #870, opened 2022-11-29
Add support for the `aliases` property for executor plugins


mszostok:
## Overview

Currently, the aliases are built-in into Botkube core logic. We should support setting them in the YAML configuration too:
```yaml
aliases: # shortcuts
  kc: 
     command: kubectl
     displayName: "...." # optional field
  kgp:
     command: kubectl get pod
````

With such configuration, the`kubectl` executor plugin can be used with the following command prefixes: `kubectl`, `kc`. Additionally, `kgp` resolves to the whole command that was defined.

## Acceptance Criteria

- end user can define executor aliases
- update documentation
- add e2e test
- list possible aliases under `list executors`

## Reason

Simplify typing plugin executor in chat window.

**Comment by mszostok** [comment]
Double check before doing the implementation whether we should 
- allowing to set the alias for a whole command definition e.g. `kcp=kubectl get pod`
- having them global per cluster

**Comment by mszostok** [comment]
Questions:
- should we also support `displayName`/`description` to support less experienced k8s users. Based on e.g dispaly Name they could figure it out what a given command does and know that they should run it.

- should we support global aliases, or aliases per cluster?

**Global aliases:**
```yaml
executors:
  'team-a-kubectl':
    botkube/kubectl@v1.0.0:
      enabled: true
      config: ... 
         

  'team-b-kubectl':
    botkube/kubectl@v1.0.0:
      enabled: true
      config: ...

communications:
  default-group:
    socketSlack:
      enabled: true
      channels:
        all:
          name: all
          bindings:
            executors:
              - plugin-kubectl 

aliases: # shortcuts
  kc: kubectl
  kgp: kubectl get pod
```
**Executor aliases:**

```yaml
executors:
  'team-a-kubectl':
    botkube/kubectl@v1.0.0:
      enabled: true
      config: ...     
      aliases:  &aliases
         kc: kubectl # displayName/description?
         kcg: kubectl get pod
         

  'team-b-kubectl':
    botkube/kubectl@v1.0.0:
      enabled: true
      config: ...
      aliases: *aliases

communications:
  default-group:
    socketSlack:
      enabled: true
      channels:
        all:
          name: all
          bindings:
            executors:
              - plugin-kubectl 
```

PR #959, opened 2023-02-01
Validate alias commands and display builtin command aliases

pkosiec:
## Description

- Validate alias command prefixes
- Display builtin command aliases
- Use proper context for querying Botkube config from GraphQL and editing SourceBindings

## Screenshot

image

## Testing

No need to test it manually as the functionality was covered by unit and integration tests.

## Related issue(s)

Resolves https://github.com/kubeshop/botkube/issues/870
Resolves https://github.com/kubeshop/botkube/issues/908

### DIFF
--- a/cmd/botkube/main.go
+++ b/cmd/botkube/main.go
@@ -67,10 +67,15 @@ func main() {
 
 // run wraps the main logic of the app to be able to properly clean up resources via deferred calls.
 func run() error {
+	// Set up context
+	ctx := signals.SetupSignalHandler()
+	ctx, cancelCtxFn := context.WithCancel(ctx)
+	defer cancelCtxFn()
+
 	// Load configuration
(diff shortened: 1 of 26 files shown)


cmd/botkube/main.go
```diff
@@ -67,10 +67,15 @@ func main() {
 
 // run wraps the main logic of the app to be able to properly clean up resources via deferred calls.
 func run() error {
+	// Set up context
```
mszostok, line 70: If that's OK with you, I would like to have the `run(ctx)` and those 3 commands moved to `main` 🙂

mszostok (approved):
🚀

cmd/botkube/main.go
pkosiec: That works for me 👍

Issue #933, opened 2023-01-11
Update Botkube configuration with RBAC-related properties


pkosiec:
## Acceptance Criteria

Update Botkube configuration:

   - Introduce `executors[name][pluginname]context` and `source[name][pluginname]context` fields.
   - Initially support for executors:

     - static mapping for both users and groups,
     - channel-based group mapping.

     Create a follow-up task to support additional mappings and monitor the community demand for it:

     - email-based and username-based user mapping,
     - user-group-based group mapping.

   - Add ability to set default namespace as a part of plugin config (this is missing from the proposal)
      - Suggestion: `context.kubeconfig.defaultNamespace`
   - For sources, initially support static user and group mapping. Nice to have: channel-based group mapping if the implementation won't be hard.
   - Validate executor and source bindings configuration and return error if kubeconfig RBAC mapping is different for the same bound plugins.
   - To keep the previous behavior:
      -  Create ClusterRole and ClusterRoleBinding for `botkube-plugins-read-only`
          - provide an option to enable/disable this creation
      - Consolidate the `kubectl-read-only` and `helm` group under a single group, e.g. `botkube-read-only`  and for all executors under this group use the static group mapping to `botkube-plugins-read-only`.

   - Allow restricting access to specific groups and/or users for impersonation in Helm chart.
   - Make sure that `rbac` property is optional
      - If the rbac context is not provided (as a result, `rbac.user`|`group.type` is empty), Botkube won't create a temporary Kubeconfig for a given plugin.

See the RBAC proposal: [link](https://github.com/kubeshop/botkube/blob/main/docs/proposal/2022-12-23-rbac.md)

Docs will be done as a part of #935.

## Related issues & reason

https://github.com/kubeshop/botkube/issues/712

PR #960, opened 2023-02-01
Update config to support RBAC policy rules

josefkarasek:
## Update config to support RBAC policy rules

- [ ] Update Executor config
- [ ] Update Source config
- [ ] Validate config
- [ ] Add default namespace
- [ ] Add Helm options
- [ ] Add unit tests

## Related issue(s)

See also #933


pkg/config/config.go
```diff
@@ -297,8 +297,38 @@ type Plugins map[string]Plugin
...
+type PolicySubjectType string
+
+const (
+	EmptyPolicySubjectType       PolicySubjectType = ""
+	StaticPolicySubjectType      PolicySubjectType = "Static"
+	ChannelNamePolicySubjectType PolicySubjectType = "ChannelName"
+	// UserGroupNamePolicySubjectType PolicySubjectType = "UserGroupName"
+	// EmailPolicySubjectType         PolicySubjectType = "Email"
+)
+
```
josefkarasek, line 331: Example Config:
```golang
config.Config{
  Executors: map[string]config.Executors{
    "echo": config.Executors{
      Plugins: config.Plugins{
        "botkube/echo": config.Plugin{
          Enabled: true,
          Context: config.PluginContext{
            RBAC: config.PolicyRule{
              // User: config.PolicySubject{
              // 	Type:   config.EmailPolicySubjectType,
              // 	Prefix: "",
              // },
              // Group: config.PolicySubject{
              // 	Type:   config.EmptyPolicySubjectType,
              // 	Prefix: "",
              // },
              User: config.PolicySubject{
                Type:   config.StaticPolicySubjectType,
                Prefix: "",
                Static: config.StaticSubject{Value: []string{"default"}},
              },
              Group: config.PolicySubject{
                Type:   config.StaticPolicySubjectType,
                Prefix: "",
                Static: config.StaticSubject{Value: []string{"developers"}},
              },
            },
          },
        },
      },
    },
  },
}
```

helm/botkube/templates/clusterrole.yaml
```diff
@@ -21,4 +21,18 @@ rules:
   - apiGroups: [""]
     resources: ["namespaces"]
     verbs: ["get"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
```
josefkarasek, line 25: Now it's under flag `.Values.rbac.create`, which should retain old behaviour as  the default.
When ready to release, we should add a new feature flag for RBAC.

helm/botkube/templates/clusterrolebinding.yaml
```diff
@@ -13,6 +13,24 @@ roleRef:
...
+    app.kubernetes.io/instance: {{ .Release.Name }}
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: botkube-plugins-read-only
+subjects:
 - kind: ServiceAccount
   name: {{ include "botkube.serviceAccountName" . }}
   namespace: {{ .Release.Namespace }}
```
pkosiec, line 36: The subject should be a group we will use for Kubeconfig creation. E.g. as you already defined `botkube-plugins-read-only` in the `context.rbac` group mapping, then:

```suggestion
- kind: Group
  name: botkube-plugins-read-only
  apiGroup: rbac.authorization.k8s.io  
```

helm/botkube/values.yaml
```diff
@@ -427,6 +427,14 @@ executors:
...
         # -- Location for storing cached files. Must be under the Helm config directory.
         helmCacheDir: "/tmp/helm/.cache"
+      context:
+        defaultNamespace: "botkube"
+        rbac:
+          group:
+            type: Static
+            prefix: ""
+            static:
+              value: ["botkube-plugins-read-only"]
```
pkosiec, line 437: 1. Please describe the properties with `# -- ` comments which will generate docs for them.
2. Also, please consider defining static user mapping as well. I know we'll fallback to `default` user, but maybe it would be better to also document this as a part of Helm chart values 🤔

pkg/config/config.go
```diff
@@ -213,6 +213,11 @@ type Sources struct {
 	Plugins     Plugins          `koanf:",remain"`
 }
 
+// GetPlugins returns Sources.Plugins
```
pkosiec, line 216: Please add a period for each new comment 👍 
```suggestion
// GetPlugins returns Sources.Plugins.
```

pkg/config/config.go
```diff
@@ -290,15 +295,71 @@ type IngressRecommendations struct {
...
+
+// PolicyRule is the RBAC rule.
+type PolicyRule struct {
+	// User is the policy subject for user.
+	User PolicySubject
+	// Group is the policy subject for group.
+	Group PolicySubject
 }
 
+type PolicySubject struct {
```
pkosiec, line 334: Please add a comment and make sure other exported types/function/methods also have it 👍

pkg/config/config.go
```diff
@@ -290,15 +295,71 @@ type IngressRecommendations struct {
...
+
+const (
+	// EmptyPolicySubjectType is the empty policy type.
+	EmptyPolicySubjectType PolicySubjectType = ""
+	// StaticPolicySubjectType is the static policy type.
+	StaticPolicySubjectType PolicySubjectType = "Static"
+	// ChannelNamePolicySubjectType is the channel name policy type.
+	ChannelNamePolicySubjectType PolicySubjectType = "ChannelName"
+	// UserGroupNamePolicySubjectType PolicySubjectType = "UserGroupName"
+	// EmailPolicySubjectType         PolicySubjectType = "Email"
```
pkosiec, line 360: I think we can remove them for now, as we're not sure if they'll be needed 👍 
```suggestion
```

helm/botkube/templates/clusterrolebinding.yaml
pkosiec, line 36: Also, maybe this should be configurable too, just to make it visibile for users that this subject is used in `rbac.context` for plugins 🤔 WDYT?

pkg/config/validator.go
```diff
@@ -307,7 +312,48 @@ func validateSourceBindings(sl validator.StructLevel, sources map[string]Sources
 		}
 	}
 
-	validateBindPlugins(sl, enabledPluginsViaBindings)
+	validateBoundPlugins(sl, enabledPluginsViaBindings)
+	validatePluginRBAC(sl, sources, bindings)
+}
+
```
pkosiec, line 318: Please move the interface `providesPlugins` here, as this is the only usage.
Also, you can consider changing the name to `pluginProvider` or sth like this (to use nouns) 👍

pkg/config/validator.go
```diff
@@ -307,7 +312,48 @@ func validateSourceBindings(sl validator.StructLevel, sources map[string]Sources
...
+		plugins := hasPlugins[b]
+		for pluginKey, plugin := range plugins.GetPlugins() {
+			if !plugin.Enabled {
+				continue
+			}
+			if g := groups[pluginKey]; len(g) == 0 {
+				groups[pluginKey] = []string{b}
+			} else {
+				groups[pluginKey] = append(groups[pluginKey], b)
+			}
```
pkosiec, line 332: I believe we can append to nil slice - please verify if that works:
```suggestion
			groups[pluginKey] = append(groups[pluginKey], b)
```
Alternatively, this will be simpler:

```suggestion
			if groups[pluginKey] == nil {
				groups[pluginKey] = []string{}
			}
			groups[pluginKey] = append(groups[pluginKey], b)
```

But I believe the first option will work 👍

pkg/config/validator.go
```diff
@@ -307,7 +312,48 @@ func validateSourceBindings(sl validator.StructLevel, sources map[string]Sources
 		}
 	}
 
-	validateBindPlugins(sl, enabledPluginsViaBindings)
+	validateBoundPlugins(sl, enabledPluginsViaBindings)
+	validatePluginRBAC(sl, sources, bindings)
+}
+
+func validatePluginRBAC[P providesPlugins](sl validator.StructLevel, hasPlugins map[string]P, bindings []string) {
```
pkosiec, line 319: I'd recommend renaming `hasPlugins` to something else (I know you probably won't like `pluginMap` 😄  - `pluginConfigs`?) `HasXXX` pattern is usually used for bool variables, which might be confusing.

pkg/config/validator.go
```diff
@@ -307,7 +312,48 @@ func validateSourceBindings(sl validator.StructLevel, sources map[string]Sources
...
+				p2 := occurrences[i]
+				rbac2 := hasPlugins[p2].GetPlugins()[plugin].Context.RBAC
+				defaultNS2 := hasPlugins[p2].GetPlugins()[plugin].Context.Kubeconfig.DefaultNamespace
+				if !reflect.DeepEqual(rbac1, rbac2) {
+					sl.ReportError(bindings, p1, p1, invalidPluginRBACTag, p2)
+				}
+				if defaultNS1 != defaultNS2 {
+					sl.ReportError(bindings, p1, p1, invalidPluginDefaultNSTag, p2)
+				}
+			}
```
pkosiec, line 354: 1. Let's simplify the code:
- add a variable to avoid having `pluginConfigs[p2].GetPlugins()[plugin]` two times
- invert the if condition to reduce indentation
- avoid potential panics. Yes, they shouldn't happen in such case, but better to be on a safe side 👍 

Full code:
```go
        // 2. compare RBAC of duplicates
	for plugin, occurrences := range groups {
		if len(occurrences) < 2 {
			continue
		}

		// take the head of occurrences
		p1 := occurrences[0]
		p1Cfg, ok := pluginConfigs[p1].GetPlugins()[plugin]
		if !ok {
			continue
		}
		rbac1 := p1Cfg.Context.RBAC
		defaultNS1 := p1Cfg.Context.Kubeconfig.DefaultNamespace

		// compare the head with the tail
		for i := 1; i < len(occurrences); i++ {
			p2 := occurrences[i]
			p2Cfg, ok := pluginConfigs[p1].GetPlugins()[plugin]
			if !ok {
				continue
			}
			
			rbac2 := p2Cfg.Context.RBAC
			defaultNS2 := p2Cfg.Context.Kubeconfig.DefaultNamespace
			if !reflect.DeepEqual(rbac1, rbac2) {
				sl.ReportError(bindings, p1, p1, invalidPluginRBACTag, p2)
			}
			if defaultNS1 != defaultNS2 {
				sl.ReportError(bindings, p1, p1, invalidPluginDefaultNSTag, p2)
			}
		}

	}
```

pkg/config/validator.go
```diff
@@ -117,6 +120,8 @@ func registerBindingsValidator(validate *validator.Validate, trans ut.Translator
 		conflictingPluginRepoTag:    "{0}{1}",
 		conflictingPluginVersionTag: "{0}{1}",
 		invalidPluginDefinitionTag:  "{0}{1}",
+		invalidPluginRBACTag:        "RBAC of {0} and {1} are different. But have to be identical.",
+		invalidPluginDefaultNSTag:   "Default namespace of {0} and {1} are different. But have to be identical.",
```
pkosiec, line 124: I think it would be great to provide info about the specific plugin key - we already have such data so it will be easy.
Then the messages could be something like:
```suggestion
		invalidPluginRBACTag:        "RBAC configuration for plugin '{0}' used in  '{0}' and '{1}' bindings must be identical when used together",
		invalidPluginDefaultNSTag:   "Default namespace for plugin '{0}' used in '{0}' and '{1}' bindings must be identical when used together.",
```

BTW I think later after impl we can remove the verification check for default namespaces. Merging `context.kubeconfig` will be relatively eas - we can take the last value 🤔  But let's keep it for now.

pkosiec (commented):
Overall looks good, just a few small comments.

helm/botkube/values.yaml
josefkarasek: As we recently discussed (in a different PR) naming - noun plural vs singular - I'm thinking the same about `value`.
I'm more inclined towards the plural form or something like `references`. What do you think?

helm/botkube/templates/clusterrolebinding.yaml
josefkarasek: I see more issues than benefits here. The idea of this ClusterRoleBinding is to retain current functionality.
I think that once we progress with implementing RBAC, there will be more specific configuration options for users to use.

helm/botkube/values.yaml
pkosiec: `references` sounds to me like it should be defined somewhere in the config 🤔 maybe `values` then? (only for group mapping)

helm/botkube/templates/clusterrolebinding.yaml
pkosiec: Sorry, I meant just the group name, not the whole subject. The only problem is that users will need to use the name in the config so they do need to know it 🤔 How they can learn it intuitively, if it is not found in the `values.yaml` doesn't support it? 🤔 I think that wouldn't cause too much issues, as by default we will keep it as `botkube-plugins-read-only` and in that way user will be able to find the occurrence when configuring Helm chart.

pkg/config/testdata/TestLoadConfigSuccess/config.golden.yaml
```diff
@@ -331,6 +331,20 @@ sources:
                 enabled: true
                 config:
                     field: value
+                context:
+                    kubeconfig:
+                        defaultnamespace: ""
```
pkosiec, line 336: 🤔 It should be camel case. Can you add the yaml struct tags for the config to make sure all properties are marshalled with proper names? Thanks!

pkosiec (approved):
LGTM 👍 Please see my one comment before merge. Thanks!

PR #963, opened 2023-02-01
Support include/exclude regexes for event and resource constraints

### LINKED ISSUE #954: Muting certain events
## Overview

I am looking for a way to mute certain events at the source level. I found that I can write a regex for `event.message` to filter the events, but the Go regex engine does not allow for negative lookahead/lookbehind, so writing and maintaining the right regex to allow everything except what I want to filter becomes a mess.

It would be nice if I could write a negative filter, to allow anything except for what is matched by the criteria.

## Acceptance Criteria

- Allow negative filters on `event.message` and `event.kind`.
- When the criteria matches, the event is ignored, when the criteria does not match the event is processed.

## Reason

We need a way to ignore certain cluster events, but allow everything else.

pkosiec:
## Description

- Support `include/exclude` regexes for `event.reason`, `event.message` and `resources[].name`.

## Breaking changes

_(To copy and paste in https://github.com/kubeshop/botkube/issues/972)_

- Kubernetes source constraints `event.reason`, `event.message` and `resources[].name` are now objects with two lists: `include` and `exclude`. Both take exact values or regex patterns. Exclude takes precedence over include.

   This allows you to set `exclude` patterns, and also unifies the configuration with `namespaces` constraints.

## Notes - open question
I didn't rename `namespaces` to `namespace` or `event.types` to `event.type` as I weren't sure if that's a good change. However, if you have any ideas, then please let me know - we're already doing a breaking change here, so I can unify the config properties more 🙂 

Maybe we should rename `namespaces` to `namespace` for consistency (as it is an object), and keep plural form for `event.types` as it is an array? Let me know 🙂 

## TODO

- [x] Update docs: https://github.com/kubeshop/botkube-docs/pull/211

## Testing

Check out the PR (`gh pr checkout 963`).

Use the configuration:
```
communications:
  "default-group":
    socketSlack:
      enabled: true
      appToken: "xapp-"
      botToken: "xoxb-"
      notification:
        type: "short"
      channels:
        "default":
          name: botkube-demo
          bindings:
            executors:
              - kubectl-read-only
              - helm
              - plugin-based
            sources:
              - k8s-all-events
              - k8s-recommendation-events
              - plugin-based
        "secondary":
          name: priv-channel
          bindings:
            executors: []
            sources:
              - k8s-all-events-except
sources:
  'k8s-all-events-except':
    kubernetes:
      namespaces:
        include:
          - ".*"
      event:
        types:
          - create
          - delete
          - error
        message:
          include: [".*"]
          exclude: ["^Readiness probe failed.*$"]
      resources:
        - type: v1/pods
        - type: v1/services
        - type: networking.k8s.io/v1/ingresses
        - type: v1/nodes
        - type: v1/namespaces
        - type: v1/persistentvolumes
        - type: v1/persistentvolumeclaims
        - type: v1/configmaps
        - type: rbac.authorization.k8s.io/v1/roles
        - type: rbac.authorization.k8s.io/v1/rolebindings
        - type: rbac.authorization.k8s.io/v1/clusterrolebindings
        - type: rbac.authorization.k8s.io/v1/clusterroles
        - type: apps/v1/daemonsets
          event:
            types:
              - create
              - update
              - delete
              - error
          updateSetting:
            includeDiff: true
            fields:
              - status.numberReady
        - type: batch/v1/jobs
          event:
            types:
              - create
              - update
              - delete
              - error
          updateSetting:
            includeDiff: true
            fields:
              - status.conditions[*].type
        - type: apps/v1/deployments
          event:
            types:
              - create
              - update
              - delete
              - error
          updateSetting:
            includeDiff: true
            fields:
              - status.availableReplicas
        - type: apps/v1/statefulsets
          event:
            types:
              - create
              - update
              - delete
              - error
          updateSetting:
            includeDiff: true
            fields:
              - status.readyReplicas

settings:
  clusterName: "dev"
  upgradeNotifier: false
```

Run Botkube:
```
export BOTKUBE_SETTINGS_KUBECONFIG=$KUBECONFIG # optional
export BOTKUBE_CONFIG_PATHS="$(pwd)/helm/botkube/values.yaml,$(pwd)/comm_config.yaml"
go run cmd/botkube/main.go
```

Run the command to trigger events:
```
k delete po -n kube-system -l k8s-app=metrics-server
```

See events with `Readiness probe failed` message:
image

On `priv-channel`, they will be filtered out nicely:

image

## Related issue(s)

#954

### DIFF
--- a/internal/source/registration.go
+++ b/internal/source/registration.go
@@ -4,9 +4,9 @@ import (
 	"context"
 	"fmt"
 	"reflect"
-	"regexp"
 	"strings"
 
+	"github.com/dlclark/regexp2"
 	"github.com/sirupsen/logrus"
 	coreV1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
@@ -208,18 +208,30 @@ func matchRegexForStringsIfDefined(regexStr string, str []string) (bool, error) 
 		return true, nil
 	}
 
-	regex, err := regexp.Compile(regexStr)
+	regex, err := regexp2.Compile(regexStr, regexp2.None)
 	if err != nil {
 		return false, fmt.Errorf("while compiling regex: %w", err)
 	}
 
+	if len(str) == 0 {
+		// no messages, so let's check if regex matches empty string
+		str = append(str, "")
+	}
+
+	errs := multierror.New()
 	for _, s := range str {
-		if regex.MatchString(s) {
-			return true, nil
+		match, err := regex.MatchString(s)
+		if err != nil {
+			errs = multierror.Append(errs, fmt.Errorf("while matching regex %s: %w", regexStr, err))
+			continue
+		}
+
+		if match {
+			return true, errs.ErrorOrNil()
 		}
 	}
 
-	return false, nil
+	return false, errs.ErrorOrNil()
 }
 
 func kvsSatisfiedForMap(expectedKV, obj map[string]string) bool {
--- a/internal/source/registration.go
+++ b/internal/source/registration.go
@@ -177,7 +177,6 @@ func (r registration) shouldSendEventToRoute(route route, event event.Event) (bo
 	// event message
 	if route.event.Message.AreConstraintsDefined() {
 		var anyMsgMatches bool
-		var lastErr error
 
 		eventMsgs := event.Messages
 		if len(eventMsgs) == 0 {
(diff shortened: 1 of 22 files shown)


pkosiec:
Converting to draft after internal conversation about our preferred approach. The agreement is to rework the PR to:
- remove the advanced regex engine,
- introduce `include`/`exclude` lists in all of the source filters and use Go regexes there.

Converting this PR to draft then 🙂

internal/source/registration.go
```diff
@@ -145,81 +144,104 @@ func (r registration) sourcesForEvent(routes []route, event event.Event) ([]stri
...
 		}
 
-		// labels
-		if !kvsSatisfiedForMap(route.labels, event.ObjectMeta.Labels) {
-			continue
+		for _, msg := range eventMsgs {
+			match, err := route.event.Message.IsAllowed(msg)
+			if err != nil {
+				lastErr = err
+				continue
```
josefkarasek, line 192: What's the purpose of `lastErr`? Why wait for the last error and not break on first error?

josefkarasek (approved):
lgtm :+1:

internal/source/registration.go
pkosiec: 🤔 Right... It will be always the same error about compiling the same regex. Good catch, will fix that!

PR #968, opened 2023-02-03
Use new format to download deps

mszostok:
## Description

@huseyinbabal found out an issue that `gh` plugin index cannot be built on darwin/arm64, the problem is that we call the `DownloadDependencies` on plugin start that tries to download deps. We start the plugin because we want to only fetch the metadata for the index generation. So here are 2 issues:
- calling the `DownloadDependencies` - however it's not needed anymore so I will delete that and use MetadataOutput.Dependecies
- the `gh` CLI doesn't have `darwin/arm64` binary, we can build our own if that will be needed. However, if the plugin will be run on a k8s then this issue won't happen as we have a proper `linux/arm64` arch.


cmd/executor/helm/README.md
```diff
@@ -97,7 +97,6 @@ Only Slack integration is described, but described steps are the same for all co
 
    ```bash
    @Botkube helm help
-   @Botkube helm --help
```
josefkarasek, line 100: Good catch, I missed this one.

cmd/executor/gh/main.go
```diff
@@ -101,7 +101,7 @@ func (e *GHExecutor) Execute(ctx context.Context, in executor.ExecuteInput) (exe
 }
 
 // Help returns help message
-func (*GHExecutor) Help(ctx context.Context) (interactive.Message, error) {
+func (*GHExecutor) Help(_ context.Context) (interactive.Message, error) {
```
josefkarasek, line 104: Or no name, as in `Metadata(context.Context)` above.

josefkarasek (approved):
:+1:

Issue #841, opened 2022-11-08
Extract Kubectl executor feature as a plugin


huseyinbabal:
## Overview

Botkube supports `kubectl` execution and it is a typical Executor plugin. We need to extract it as a standalone plugin.

## Acceptance Criteria

- Extract `kubectl` executor as a plugin and maintain it Botkube main repository
- Ensure it is backward compatible.
- Plugin parameters will be documented in the plugin's README.md page, and they can be provided from Botkube as an environment variable like `KUBERNETES_...` so that they can be passed to the executable [in this line](https://github.com/huseyinbabal/botkube-plugins-playground/blob/c85cb7b84296a2f41c2dcdd8cac77c0e9dd9c69a/plugin/manager.go#L159)
- Remember about interactivity
- Update gRPC executor API
- Make sure anonymized telemetry is still supported - we can lose some data, but confirm with @brampling if we can do that
- Update the document

## Additional scope / notes
- Ensure it supports the RBAC design according to #712 (design https://github.com/kubeshop/botkube/blob/main/docs/proposal/2022-12-23-rbac.md), if #934 is already done. If not, ensure we can easily support this in the plugin.

pkosiec:
See @mszostok's PR: https://github.com/kubeshop/botkube/pull/857

mszostok:
Implementation:
- https://github.com/kubeshop/botkube/pull/973
- https://github.com/kubeshop/botkube/pull/982
- https://github.com/kubeshop/botkube/pull/986

Docs:
- https://github.com/kubeshop/botkube-docs/pull/213
- https://github.com/kubeshop/botkube-docs/pull/215

PR #979, opened 2023-02-13
Introduce global bot name placeholder

mszostok:
## Description

- Introduce global bot name placeholder. Now, instead of having direct dependency to a given bot name, you can use `{{BotName}}` (`api.MessageBotNamePlaceholder`) placeholder in your messages, and it will be properly handled by a given communication platform. It works not only in commands, but also in all other message properties.
- Unify sending messages in communication platforms.

Reason:

Currently, each bot (Slack, Discord, etc.) exposed the `BotName()` method and were passing this across all our code. It was only to construct a proper interactive message. Such approach doesn't scale well. Paweł already introduced the concept of universal bot name placeholder, which is great. Now it's up to the bot implementation to render the message with a proper bot name. Such decoupling is needed for extracting `kubectl` as plugin. In this PR, I use the universal bot placeholder across all our code base.

## Related issue(s)

- #841


pkg/bot/discord.go
```diff
@@ -153,21 +153,16 @@ func (b *Discord) SendEvent(_ context.Context, event event.Event, eventSources [
...
-// SendGenericMessage sends interactive message to selected Discord channels.
+// SendMessage sends interactive message to selected Discord channels.
 // Context is not supported by client: See https://github.com/bwmarrin/discordgo/issues/752.
-func (b *Discord) SendGenericMessage(_ context.Context, genericMsg interactive.GenericMessage, sourceBindings []string) error {
-	msg := genericMsg.ForBot(b.BotName())
-
+func (b *Discord) SendMessage(_ context.Context, msg interactive.CoreMessage, sourceBindings []string) error {
 	errs := multierror.New()
 	for _, channelID := range b.getChannelsToNotify(sourceBindings) {
-		b.log.Debugf("Sending message to channel %q: %+v", channelID, msg)
```
josefkarasek, line 163: Can we retain some of the debug logging?

josefkarasek (approved):
lgtm :+1:

pkg/bot/discord.go
mszostok: it is still logged in the `send` method, so I just removed it to don't have duplicates :)

Issue #840, opened 2022-11-08
Extract Kubernetes source feature as a plugin


huseyinbabal:
## Importan Update
Kubernetes plugin is successfully extracted as a plugin, and it works now as expected, but since entire Botkube system depends on Kubernetes event, we needed to touch all kind of component including filters, recommendations, reporting, etc... We introduced follow-up items as follows;

- [x] #993 - @mszostok
- [x] #994 - @mszostok
- [x] #995 - @mszostok
- [x] #996 - @mszostok
- [x] #997 - @mszostok
- [x] #998 
- [x] #999 
- [x] #1000 - @mszostok 
- [x] Remove `botkube.io/channel: {channel}` annotation from k8s source

Out of scope:
- [ ] #1009
- [ ] #1001 

## Overview

Botkube supports Kubernetes events. It is a typical example of a Source plugin. We need to extract it as a standalone plugin.

## Acceptance Criteria

- Once we extract this as a plugin, it will be a sub-process in the pod which means it will use the same service account as other plugins will. For now, we can assume that customer will prepare a unified role and attach it to the service account. This is something like using your workstation machine for kubectl, helm, and does not care which service account you use in the beginning.
- Currently, Kubernetes source events are tightly coupled with Botkube, we can add a sourcing mechanism as a Kubernetes plugin. You can see an independent example [here](https://github.com/huseyinbabal/botkube-plugins/tree/main/plugins/kubernetes)
- It should be background compatible.
- Plugin-specific parameters should be passed during plugin initialization. Hashicorp's Go plugin system runs the plugin executable as a subprocess, so it also accepts exec arguments where we can resolve configuration and pass them as separate arguments
- Plugin parameters will be documented in the plugin's README.md page, and they can be provided from Botkube as an environment variable like `KUBERNETES_...` so that they can be passed to the executable in this [line](https://github.com/huseyinbabal/botkube-plugins-playground/blob/c85cb7b84296a2f41c2dcdd8cac77c0e9dd9c69a/plugin/manager.go#L159)
- Since each source plugin has different data structure, we cannot have a general data structure except we specified #Cloud Events. data field has free data structure, but we can convert them to map then extract any section based on customer definition by using Go template. data field has free flat structure, a simple key-value pairs with one level, but we can put mandatory fields in this struct like message, user, etc. Or we can put mandatory fields to cloud events schema as shown below.
```
{
"id" : "e2361318-e50b-472e-8b17-13dbac1daac1",   
[...]

Issue #972, opened 2023-02-06
Release Botkube v1.0


pkosiec:
## Overview

This task aggregates all the to-dos which should be done when it comes to the release.

## AC
- [ ] Go through the release process instruction and release new Botkube (https://docs.botkube.io/community/contribute/release)
- [ ] Describe any breaking changes in the GitHub release (see the issues with `breaking` label)
- [ ] Test the release candidate. Include:
    - Manual tests for Slack interactivity
    - Test telemetry
    - Aggregate and coordinate the manual testing described previously (e.g. https://github.com/kubeshop/botkube/issues/904, https://github.com/kubeshop/botkube/issues/957)
- [ ] Schedule all community activities - blog post, sharing the news in social media, livestream, etc. (@brampling)

mszostok:
Breaking changes:
- `notification: short/long` was removed, it's up to the plugin to return a suitable message.

huseyinbabal:
Breaking Changes:
- `kubernetes` configuration is named as `botkube/kubernetes`
- `filters` configuration is removed from main configuration, now it is under Kubernetes source plugin.
### Before:
```yaml
# ...trimmed
filters:
  kubernetes:
    objectAnnotationChecker: true
    nodeEventsChecker: true
#...trimmed
```
### After:
```yaml
...trimmed
 'k8s-all-events':
    displayName: "Kubernetes Info"
    botkube/kubernetes:
      enabled: true
      config:
        filters:
          kubernetes:
            objectAnnotationChecker: true
            nodeEventsChecker: true
# ...trimmed
```

mszostok:
Make sure that the plugin repository is up to date: https://github.com/kubeshop/botkube-plugins-template/tree/main

pkosiec:
## Release activities
- [ ] Ensure that all security vulnerabilities are fixed or mitigated
  
  * assignee: 
- [x] Cut initial release candidate
  
  * assignee: @huseyinbabal 
- [ ] Finalize release
  
  * assignee: @huseyinbabal 
- [x] Test Slack  (RTM)
  
  * assignee: CI
  * it's tested as a part of CI.
- [ ]  Test Slack  (Socket)
  
  * test all interactive buttons/modals
  * assignee: 
- [ ]  Test Mattermost
  
  * assignee: 
- [ ]  Test Discord
  
  * assignee: 
- [ ]  Test MS Teams
  
  * assignee: 

- [ ] Test sinks:
   - Test Elasticsearch 
[...]

PR #985, opened 2023-02-17
Kubernetes source plugin

huseyinbabal:
⚠️ Kubernetes plugin extraction was a huge task, but we couldn't figured it out in the beginning. So, you see a failing tests (2 of them only), but I intentionally left them as is since I have a couple of followup tasks for this PR. You can see them all [here](https://github.com/kubeshop/botkube/issues/840)  

## Description

- Kubernetes plugin is implemented
- Filters are removed from core configuration and  moved to kubernetes plugin
- Actions are managed in core, not in kubernetes source anymore
- Recommendations are part of kubernetes plugin, they are removed from core
- Please see followup tasks [here](https://github.com/kubeshop/botkube/issues/840) 

### Build Kubernetes Plugin
Following command will build kubernetes plugin 
```shell
PLUGIN_DOWNLOAD_URL_BASE_PATH=http://localhost:8080/plugin-dist make gen-plugins-index
```

### Run Plugin Server
```shell
npx serve --listen 8080
```

### Run Botkube With the Following Configurations
```yaml

communications:
  'default-group':
    # Settings for Slack
     socketSlack:
       enabled: true
       appToken: "xapp-..."
       botToken: "xoxb-..."
      notification:
        type: "long" # TODO: Change to `short` during testing
      channels:
        'default':
          name: botkube-demo
          bindings:
            sources:
              # - plugin-based
              # - k8s-events
              - k8s-annotated-cm-delete
              - k8s-updates
              # - k8s-pod-create-events
                #              - plugin-based
            executors:
              - kubectl-plugin
        'secondary':
          name: botkube-demo-secondary
          bindings:
            sources:
              # - plugin-based
              - k8s-events
              # - k8s-annotated-cm-delete
              # - k8s-pod-create-events
                #              - plugin-based
            executors:
              - kubectl-plugin

settings:
  clusterName: gke-playground
  configWatcher: true
  lifecycleServer:
    deployment:
      name: botkube
      namespace: botkube
    port: "2113"

configWatcher:
  enabled: false
  initialSyncTimeout: 0

sources:
  'k8s-events':
    displayName: "K8s recommendations"
    'botkube/kubernetes':
      enabled: true
      config:
        kubeConfig: /Users/huseyin/.kube/config
        log:
          level: debug
        recommendations:
          pod:
            noLatestImageTag: true
            labelsSet: true
          ingress:
            backendServiceValid: false
            tlsSecretValid: false
        namespaces:
          include:
            - botkube
        event:
          types:
            - create
            - update
        resources:
          - type: v1/configmaps
  'k8s-annotated-cm-delete':
    displayName: "K8s ConfigMap delete events"
    'botkube/kubernetes':
      enabled: true
      config:
        kubeConfig: /Users/huseyin/.kube/config
        log:
          level: debug
        namespaces:
          include:
            - botkube
        # labels:
        #   test.botkube.io: "true"
        event:
          types:
            - delete
        resources:
          - type: v1/configmaps

  'k8s-pod-create-events':
    'botkube/kubernetes':
      enabled: true
      config:
        log:
          level: debug
        kubeConfig: /Users/huseyin/.kube/config
        namespaces:
          include:
            - botkube
        event:
          types:
            - create
        resources:
          - type: v1/pods

  'k8s-updates':
    displayName: "K8s ConfigMaps updates"
    'botkube/kubernetes':
      enabled: true
      config:
        log:
          level: debug
        kubeConfig: /Users/huseyin/.kube/config
        namespaces:
          include:
            - default
        event:
          types:
            - create
            - update
            - delete
        resources:
          - type: v1/configmaps
            namespaces:
              include:
                - botkube
            event: # overrides top level `event` entry
              types:
                - update
            updateSetting:
              includeDiff: true
              fields:
                - data
executors:
  'kubectl-plugin':
    botkube/kubectl:
      enabled: true
      # namespaces:
      #   # -- List of allowed Kubernetes Namespaces for command execution.
      #   # It can also contain a regex expressions:
      #   #  `- ".*"` - to specify all Namespaces.
      #   include:
      #     - ".*"
      #   # -- List of ignored Kubernetes Namespace.
      #   # It can also contain a regex expressions:
      #   #  `- "test-.*"` - to specify all Namespaces.
      #   exclude: []
      # # -- If true, enables `kubectl` commands execution.
      # enabled: true
      # ## List of allowed `kubectl` commands.
      # commands:
      #   # -- Configures which `kubectl` methods are allowed.
      #   verbs: ["api-resources", "api-versions", "cluster-info", "describe", "explain", "get", "logs", "top"]
      #   # -- Configures which K8s resource are allowed.
      #   resources: ["deployments", "pods", "namespaces", "daemonsets", "statefulsets", "storageclasses", "nodes", "configmaps", "services", "ingresses"]
      # # -- Configures the default Namespace for executing Botkube `kubectl` commands. If not set, uses the 'default'.
      # defaultNamespace: default
      # # -- If true, enables commands execution from configured channel only.
      # restrictAccess: false

actions:
  'describe-created-resource':
    # -- If true, enables the action.
    enabled: false
    # -- Action display name posted in the channels bound to the same source bindings.
    displayName: "Describe created resource"
    # -- Command to execute when the action is triggered. You can use Go template (https://pkg.go.dev/text/template) together with all helper functions defined by Slim-Sprig library (https://go-task.github.io/slim-sprig).
    # You can use the `{{ .Event }}` variable, which contains the event object that triggered the action. See all available event properties on https://github.com/kubeshop/botkube/blob/main/pkg/event/event.go.
    # @default -- See the `values.yaml` file for the command in the Go template form.
    command: "kubectl describe {{ .Event.Kind | lower }}{{ if .Event.Namespace }} -n {{ .Event.Namespace }}{{ end }} {{ .Event.Name }}"

    # -- Bindings for a given action.
    bindings:
      # -- Event sources that trigger a given action.
      sources:
        #- plugin-based
      # -- Executors configuration used to execute a configured command.
      executors:
        - kubectl-plugin
plugins:
  repositories:
    botkube:

### To See Notifications in Action

```shell
kubectl run s9 --image=nginx:latest -n botkube
```

```shell
kubectl create configmap my-config-def --from-literal=key1=config1 -n botkube
```

and then update configmap to see diff is shown in slack

## Related issue(s)
Relates #840


pkg/api/executor/grpc_adapter.go
```diff
@@ -53,7 +53,7 @@ type (
 		//
 		//	Old approach:
 		//	  return executor.ExecuteOutput{
-		//	  	Data: data,
+		//	  	Event: data,
```
mszostok, line 56: please revert
```suggestion
		//	  	Data: data,
```

pkg/api/source/grpc_adapter.go
```diff
@@ -172,6 +218,22 @@ func (p *grpcServer) Stream(req *StreamRequest, gstream Source_StreamServer) err
 			if err != nil {
 				return err
 			}
+		case msg, ok := <-stream.Message:
+			if !ok {
+				return nil // output closed, no more chunk logs
```
mszostok, line 223:
```suggestion
				return nil // output closed, no more messages
```

pkg/event/event.go
```diff
@@ -71,7 +71,10 @@ var LevelMap = map[config.EventType]config.Level{
 func New(objectMeta metaV1.ObjectMeta, object interface{}, eventType config.EventType, resource string) (Event, error) {
 	typeMeta := k8sutil.GetObjectTypeMetaData(object)
 	event := Event{
-		TypeMeta:   typeMeta,
```
mszostok, line 74: I know that we will fix that in the follow-up task. But let's remove that already so we will have a better diff with `internal/source/kubernetes/event/event.go` - in the rest of the code you can import that moved event instead.

pkg/k8sutil/resource.go
```diff
@@ -14,6 +14,11 @@ import (
 	"k8s.io/client-go/dynamic"
```
mszostok, line 14: please remove that file and use `internal/source/kubernetes/k8sutil/resource.go` instead.

mszostok (commented):
I still need to check 5 more files:
- internal/source/dispatcher.go
- internal/source/kubernetes/registration.go
- internal/source/kubernetes/router.go
- internal/source/kubernetes/source.go
- internal/source/scheduler.go

And run some manual tests.

mszostok:
You need to move `filters` under `k8s` source:
https://github.com/kubeshop/botkube/blob/09e4a265cc182ee259638bf2daf13e1aac016c7a/helm/botkube/values.yaml#L411-L420

mszostok:
As a part of this PR, please also document all breaking under: https://github.com/kubeshop/botkube/issues/972. Thanks! 🙇

internal/source/kubernetes/source.go
```diff
@@ -0,0 +1,667 @@
...
+    "Event": {
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
+        "types": {
+          "description": "Lists all event types to be watched.",
+          "type": "array",
+          "items": {
+            "type": "string",
+            "enum": [
```
mszostok, line 474: probably copy-past error :)

internal/source/kubernetes/config/config.go
```diff
@@ -0,0 +1,307 @@
...
+}
+
+// Filters contains configuration for built-in filters.
+type Filters struct {
+	Kubernetes KubernetesFilters `yaml:"kubernetes"`
+}
+
+// KubernetesFilters contains configuration for Kubernetes-related filters.
+type KubernetesFilters struct {
+	// ObjectAnnotationChecker enables support for `botkube.io/disable` and `botkube.io/channel` resource annotations.
```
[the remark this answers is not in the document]
pkosiec: Yes, I think we can remove `botkube.io/disable`, but I know at least a few people use it, so we should introduce `exclude` annotations and labels instead as a follow-up 🤔

cmd/executor/gh/main.go
```diff
@@ -105,15 +105,15 @@ func (*GHExecutor) Help(context.Context) (api.Message, error) {
 }
 
 var depsDownloadLinks = map[string]api.Dependency{
-	// Links source: https://github.com/cli/cli/releases/tag/v2.21.2
+	// Links source: https://github.com/cli/cli/releases/tag/v2.22.1
```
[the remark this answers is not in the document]
pkosiec: Actually it is upgrading (but there's 1.23.0 already)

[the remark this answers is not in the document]
huseyinbabal: I will revert and test it again, I had a problem with gh plugin generation at that time, I will test with up to date version

pkosiec (commented):
Great work! Looks like the basic functionality works OK, I tested a few configurations and they worked well. Telemetry also OK 👍 

I agree with all @mszostok comments, and here are just a few ones from my side 🙂

internal/source/kubernetes/k8sutil/resource.go
```diff
@@ -0,0 +1,103 @@
+package k8sutil
```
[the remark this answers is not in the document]
huseyinbabal: Let me explain;

As you already know, we haven't implemented all the platforms yet, platform renderers depends on `event` package, and `event` depends on k8sutil, that's why I keep them, we can remove that after platform migration, wdyt?

pkg/event/event.go
huseyinbabal: Isn't that dangerous to depend a public package to a plugin internals? I just wanted to wait until platform is removed, this event package is still used in a couple of places.

pkg/k8sutil/resource.go
huseyinbabal: As described above, I didn't wanted to depend our core to a plugin internal. I will be removing k8sutils anyways after platform migration, wdyt?

internal/source/kubernetes/source.go
```diff
@@ -0,0 +1,667 @@
...
+		strBuilder.WriteString(fmt.Sprintf("%s%s\n", msgPrefix, m))
+	}
+
+	return strBuilder.String()
+}
+
+func jsonSchema() api.JSONSchema {
+	return api.JSONSchema{
+		Value: heredoc.Docf(`
+{
```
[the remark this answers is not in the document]
huseyinbabal: I tested other plugins' json specs and it gives schema not found on validation, can we elaborate on that part?

internal/source/kubernetes/source.go
```diff
@@ -0,0 +1,667 @@
...
+        "filters": {
+          "description": "Filter settings for various sources.",
+          "$ref": "#/definitions/Filter"
+        },
+        "actionVerbs": {
+          "description": "Allowed verbs for actionable events.",
+          "type": "array",
+          "items": {
+            "type": "string",
+            "enum": [
```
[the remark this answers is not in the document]
huseyinbabal: showing possible values will prevent flakiness, and I thought that, time to time we can change that options wdyt? I put current values we have in values.yaml, but I can definitely extend that to all possible values

internal/source/kubernetes/config/config.go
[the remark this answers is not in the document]
huseyinbabal: I completely moved the logic, and didn't wanted to touch internals. Handling as followup is fine by me.

internal/source/kubernetes/config/config.go
```diff
@@ -0,0 +1,307 @@
...
+	"gopkg.in/yaml.v3"
+
+	"github.com/kubeshop/botkube/pkg/api/source"
+	"github.com/kubeshop/botkube/pkg/ptr"
+)
+
+// Config Kubernetes configuration
+type Config struct {
+	KubeConfig           string `yaml:"kubeConfig"`
+	ClusterName          string
```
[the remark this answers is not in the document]
huseyinbabal: ok, I will remove them from config and pass the existing values from our configuration.

internal/source/kubernetes/event/event.go
```diff
@@ -0,0 +1,145 @@
...
+	Title           string
+	Name            string
+	Namespace       string
+	Messages        []string
+	Type            config.EventType
+	Reason          string
+	Error           string
+	Level           config.Level
+	Cluster         string
+	Channel         string
```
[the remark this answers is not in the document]
huseyinbabal: There is one usage in node annotation checker, is it ok to keep it for now?

cmd/source/kubernetes/main.go
```diff
@@ -0,0 +1,19 @@
+package main
```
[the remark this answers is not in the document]
huseyinbabal: I just saw this, never happened to me but this is something we need to spot if we see that second time.

cmd/source/kubernetes/main.go
```diff
@@ -0,0 +1,19 @@
...
+
```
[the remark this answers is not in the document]
huseyinbabal: That is because slack renderer has very special functionality to render kubernetes metadata like that, so rendering used to be handled on slack_renderer. However, in plugin structure, we render messages in advance, and I tried to unified them as a general message format. Again, I am open to suggestions here :)

pkg/event/event.go
mszostok: it's not, actually it's a proper solution to import some parts from `internal` in `pkg`. If you want I can explain on 1:1 call

internal/source/kubernetes/config/config.go
```diff
@@ -0,0 +1,307 @@
...
+	InformerReSyncPeriod *time.Duration     `yaml:"informerReSyncPeriod"`
+	Log                  *Log               `yaml:"log"`
+	Recommendations      *Recommendations   `yaml:"recommendations"`
+	Event                *KubernetesEvent   `yaml:"event"`
+	Resources            []Resource         `yaml:"resources" validate:"dive"`
+	ActionVerbs          []string           `yaml:"actionVerbs" validate:"dive"`
```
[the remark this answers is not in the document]
huseyinbabal: @pkosiec we don't enable plugins if they are not bound to channels. @mszostok if you are ok too, I will change that logic in a way that plugins will be always enabled.

cmd/source/kubernetes/main.go
```diff
@@ -0,0 +1,19 @@
...
+import (
```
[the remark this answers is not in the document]
huseyinbabal: @pkosiec we don't enable plugins if they are not bound to channels. @mszostok if you are ok too, I will change that logic in a way that plugins will be always enabled.

internal/source/kubernetes/config/config.go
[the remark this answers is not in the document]
huseyinbabal: Basically, we have command in the context of actions only, that is why I have added like that.

internal/source/kubernetes/commander/guard.go
```diff
@@ -0,0 +1,258 @@
+package commander
```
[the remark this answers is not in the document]
huseyinbabal: followup task : extract to pkg to share between kubectl and k8s

cmd/source/kubernetes/main.go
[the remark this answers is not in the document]
huseyinbabal: followup

cmd/source/kubernetes/main.go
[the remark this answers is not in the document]
huseyinbabal: followup

internal/source/kubernetes/config/config.go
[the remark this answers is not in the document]
mszostok: OK, to do that in the follow-up 👍

internal/source/scheduler.go
```diff
@@ -97,27 +112,31 @@ func (d *Scheduler) schedule(ctx context.Context, bindSources []string) error {
 	// Holds the array of configs for a given plugin.
 	// For example, ['botkube/kubernetes@v1.0.0']->[]{"cfg1", "cfg2"}
```
mszostok, line 113: Please update the comment to reflect current approach. Thanks!

pkg/api/source/grpc_adapter.go
```diff
@@ -23,13 +25,50 @@ type (
 	StreamInput struct {
 		// Configs is a list of Source configurations specified by users.
 		Configs []*Config
```
mszostok, line 27: Right now there is no option that we will get a list of configs as we start multi-processes. When we will change do documentation and simplify this logic here?

Should we also do that for `executors`? 
Maybe we should do that as a follow-up after RBAC? WDYT @pkosiec @huseyinbabal?

internal/source/kubernetes/source.go
```diff
@@ -0,0 +1,687 @@
...
+	s.config = cfg
+	s.logger = loggerx.New(loggerx.Config{
+		Level: cfg.Log.Level,
+	})
+	s.ctx = ctx
+	s.clusterName = input.Context.ClusterName
+	s.kubeConfig = input.Context.KubeConfig
+	go consumeEvents(s)
+	return out, nil
+}
```
mszostok, line 91: Let's clean-up the code in this function.
```go
func (*Source) Stream(ctx context.Context, input source.StreamInput) (source.StreamOutput, error) {
	cfg, err := config.MergeConfigs(input.Configs)
	if err != nil {
		return source.StreamOutput{}, fmt.Errorf("while merging input configs: %w", err)
	}

	s := Source{
		startTime: time.Now(),
		messageCh: make(chan source.Message),
		config:    cfg,
		logger: loggerx.New(loggerx.Config{
			Level: cfg.Log.Level,
		}),
		ctx:         ctx,
		clusterName: input.Context.ClusterName,
		kubeConfig:  input.Context.KubeConfig,
	}

	go consumeEvents(s)

	return source.StreamOutput{
		Message: s.messageCh,
	}, nil
}
```

mszostok (commented):
Overall good work 🚀

Once all comments will be resolved I will play with it locally as Paweł did.
Please also track all work that you would like to do in the follow-ups. Thanks!

helm/botkube/values.yaml
```diff
@@ -135,260 +135,278 @@ sources:
     displayName: "Kubernetes Info"
```
mszostok, line 135: please update line 118 from `kubernetes` to `botkube/kubernetes`

internal/source/kubernetes/source.go
```diff
@@ -0,0 +1,687 @@
...
+
+	if len(e.Kind) <= 0 {
+		s.logger.Warn("sendEvent received e with Kind nil. Hence skipping.")
+		return
+	}
+
+	recRunner, recCfg := s.recommFactory.New(s.config)
+	err := recRunner.Do(s.ctx, &e)
+	if err != nil {
+		s.logger.Errorf("while running recommendations: %w", err)
```
mszostok, line 218:
```suggestion
		s.logger.Errorf("while running recommendations: %w", err)
		return
```

pkg/api/source/grpc_adapter.go
```diff
@@ -23,13 +25,50 @@ type (
...
+		//   - api.NewCodeBlockMessage("body", true)
+		//   - api.NewPlaintextMessage("body", true)
+		Message chan Message
+	}
+
+	Message struct {
+		Data      api.Message
+		Metadata  any
+		Telemetry map[string]interface{}
 	}
```
mszostok, line 72:
```go
	StreamOutput struct {
		Event chan Event
	}

	Event struct {
		Message          api.Message
		Metadata         any
		AnonymizedFields map[string]interface{}
	}
```

WDYT about such terminology? @huseyinbabal @pkosiec ?

internal/source/kubernetes/commander/guard.go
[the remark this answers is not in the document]
huseyinbabal: I handled that in this PR

pkg/api/source/grpc_adapter.go
huseyinbabal: I can create a followup task for this, on my list. I also want to unify that, didn't wanted to confuse by having single config in sources and multiple configs in executors. Noted followup task.

internal/source/kubernetes/source.go
huseyinbabal: good catch

pkg/api/source/grpc_adapter.go
huseyinbabal: I am ok with the wording, changing accordingly

cmd/source/kubernetes/main.go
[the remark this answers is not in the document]
huseyinbabal: @mszostok is on it

cmd/source/kubernetes/main.go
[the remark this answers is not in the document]
huseyinbabal: https://github.com/kubeshop/botkube/issues/1000

pkg/api/source/grpc_adapter.go
huseyinbabal: Followup https://github.com/kubeshop/botkube/issues/1001

internal/source/kubernetes/config/config.go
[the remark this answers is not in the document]
huseyinbabal: changed like suggested

huseyinbabal:
LGTM with minor comments 🚀 

> Overall good work 🚀
> 
> Once all comments will be resolved I will play with it locally as Paweł did. Please also track all work that you would like to do in the follow-ups. Thanks!

I addressed all the feedbacks.

helm/botkube/values.yaml
```diff
@@ -135,260 +135,278 @@ sources:
...
+          # exclude: []
+
+        # -- Describes event constraints for Kubernetes resources.
+        # These constraints are applied for every resource specified in the `resources` list, unless they are overridden by the resource's own `events` object.
+        event:
+          # -- Lists all event types to be watched.
+          types:
+            - create
+            - delete
+            - error
```
mszostok, line 173: NOTE: after this merging this PR we won't merge configuration anymore. As a result, enabling both `k8s-err-events` and 'k8s-all-events' results in duplicated `error` messages. Probably will be good to mention that in the release notes/docs.

cmd/botkube/main.go
```diff
@@ -129,7 +128,7 @@ func run(ctx context.Context) error {
 	if err != nil {
 		return reportFatalError("while loading k8s config", err)
 	}
-	dynamicCli, discoveryCli, mapper, err := getK8sClients(kubeConfig)
+	_, discoveryCli, _, err := getK8sClients(kubeConfig)
```
mszostok, line 131: it will be good to get rid of not used clients

internal/source/kubernetes/source.go
```diff
@@ -0,0 +1,692 @@
...
+// Metadata returns metadata of Kubernetes configuration
+func (s *Source) Metadata(_ context.Context) (api.MetadataOutput, error) {
+	return api.MetadataOutput{
+		Version:     s.pluginVersion,
+		Description: description,
+		JSONSchema:  jsonSchema(),
+	}, nil
+}
+
+func consumeEvents(s Source, ctx context.Context) {
```
mszostok, line 104: context.Context always should be as a first parameter

pkg/execute/factory.go
```diff
@@ -7,19 +7,18 @@ import (
 	"github.com/slack-go/slack"
 
 	"github.com/kubeshop/botkube/internal/audit"
+	config2 "github.com/kubeshop/botkube/internal/command"
```
mszostok, line 10:
```suggestion
	guard "github.com/kubeshop/botkube/internal/command"
```

let's rename that to make it more readable

pkg/filterengine/filterengine.go
```diff
@@ -10,6 +10,11 @@ import (
 	"github.com/kubeshop/botkube/pkg/maputil"
```
mszostok, line 10: actually this file can be deleted, and in the test we should use `filters.DisableAnnotation` from k8s source. I will handle that on my PR, don't worry about that.

mszostok (approved):
LGTM 👍 but before merge please address my last comments, and this one too: https://github.com/kubeshop/botkube/pull/985#discussion_r1121533684

Also please wait for @pkosiec feedback and approve. Thanks!

pkosiec (approved):
🚀 Awesome work!

helm/botkube/values.yaml
huseyinbabal: Ok, I will be stating that in both release docs and main doc

huseyinbabal:
Addressed all the feedbacks, will check locally and focus on docs changes.

PR #988, opened 2023-02-21
Add default remote config provider endpoint

pkosiec:
## Description

- Add default remote config provider endpoint

### DIFF
--- a/internal/status/reporter.go
+++ b/internal/status/reporter.go
@@ -16,7 +16,7 @@ type StatusReporter interface {
 }
 
 func NewStatusReporter(logger logrus.FieldLogger, gql *graphql.Gql) StatusReporter {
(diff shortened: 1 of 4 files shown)


internal/status/reporter.go
```diff
@@ -16,7 +16,7 @@ type StatusReporter interface {
 }
 
 func NewStatusReporter(logger logrus.FieldLogger, gql *graphql.Gql) StatusReporter {
-	if _, provided := os.LookupEnv(graphql.GqlProviderEndpointEnvKey); provided {
+	if _, provided := os.LookupEnv(graphql.GqlProviderIdentifierEnvKey); provided {
```
pkosiec, line 19: If you think we should have additional env like `CONFIG_PROVIDER_ENABLED` or something like that, let me know 🙂

josefkarasek (approved):
lgtm :+1:

PR #1004, opened 2023-03-03
Add resource version for status reporting and update configuration types

pkosiec:
## Description

- Update approach for getting Botkube config
- Add resource version for status reporting
- Modify Botkube config to use it externally
   - Fix marshalling plugins under source / executors 
   - To allow configuring logger, I needed to make this config public.
   - I needed to reorganize the code to avoid import cycles.
- Add skeleton for deployment config updater
    - currently it updates resource version only for status reporting

## Testing

Run with remote config:
```
export CONFIG_PROVIDER_IDENTIFIER="c6100422-89a7-42ee-a0bf-ebcb24cbb785"
export CONFIG_PROVIDER_ENDPOINT="http://localhost:8080/graphql"
go run ./cmd/botkube/main.go
```


cmd/botkube/main.go
```diff
@@ -48,12 +49,13 @@ import (
...
-	commGroupFieldKey    = "commGroup"
-	healthEndpointName   = "/healthz"
-	printAPIKeyCharCount = 3
+	componentLogFieldKey  = "component"
+	botLogFieldKey        = "bot"
+	sinkLogFieldKey       = "sink"
+	commGroupFieldKey     = "commGroup"
+	healthEndpointName    = "/healthz"
+	printAPIKeyCharCount  = 3
+	configUpdaterInterval = 15 * time.Second
```
pkosiec, line 58: I will make this configurable in one of later PRs.

pkg/config/testdata/TestLoadConfigSuccess/config.golden.yaml
```diff
@@ -326,23 +326,22 @@ sources:
                 my-annotation: "true"
             labels:
                 my-label: "true"
-        plugins:
```
pkosiec, line 329: This was invalid nesting when marshalling config, which was fixed in this PR 🙂

internal/status/gql_reporter.go
```diff
@@ -38,12 +46,16 @@ func (r *GraphQLStatusReporter) ReportDeploymentStartup(ctx context.Context) (bo
...
 	var mutation struct {
-		Success bool `graphql:"reportDeploymentShutdown(id: $id)"`
+		Success bool `graphql:"reportDeploymentShutdown(id: $id, resourceVersion: $resourceVersion)"`
 	}
 	variables := map[string]interface{}{
-		"id": graphql.ID(r.gql.DeploymentID),
+		"id":              graphql.ID(r.gql.DeploymentID),
+		"resourceVersion": r.getResourceVersion(),
 	}
 	if err := r.gql.Cli.Mutate(ctx, &mutation, variables); err != nil {
```
josefkarasek, line 60: What will happen when we fail with wrong resource version? Shouldn't we try to fetch Deployment (outside of `interval`) and try again?

josefkarasek:
I see 2 issues:

* Botkube issues request with invalid resource version - we just fail. But we should try to get valid resource version and try again.
* Future extensibility - new component will modify remote Deployment and will get new resource version. How can this resource version be shared with other components, such as status reporter?

Both these issues stem from the _push_ semantics of distributing resource version.
Eg k8s approach is `list` & `watch` and `get` when update fails on invalid resource version.

internal/status/gql_reporter.go
pkosiec: Yes, I agree, I'll add retry for that 👍

pkosiec:
> Future extensibility - new component will modify remote Deployment and will get new resource version. How can this resource version be shared with other components, such as status reporter?

The idea is that Config Updater will update the resource version. See this snippet (`setResourceVersionForAll`): https://github.com/kubeshop/botkube/pull/1004/files#diff-d3a4641227ada1de58b084392d0235221af3fccff95a05fae276381fd76f13c9R92-R97

josefkarasek:
> > Future extensibility - new component will modify remote Deployment and will get new resource version. How can this resource version be shared with other components, such as status reporter?
> 
> The idea is that Config Updater will update the resource version. See this snippet (`setResourceVersionForAll`): https://github.com/kubeshop/botkube/pull/1004/files#diff-d3a4641227ada1de58b084392d0235221af3fccff95a05fae276381fd76f13c9R92-R97

That's every `interval:15s` - all _update_ operations will have to assume their resource version is invalid and include retry on failure. That's OK, but we should remember to do this retry check.

pkosiec:
@josefkarasek retry added - you can test it by starting the backend server, running Botkube to fetch the initial configuration and then quickly close it 😄 Or wait for startup reporting and then close Botkube to report shutdown.

josefkarasek (approved):
lgtm :+1:

PR #1005, opened 2023-03-03
Add event rendering for all platforms, add isInteractive to stream co…

mszostok:
…

## Description

- Add event rendering for all platforms, 
- Add isInteractive to stream context, 
- Fix e2e tests
- Update Prometheus source


mszostok:
FYI: I updated the emojis to be displayed well on all platforms. In the recorded demo you can see that for Teams the ⚠️ was not printed nicely. What's more the update it's not IMO a warning event but more info so I used 💡 :
old:
Screenshot 2023-03-07 at 14 39 48

new:
Screenshot 2023-03-07 at 14 41 31

huseyinbabal:
@mszostok done with product level testing among code level one, will finish review soon
Screen Shot 2023-03-08 at 12 26 41
Screen Shot 2023-03-08 at 12 26 50
Screen Shot 2023-03-08 at 12 26 58
Screen Shot 2023-03-08 at 12 27 03
Screen Shot 2023-03-08 at 12 46 22

pkg/api/source/grpc_adapter.go
```diff
@@ -49,14 +49,24 @@ type (
...
+		//	  out := source.StreamOutput{
+		//	  	Event: make(chan source.Event),
 		//	  }
+		//	  go func() {
+		//	  	out.Event <- source.Event{
+		//	  		Message: api.NewPlaintextMessage("Ticker Event", true),
+		//	  	}
+		//	  }()
+		//	  return out, nil
 		Output chan []byte
```
huseyinbabal, line 70: Is there any reason of still keeping this?

huseyinbabal (approved):
So much things done, thanks 🚀 . There is only one minor comment.

pkg/api/source/grpc_adapter.go
mszostok: that's a good question. We have 2 options and I don't have a strong opinion:
1. Mark it as deprecated and remove it in next release
2. Remove it already and change all related plugins

We need to do that for executor and source. 

Personally, I would state with the 1st option just to get used to at least one release deprecation policy.

@huseyinbabal @pkosiec WDYT?

huseyinbabal: Better to handle later then 🙂

Issue #745, opened 2022-09-21
Decrease cyclomatic complexity of main function


huseyinbabal:
## Description
Since the [main.go](https://github.com/kubeshop/botkube/blob/main/cmd/botkube/main.go) is too big, linter check fails due to cyclomatic complexity as shown below.

```
WARN [linters context] structcheck is disabled because of generics. You can track the evolution of the generics support by following the https://github.com/golangci/golangci-lint/issues/2649.
cmd/botkube/main.go:59:1: cyclomatic complexity 31 of func `run` is high (> 30) (gocyclo)
func run() error {
^
```

## Expected behavior

Linter check should succeed

## Actual behavior

It fails due to the cyclomatic complexity threshold

## Steps to reproduce

The threshold is increased to 35, it can be reproduced by setting it to 30 which is default value and run `make lint-fix`.

PR #1006, opened 2023-03-06
Reload Botkube on remote config change

pkosiec:
## Description

- Reload Botkube on remote config change

## Limitations

As the TODO comment states, right now Botkube will reload on every config change:

```
	// TODO: check if notifications are enabled and if so:
	//  - update notifications for a given channel (this needs a global state)
	//  - send message to a given channel (this needs a rework for the notifier executor)
	//    - updating notifications should happen after ConfigMap update, not before
	//    - same for remote config reloader
	//  - do not restart the app
```

This is something to address in future.

## Follow-ups
- https://github.com/kubeshop/botkube/issues/1011
- https://github.com/kubeshop/botkube/issues/1012


internal/config/reloader/restarter.go
```diff
@@ -0,0 +1,66 @@
...
+
+	r.log.Infof(`Reloading te the deployment "%s/%s"...`, r.deploy.Namespace, r.deploy.Name)
+	// This is what `kubectl rollout restart` does.
+	restartData := fmt.Sprintf(k8sDeploymentRestartPatchFmt, time.Now().String())
+	_, err = r.k8sCli.AppsV1().Deployments(r.deploy.Namespace).Patch(
+		ctx,
+		r.deploy.Name,
+		types.StrategicMergePatchType,
+		[]byte(restartData),
+		metav1.PatchOptions{FieldManager: "kubectl-rollout"},
```
mszostok, line 59: It looks like we should come up with our custom name: https://kubernetes.io/docs/reference/using-api/server-side-apply/#managers

internal/config/reloader/remote.go
```diff
@@ -0,0 +1,203 @@
...
+	"github.com/kubeshop/botkube/pkg/config"
+)
+
+// DeploymentClient defines GraphQL client.
+type DeploymentClient interface {
+	GetResourceVersion(ctx context.Context) (int, error)
+	GetConfigWithResourceVersion(ctx context.Context) (remote.Deployment, error)
+}
+
+// NewRemote returns new ConfigUpdater.
```
mszostok, line 23:
```suggestion
// NewRemote returns new RemoteConfigReloader.
```

internal/config/reloader/remote.go
```diff
@@ -0,0 +1,203 @@
...
+		currentCfg:    cfg,
+		resVersion:    cfgVer,
+		interval:      cfg.ConfigWatcher.Remote.PollInterval,
+		deployCli:     deployCli,
+		resVerHolders: resVerHolders,
+		restarter:     restarter,
+	}
+}
+
+type RemoteConfigReloader struct {
```
mszostok, line 36: not sure what's your approach on this PR but it's quite random when it comes to commenting exported struct/methods 😄

cmd/botkube/main.go
```diff
@@ -74,11 +74,11 @@ func run(ctx context.Context) error {
 	// Load configuration
 	intconfig.RegisterFlags(pflag.CommandLine)
 
-	remoteCfgSyncEnabled := graphql.IsRemoteConfigEnabled()
-	gqlClient := graphql.NewDefaultGqlClient()
-	deployClient := intconfig.NewDeploymentClient(gqlClient)
+	remoteCfg, remoteCfgEnabled := remote.GetConfig()
```
mszostok, line 77: we need to finally clean up this main func. We are creating real clients just to not use them in case of `remoteCfgEnabled`. 

With such pattern we will in a moment end-up with creating no-op implementation that we will pass to other "getter" to return no-op with no-op deps 🙈

mszostok (commented):
one of you will end-up with conflicts 😄 
https://github.com/kubeshop/botkube/pull/1007

mszostok (approved):
Tested e2e, works great 🚀

internal/config/reloader/remote.go
pkosiec: Whenever I see I didn't put a comment, I try to do it 😄 Will fix it - I think we have the linter errors silenced for internal 🤔 (the comments-related check)

cmd/botkube/main.go
pkosiec: True :/ I tried to improve things a bit, but I know it's not enough. Will sync with you tomorrow 👍

pkosiec: Ok, ensured that we don't create clients when remote config is enabled:

```go
	remoteCfg, remoteCfgEnabled := remote.GetConfig()
	var (
		gqlClient    *remote.Gql
		deployClient *remote.DeploymentClient
	)
	if remoteCfgEnabled {
		gqlClient = remote.NewDefaultGqlClient(remoteCfg)
		deployClient = remote.NewDeploymentClient(gqlClient)
	}
```

Let's follow-up on this in https://github.com/kubeshop/botkube/issues/745, as this is IMO quite connected

helm/botkube/templates/deployment.yaml
```diff
@@ -109,18 +111,12 @@ spec:
...
-            - name: BOTKUBE_SETTINGS_LIFECYCLE__SERVER_DEPLOYMENT_NAME
-              value: "{{ include "botkube.fullname" . }}"
             - name: BOTKUBE_CONFIG__WATCHER_DEPLOYMENT_NAMESPACE
               value: "{{.Release.Namespace}}"
             - name: BOTKUBE_CONFIG__WATCHER_DEPLOYMENT_NAME
               value: "{{ include "botkube.fullname" . }}"
-            {{- if .Values.config.provider.endpoint }}
             - name: CONFIG_PROVIDER_ENDPOINT
               value: {{ .Values.config.provider.endpoint }}
-            {{- end }}
```
joselcvarela, line 119: Hey guys,
Nice job with this bot, really helpful 👍

I just saw the `{{- if .Values.config.provider.endpoint }}` was removed.
Any particular reason for it?
This is an old commit but this change is still on HEAD.

Thank you.

mszostok: Hi @joselcvarela  👋 

I'm glad you like Botkube! 🚀 

Regarding this change, while we could maintain it, we're currently using the default value here:https://github.com/kubeshop/botkube/blob/main/helm/botkube/values.yaml#L1228.

As long as you don't remove or override it, it should render properly. 
However, this endpoint will only be utilized if the `CONFIG_PROVIDER_IDENTIFIER` is set; otherwise, it will be ignored, and the local configuration will be used. In such cases, no external calls are executed. You can find the code responsible for this [here](https://github.com/kubeshop/botkube/blob/875f88c3c8d2cb06d0e6c09b3e16f3f9f3d3ef69/internal/config/remote/config.go#L24-L34).

Did you encounter any problems with it?

jlcv-github-sandbox: Thank you for your response.
No, I haven't seen any issue so far.
I was just looking through the values passed to templates, and this part was intriguing me 👍
Keep on with the good job 🙌

joselcvarela: Thank you for your response.
No, I haven't seen any issue so far.
I was just looking through the values passed to templates, and this part was intriguing me 👍
Keep on with the good job 🙌

PR #1008, opened 2023-03-07
Webhook refactored after k8s plugin extraction

huseyinbabal:
## Description

Changes proposed in this pull request:

- Webhook sink is updated to accept new contract `api.Message`
- Introduced rawData param to send original data to sinks
- Collector is updated to enable plugins for also sinks

### Build plugins
```bash
PLUGIN_DOWNLOAD_URL_BASE_PATH=http://localhost:8080/plugin-dist make gen-plugins-index
```

### Run plugin server
```bash
npx serve --listen 8080
```

### Run Botkube with the following  config
ℹ️ You can use [Webhook Site Tester](https://webhook.site) for webhook URL

```bash
export BOTKUBE_CONFIG_PATHS=/tmp/botkube.yml
export BOTKUBE_SETTINGS_KUBECONFIG=${USER}/.kube/config

go run cmd/botkube/main.go
```

```yaml
## /tmp/botkube.yml
communications:
  'default-group':
    webhook:
      enabled: true
      url: 'webhook_url'
      bindings:
        sources: 
          - k8s-events
          - k8s-updates


settings:
  clusterName: gke-playground
  configWatcher: true
  log:
    level: debug
  lifecycleServer:
    deployment:
      name: botkube
      namespace: botkube
    port: "2113"

configWatcher:
  enabled: false
  initialSyncTimeout: 0

sources:
  'k8s-events':
    displayName: "K8s recommendations"
    'botkube/kubernetes':
      enabled: true
      config:
        log:
          level: debug
        recommendations:
          pod:
            noLatestImageTag: true
            labelsSet: true
          ingress:
            backendServiceValid: false
            tlsSecretValid: false
        namespaces:
          include:
            - botkube
        event:
          types:
            - create
            - update

  'k8s-updates':
    displayName: "K8s ConfigMaps updates"
    'botkube/kubernetes':
      enabled: true
      config:
        log:
          level: debug
        namespaces:
          include:
            - default
        event:
          types:
            - create
            - update
            - delete
        resources:
          - type: v1/configmaps
            namespaces:
              include:
                - botkube
            event: # overrides top level `event` entry
              types:
                - update
            updateSetting:
              includeDiff: true
              fields:
                - data

plugins:
  repositories:
    botkube:
      url: http://localhost:8080/plugins-index.yaml

```

### Create a pod to see a notification
```bash
kubectl run nginx_webhook --image=nginx:latest -n botkube
```

## Related issue(s)
Fixes #999

PR #1010, opened 2023-03-07
Elasticsearch sink refactor

### LINKED ISSUE #998: Refactor Elasticsearch sink
Elasticsearcy type mapping highly depends on kubernetes event.Event struct. As part of kubernetes plugin extraction, now contracted struct is api.Message and we need to refactor this mapping.

huseyinbabal:
## Description

- Elasticsearch sink is updated to accept new contract `api.Message`
- Introduced rawData param to send original data to sinks
- Collector is updated to enable plugins for also sinks

### Build plugins
```bash
PLUGIN_DOWNLOAD_URL_BASE_PATH=http://localhost:8080/plugin-dist make gen-plugins-index
```

### Run plugin server
```bash
npx serve --listen 8080
```

### Run Botkube with the following  config

```bash
export BOTKUBE_CONFIG_PATHS=/tmp/botkube.yml
export BOTKUBE_SETTINGS_KUBECONFIG=${USER}/.kube/config

go run cmd/botkube/main.go
```

```yaml
## /tmp/botkube.yml
communications:
  'default-group':
    elasticsearch:
      enabled: true
      server: 'elastic_url'
      username: user
      password: pass
      indices:
        'default':
          name: botkube
          type: kubernetes
          shards: 1
          replicas: 1
          bindings:
            sources: 
              - k8s-events
              - k8s-updates


settings:
  clusterName: gke-playground
  configWatcher: true
  log:
    level: debug
  lifecycleServer:
    deployment:
      name: botkube
      namespace: botkube
    port: "2113"

configWatcher:
  enabled: false
  initialSyncTimeout: 0

sources:
  'k8s-events':
    displayName: "K8s recommendations"
    'botkube/kubernetes':
      enabled: true
      config:
        log:
          level: debug
        recommendations:
          pod:
            noLatestImageTag: true
            labelsSet: true
          ingress:
            backendServiceValid: false
            tlsSecretValid: false
        namespaces:
          include:
            - botkube
        event:
          types:
            - create
            - update

  'k8s-updates':
    displayName: "K8s ConfigMaps updates"
    'botkube/kubernetes':
      enabled: true
      config:
        log:
          level: debug
        namespaces:
          include:
            - default
        event:
          types:
            - create
            - update
            - delete
        resources:
          - type: v1/configmaps
            namespaces:
              include:
                - botkube
            event: # overrides top level `event` entry
              types:
                - update
            updateSetting:
              includeDiff: true
              fields:
                - data

plugins:
  repositories:
    botkube:

```

### Create a pod to see a notification
```bash
kubectl run nginx_es --image=nginx:latest -n botkube
```

## Related issue(s)
Fixes #998

### DIFF
--- a/pkg/sink/elasticsearch.go
+++ b/pkg/sink/elasticsearch.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"crypto/tls"
 	"fmt"
+	"github.com/kubeshop/botkube/pkg/sliceutil"
 	"net/http"
 	"os"
 	"time"
@@ -172,11 +173,14 @@ func (e *Elasticsearch) flushIndex(ctx context.Context, indexCfg config.ELSIndex
 }
 
 // SendEvent sends an event to a configured elasticsearch server.
-func (e *Elasticsearch) SendEvent(ctx context.Context, rawData any, _ []string) error {
+func (e *Elasticsearch) SendEvent(ctx context.Context, rawData any, sources []string) error {
 	e.log.Debugf(">> Sending to Elasticsearch: %+v", rawData)
 
 	errs := multierror.New()
 	for _, indexCfg := range e.indices {
+		if !sliceutil.Intersect(indexCfg.Bindings.Sources, sources) {
+			continue
+		}
 		err := e.flushIndex(ctx, indexCfg, rawData)
 		if err != nil {
 			errs = multierror.Append(errs, fmt.Errorf("while sending event to Elasticsearch index %q: %w", indexCfg.Name, err))
--- a/pkg/sink/elasticsearch.go
+++ b/pkg/sink/elasticsearch.go
@@ -80,6 +80,7 @@ func NewElasticsearch(log logrus.FieldLogger, c config.Elasticsearch, reporter A
 			elastic.SetSniff(false),
 			elastic.SetHealthcheck(false),
 			elastic.SetGzip(false),
+			elastic.SetTraceLog(log),
 		)
 		if err != nil {
 			return nil, fmt.Errorf("while creating new Elastic client: %w", err)
@@ -91,6 +92,7 @@ func NewElasticsearch(log logrus.FieldLogger, c config.Elasticsearch, reporter A
 			elastic.SetSniff(false),
 			elastic.SetHealthcheck(false),
 			elastic.SetGzip(true),
+			elastic.SetTraceLog(log),
 		}
 
 		if c.SkipTLSVerify {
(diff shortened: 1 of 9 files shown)


pkg/sink/elasticsearch.go
```diff
@@ -174,38 +172,29 @@ func (e *Elasticsearch) flushIndex(ctx context.Context, indexCfg config.ELSIndex
...
+	e.log.Debugf(">> Sending to Elasticsearch: %+v", rawData)
 
 	errs := multierror.New()
 	for _, indexCfg := range e.indices {
-		if !sliceutil.Intersect(indexCfg.Bindings.Sources, eventSources) {
-			continue
-		}
-
-		err := e.flushIndex(ctx, indexCfg, event)
+		err := e.flushIndex(ctx, indexCfg, rawData)
```
pkosiec, line 186: That's might be problematic as I know we needed to ignore some properties from the `event.Event` as ELS didn't accept some of the nested struct data. Just saying, but it should work OK as long as all events have that in mind 👍

pkosiec (commented):
Please see the review of https://github.com/kubeshop/botkube/pull/1008. Thanks!

pkg/sink/elasticsearch.go
huseyinbabal: I tested this for kubernetes plugin and the data was shipped to elasticsearch, but I think we need to state this in the docs maybe, "we don't have control on data structure, so if you have es integration, you might get error due to your complex nested data structure coming from custom plugin".

pkg/sink/elasticsearch.go
```diff
@@ -196,8 +196,21 @@ func (e *Elasticsearch) SendMessage(ctx context.Context, event event.Event, even
 }
 
 // SendEvent is no-op.
-func (e *Elasticsearch) SendEvent(_ context.Context, _ any, _ []string) error {
-	return nil
+func (e *Elasticsearch) SendEvent(ctx context.Context, rawData any, sources []string) error {
+	e.log.Debugf(">> Sending to Elasticsearch: %+v", rawData)
+
+	errs := multierror.New()
+	for _, indexCfg := range e.indices {
```
pkosiec, line 203: This logic is invalid - we send events to all indexes and `sources` are not used. We should do

```go
		if !sliceutil.Intersect(indexCfg.Bindings.Sources, eventSources) {
			continue
		}
```
instead.

Basically please use logic from `SendMessage` in `SendEvent`, and remove `SendMessage`. Thanks!

pkosiec (approved):
LGTM but please apply comments before merge. Thanks!

pkg/sink/elasticsearch.go
huseyinbabal: Actually, I removed SendMessage now, and each es flush index is for one source, so I removed source part

PR #1014, opened 2023-03-09
Enable actions even if not bind to a specific channel

### LINKED ISSUE #1000: Enable plugin for actions
Currently, the plugins are enabled if they are binded to communication channels. Actions have dedicated bindings section, but we are not enabling them if they are not in the communication channels. So the task;

- While scheduling plugins, check also `actions` value and enable plugins if you see an entry
- They might not be in communication bindings, the action command can be silently handled.
- Ensure recommendation work if no other resources are listed (all recommendations are true, but resources list are empty - the plugin should add pods and ingress listing without user involvement)
- Add E2E test - action is bound to a source which is not bound to other channels
   - on incoming event, execute `kubectl label` or `annotate` and check if the label/annotation was added after a while

mszostok:
## Description

- Enable actions even if not bind to a specific channel

## Testing

Covered by e2e tests.

## Related issue(s)

Fix #1000


internal/plugin/collector.go
```diff
@@ -86,28 +100,39 @@ func (c *Collector) GetAllEnabledAndUsedPlugins(cfg *config.Config) ([]string, [
...
 			}
 
-			usedSourcePlugins = append(usedSourcePlugins, name)
+			l.Debug("Marking source plugin as enabled")
+			usedSourcePlugins[name] = struct{}{}
 		}
 	}
-	return usedExecutorPlugins, usedSourcePlugins
+
+	return maps.Keys(usedExecutorPlugins), maps.Keys(usedSourcePlugins)
```
pkosiec, line 137: Can you also use `maps.Keys` in `maputil.SortKeys`? Thanks!

I also searched for `sort.Strings` and saw other places where we can use it - e.g. in kubectl_cmd_builder.go:

```go
	for key := range enabledKubectls.AllowedKubectlResource {
		resources = append(resources, key)
	}
	sort.Strings(resources)

	for key := range enabledKubectls.AllowedKubectlVerb {
		verbs = append(verbs, key)
	}
	sort.Strings(verbs)
```

pkosiec (approved):
LGTM 👍

internal/plugin/collector.go
mszostok: the `kubectl_cmd_builder.go` will be removed soon so I just skipped but I can change it to 👍

pkosiec: Skip it then 👍 I forgot that this is the legacy builder.

Issue #934, opened 2023-01-11
Handle passing Kubeconfig to plugins (Executor and Sources)


pkosiec:
## Acceptance Criteria

Handle Kubeconfig passing for executors:

   - Generate and pass Kubeconfig as a part of Context for each execution of a Kubernetes-related executor. 
   - ~~Modify plugin manager to run plugins in a restricted environment.~~
      - ~~It doesn't influence the way how we run source and executor plugins. Each plugin will be ran in a single subprocess, in a single restricted subdirectory.~~
   - Ensure already existing plugins support it.
       - ConfigMap Source
       - Helm plugin
       - #841 + #840, if they are already done 
   - Add necessary tests
     - ~~use mocked file system (?) (https://github.com/spf13/afero)~~
     - testing "scheduler" that spawns plugins.
   - Make sure that plugins works by default in the backward compatible way.
   - Take into account the default k8s Namespace when creating a dedicated kubeconfig
   - Make sure that `rbac` property is optional
      - If the rbac context is not provided (as a result, rbac.user|group.type is empty), Botkube won't create a temporary Kubeconfig for a given plugin.

Relates to #841 and #840. Depending which tasks will be done earlier, ensure a proper Kubeconfig is passed.

Resources:
- See [the RBAC proposal](https://github.com/kubeshop/botkube/blob/main/docs/proposal/2022-12-23-rbac.md#plugin-isolation)
- [POC implementation](https://github.com/kubeshop/botkube/pull/909/files#diff-d29b5dae5d8d707b61f73c2330ccab2e4e787ebd554d054bbcd11b04cb2e9043R345)
- solve the issue described in [the investigation doc](https://github.com/kubeshop/botkube/tree/main/docs/investigation/rbac#process-isolation)

## Related issues & reason

- https://github.com/kubeshop/botkube/issues/712
- https://github.com/kubeshop/botkube/issues/1021 since, it is not doable to introduce isolation for now.

## Notes

Strategy for RBAC isolation:
1. Dynamically create subdirectories for a given execution
    - Plugins **MUST** respect a given kubeconfig path
2. Just start multiple process for all unique RBAC binding:
    - if `kubectl` is bind to 2 different channels and `group` strategy is used, we simply start 2 `kubectl` processes.
    - please check the cost of starting those processes (CPU/Memory)

Issue #945, opened 2023-01-18
Add e2e tests coverage for plugins RBAC


mszostok:
# Description

Add e2e tests coverage for plugins RBAC. This task depends on:
- https://github.com/kubeshop/botkube/issues/934
- https://github.com/kubeshop/botkube/issues/933

# AC

- The happy path flow:
  - test per channel RBAC
  - test static user mapping
  - test when RBAC is not provided at all
  - test that the default configuration works (it's will be a side effect of proper migration in https://github.com/kubeshop/botkube/issues/933)

## Related issues & reason

https://github.com/kubeshop/botkube/issues/712

PR #1020, opened 2023-03-20
Generate kube config based on plugin RBAC

josefkarasek:
## Description

- Add `kubeConfig string` to executor context
- Generate kube config per plugin invocation based on plugin RBAC

## Related issue(s)

https://github.com/kubeshop/botkube/issues/934


cmd/botkube/main.go
```diff
@@ -104,6 +104,11 @@ func run(ctx context.Context) error {
 		logger.Warnf("Configuration validation warnings: %v", confDetails.ValidateWarnings.Error())
 	}
 
+	restCfg, err := rest.InClusterConfig()
```
josefkarasek, line 107: This will fail when not run in a kubernetes pod. Add a flag/switch, eg rbac defined?

pkg/execute/plugin_executor.go
```diff
@@ -82,6 +95,7 @@ func (e *PluginExecutor) Execute(ctx context.Context, bindings []string, slackSt
 		Context: executor.ExecuteInputContext{
 			IsInteractivitySupported: cmdCtx.Platform.IsInteractive(),
 			SlackState:               slackState,
+			KubeConfig:               kubeconfig,
```
josefkarasek, line 98: Should all executor plugins get a kubeconfig?
If we add a new field to `api.Metadata`, can plugin executor access that field?

cmd/botkube/main.go
huseyinbabal: I did same thing here https://github.com/kubeshop/botkube/blob/3e1d8a99c8a18244d2203c12a42509e9b7799280/internal/source/kubernetes/client.go#L28

pkg/execute/plugin_executor.go
pkosiec: As it is in AC of the task:

> Make sure that rbac property is optional
>    If the rbac context is not provided (as a result, rbac.user|group.type is empty), Botkube won't create a temporary Kubeconfig for a given plugin.

So no, if context.rbac is not provided, we won't create kubeconfig. So this property should be optional (probably a pointer).

pkg/execute/plugin_executor.go
```diff
@@ -234,3 +248,39 @@ func (e *PluginExecutor) getEnabledPlugins(bindings []string, cmdName string) ([
 
 	return out, fullPluginName
 }
+
+func (e *PluginExecutor) generateKubeConfig(rbac config.PolicyRule) (string, error) {
```
pkosiec, line 252: Remember that this kubeconfig should be deleted after execution. And yes, for now we should support:

- static mapping for both users and groups
- channel-based group mapping.

proto/executor.proto
```diff
@@ -23,6 +23,7 @@ message ExecuteRequest {
 message ExecuteContext {
 	bool isInteractivitySupported = 1;
 	bytes slackState = 2;
+	string kubeConfig = 3;
```
pkosiec, line 26: Remember that kubeconfig should be also passed to Source plugins as a part of the source plugin process launch.

So: 
for executor - generate & pass kubeconfig per command execution
for source - generate & pass kubeconfig before starting the source plugin binary

pkg/execute/plugin_executor.go
mszostok: discussed f2f, for now we can don't remove them as it may generates additional problems, e.g.:
- multiple users call the executor (create dedicated files? make sure that we use the same one?)
- clean-up logic: can we remove file? what if it's still in use by other executor call?

so maybe we can skip it to just simplify the first implementation and create a follow-up task that we can refine and impl later if needed.

proto/source.proto
```diff
@@ -20,8 +20,11 @@ message StreamRequest {
 
 message StreamContext {
 	bool isInteractivitySupported = 1;
+	// kubeConfig is path to kubeConfig on local FS
 	string kubeConfig = 2;
```
josefkarasek, line 24: This isn't used anywhere.

internal/source/dispatcher.go
```diff
@@ -96,13 +99,18 @@ func (d *Dispatcher) Dispatch(dispatch PluginDispatch) error {
...
+		return fmt.Errorf("while generating kube config for %s: %w", dispatch.pluginName, err)
+	}
+
 	ctx := dispatch.ctx
 	out, err := sourceClient.Stream(ctx, source.StreamInput{
 		Configs: dispatch.pluginConfigs,
 		Context: source.StreamInputContext{
 			IsInteractivitySupported: dispatch.isInteractivitySupported,
 			ClusterName:              dispatch.cfg.Settings.ClusterName,
-			KubeConfig:               dispatch.cfg.Settings.Kubeconfig,
```
josefkarasek, line 105: This might break existing functionality. We might as well remove kube config from helm

helm/botkube/values.yaml
```diff
@@ -116,6 +116,25 @@ sources:
     # -- Describes Kubernetes source configuration.
     # @default -- See the `values.yaml` file for full object.
     botkube/kubernetes:
+      context: &defaultExecutorContext
```
huseyinbabal, line 119: name `defaultExecutoContext` seems confusing under source plugin

huseyinbabal:
@josefkarasek could you please share an rbac config for channel based configuration? I am doing e2e test, I only managed to handle static one

proto/source.proto
pkosiec: What do you mean by this comment? Shouldn't it be used? We need to pass kubeconfig to source plugin somehow, right? it would be good to pass it explicitly in grpc api same as for executors.

pkg/pluginx/kubeconfig.go
```diff
@@ -0,0 +1,34 @@
+package pluginx
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"path/filepath"
+)
+
+func PersistKubeConfig(ctx context.Context, kc []byte) (string, func(context.Context) error, error) {
```
pkosiec, line 10: Please wrap all errors

helm/botkube/values.yaml
```diff
@@ -141,6 +160,7 @@ sources:
     # -- Describes Kubernetes source configuration.
     # @default -- See the `values.yaml` file for full object.
     botkube/kubernetes:
+      context: *defaultExecutorContext
```
huseyinbabal, line 163: confusing default value

internal/plugin/kubeconfig.go
```diff
@@ -0,0 +1,80 @@
...
+func generateUserSubject(rbac config.UserPolicySubject) (user string) {
+	switch rbac.Type {
+	case config.StaticPolicySubjectType:
+		user = rbac.Prefix + rbac.Static.Value
+	}
+	return
+}
+
+func generateGroupSubject(rbac config.GroupPolicySubject) (group []string) {
+	switch rbac.Type {
```
pkosiec, line 73: When exactly the channel name based group mapping will be supported? I know that it'll block the https://github.com/kubeshop/botkube/issues/945 task.

internal/executor/helm/executor.go
```diff
@@ -104,37 +104,45 @@ func (e *Executor) Execute(ctx context.Context, in executor.ExecuteInput) (execu
 		in.Command = fmt.Sprintf("%s -n %s", in.Command, cfg.DefaultNamespace)
 	}
 
+	kubeConfigPath, deleteFn, err := pluginx.PersistKubeConfig(ctx, in.Context.KubeConfig)
+	if err != nil {
+		return executor.ExecuteOutput{}, fmt.Errorf("while writing kubeConfig file: %w", err)
+	}
+	defer func() {
+		_ = deleteFn(ctx)
```
pkosiec, line 112: can we log the delete error please?

pkg/api/executor/grpc_adapter.go
```diff
@@ -36,6 +36,9 @@ type (
 		// IsInteractivitySupported is set to true only if communication platform supports interactive Messages.
 		IsInteractivitySupported bool
 
+		// KubeConfig is the path to kubectl configuration file.
```
huseyinbabal, line 39: Better to use something like `// KubeConfig is the slice of byte representation of kubeconfig file content`

proto/source.proto
josefkarasek: I added a new field (`[]byte`) in the spirit of not breaking existing API.

Since we don't have any 3rd party plugins depending on existing `kubeconfig` field, we can think about making breaking change.

https://github.com/kubeshop/botkube/blob/069937a024add2f4a35a9c5455360702a7715d1b/proto/source.proto#L21-L28

internal/source/dispatcher.go
mszostok: 2 cents from my side: I like that option however, even Argo Workflow which is quite popular project, it doesn't support external/multi-cluster (they have issue opened for more that 2years now). If someone wants to have such functionality, they can always use https://github.com/virtual-kubelet/virtual-kubelet#providers. Of course, it brings additional maintainer cost, but it's also a more powerful and “safer” option.

Maybe at the end it just brings not necessary complexity to the Botkube itself? Probably it's a question to @brampling whether we should leave that functionality or not.

---

However, if we will decide to leave that functionality, it would be good to name it like `context.remote.kubeconfig`? so it is easier to distinguish current cluster RBAC usage vs external one.

proto/source.proto
pkosiec: > I added a new field ([]byte) in the spirit of not breaking existing API.

Let's break it, we already had multiple breaking changes 👍

PR #1033, opened 2023-03-31
skip k8s insight collection if remoteconfig is disabled

huseyinbabal:
## Description

- k8s insight collection is started even with OSS installation. Added condition to not start

## Testing

1. Run helm botkube with debug log enabled
2. You shouldn't see `time="2023-03-31T11:11:15Z" level=debug msg="Collecting Kubernetes insights"`

### DIFF
--- a/cmd/botkube/main.go
+++ b/cmd/botkube/main.go
@@ -388,7 +388,7 @@ func run(ctx context.Context) error {
 		}()
 		heartbeatReporter := heartbeat.GetReporter(remoteCfgEnabled, logger, gqlClient)
 		k8sCollector := insights.NewK8sCollector(k8sCli, heartbeatReporter, logger, reportHeartbeatInterval, reportHeartbeatMaxRetries)
-		return k8sCollector.Start(ctx)
+		return k8sCollector.Start(ctx, remoteCfgEnabled)
 	})
 
 	healthChecker.MarkAsReady()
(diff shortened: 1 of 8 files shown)


cmd/botkube/main.go
```diff
@@ -388,7 +388,7 @@ func run(ctx context.Context) error {
 		}()
 		heartbeatReporter := heartbeat.GetReporter(remoteCfgEnabled, logger, gqlClient)
 		k8sCollector := insights.NewK8sCollector(k8sCli, heartbeatReporter, logger, reportHeartbeatInterval, reportHeartbeatMaxRetries)
-		return k8sCollector.Start(ctx)
```
mszostok, line 391: why not to simply do the 
```go
	if !remoteCfgEnabled {
		k.logger.Debug("Remote config is not enabled, skipping k8s insights collection...")
		return nil
	}
```

at the beginning of the func? As a result also the `heartbeatReporter` and `k8sCollector` won't be created at all. Also the  `heartbeat.GetReporter` could be simplified.

huseyinbabal: How can't I see that :) on it...

huseyinbabal: Updated, please check

huseyinbabal: @mszostok as you suggested, I also simplified reporter

mszostok (approved):

Issue #1041, opened 2023-04-14
Kubernetes' recommendation settings ignore the namespace restriction


mszostok:
## Description

With such configuration:
```yaml
  'k8s-backend-err-events':
    displayName: "Kubernetes Errors"
    botkube/kubernetes:
      context: &default-plugin-context
        defaultNamespace: "default"
        rbac:
          group:
            type: Static
            static:
              values: [ "botkube-plugins-default" ]
      enabled: true
      config:
        recommendations: &recommendations
          pod:
            noLatestImageTag: true
            labelsSet: true
          ingress:
            backendServiceValid: true
            tlsSecretValid: true

        namespaces:
          include:
            - "backend"
        event:
          types:
            - error
        resources: &err-resources
          - type: v1/pods
          - type: v1/services
          - type: networking.k8s.io/v1/ingresses
```

we still get recommendation about resources from all namespace.

## Expected behavior

I get only recommendations for namespaces specified under `namespaces` property.

PR #1054, opened 2023-05-09
Use namespace filter in Recommendations

josefkarasek:
## Description

- Respect Kubernetes' recommendation settings

## Testing

Define source with recommendations 

```yaml
sources:
  'k8s-recommendation-events':
    displayName: "Kubernetes Recommendations"
    botkube/kubernetes:
      context: *default-plugin-context
      enabled: true
      config:
        namespaces:
          include:
            - "default"
        recommendations:
          pod:
            noLatestImageTag: true
            labelsSet: true
          ingress:
            backendServiceValid: true
            tlsSecretValid: true
```

Check that included namespaces work:

```
$ k delete pod nginx -n default; k run nginx --image=nginx -n default
```

Check that other namespaces are filtered out:

```
$ k delete pod nginx -n botkube; k run nginx --image=nginx -n botkube
```

Override config for `pods`:

```yaml
sources:
  'k8s-recommendation-events':
    displayName: "Kubernetes Recommendations"
    botkube/kubernetes:
      context: *default-plugin-context
      enabled: true
      config:
        namespaces:
          include:
            - "default"
        event:
          types:
            - create
        resources:
          - type: v1/pods
            namespaces:
              include:
                - "botkube"
        recommendations:
          pod:
            noLatestImageTag: true
            labelsSet: true
          ingress:
            backendServiceValid: true
            tlsSecretValid: true
```

Use same commands as before. Now the `default` ns should be filtered out.

## Related issue(s)

https://github.com/kubeshop/botkube/issues/1041

### DIFF
--- a/internal/source/kubernetes/router.go
+++ b/internal/source/kubernetes/router.go
@@ -176,7 +176,7 @@ func (r *Router) mergeEventRoutes(resource string, cfg *config.Config) map[confi
 			}
 
 			route := route{
-				namespaces:   resourceNamespaces(cfg.Namespaces, r.Namespaces),
+				namespaces:   resourceNamespaces(cfg.Namespaces, &r.Namespaces),
 				annotations:  resourceStringMap(cfg.Annotations, r.Annotations),
 				labels:       resourceStringMap(cfg.Labels, r.Labels),
 				resourceName: r.Name,
@@ -222,7 +222,7 @@ func (r *Router) setEventRouteForRecommendationsIfShould(routeMap *map[config.Ev
 	// Override route and get all these events for all namespaces.
 	// The events without recommendations will be filtered out when sending the event.
 	for i, r := range (*routeMap)[eventType] {
-		recommRoute.namespaces = resourceNamespaces(cfg.Namespaces, *r.namespaces)
+		recommRoute.namespaces = resourceNamespaces(cfg.Namespaces, r.namespaces)
 		(*routeMap)[eventType][i] = recommRoute
 		return
 	}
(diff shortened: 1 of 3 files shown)


internal/source/kubernetes/router.go
```diff
@@ -194,12 +194,12 @@ func (r *Router) mergeEventRoutes(resource string, cfg *config.Config) map[confi
...
 	// add routes related to recommendations
 	resForRecomms := recommendation.ResourceEventsForConfig(cfg.Recommendations)
-	r.setEventRouteForRecommendationsIfShould(&out, resForRecomms, resource)
+	r.setEventRouteForRecommendationsIfShould(&out, resForRecomms, resource, cfg)
 
 	return out
 }
 
-func (r *Router) setEventRouteForRecommendationsIfShould(routeMap *map[config.EventType][]route, resForRecomms map[string]config.EventType, resourceType string) {
+func (r *Router) setEventRouteForRecommendationsIfShould(routeMap *map[config.EventType][]route, resForRecomms map[string]config.EventType, resourceType string, cfg *config.Config) {
```
pkosiec, line 202: 1. Can we cover this change with integration tests? We already test the recommendations, but now let's ensure that we get only the events for a given namespace. So it could be just an extension of https://github.com/kubeshop/botkube/blob/main/test/e2e/bots_test.go#L978 - create same resource but in different ns and expect there's no recommendation notification about it.
2. Please also update docs to highlight that the recommendations now work like this. I think just a single sentence will do the trick: https://docs.botkube.io/configuration/source/kubernetes#recommendations

internal/source/kubernetes/router.go
```diff
@@ -224,7 +222,7 @@ func (r *Router) setEventRouteForRecommendationsIfShould(routeMap *map[config.Ev
 	// Override route and get all these events for all namespaces.
 	// The events without recommendations will be filtered out when sending the event.
 	for i, r := range (*routeMap)[eventType] {
-		recommRoute.namespaces = r.namespaces
+		recommRoute.namespaces = resourceNamespaces(cfg.Namespaces, *r.namespaces)
```
pkosiec, line 225: can we check if `r is nil first? just to be on the safe side. Thanks!

pkosiec (commented):
Overall looks good, didn't test it though. Two small things!

pkosiec:
@josefkarasek I know it's unrelated, but can you also remove line 210 from  `fmt.Println("m.resourceVersion", m.resourceVersion)` from `manager_remote.go`? Thanks 🙇

internal/source/kubernetes/router.go
josefkarasek: I changed func signature so that both args are pointers and we check for in in one place.

pkosiec (approved):
LGTM - I fully trust the integration test 👍

Issue #1045, opened 2023-04-17
Automated actions don't work with the `ChannelName` RBAC


mszostok:
## Overview

With such example configuration:
```yaml
executors:
  k8s-cli-tools:
    botkube/helm:
      enabled: true
      context:
        rbac:
          group:
            type: ChannelName

actions:
  'check-helm-release':
    enabled: true
    displayName: Display Helm release notes
    command: helm get notes {{ trimPrefix "sh.helm.release.v1." (trimSuffix ".v1"  .Event.Name) }} -n {{.Event.Namespace}}
    bindings:
      sources:
        - detect-helm-release
      executors:
        - k8s-cli-tools
```

You will observe that the action was rendered and executed. However, you will see that you don't have permissions to run it.

## Acceptance Criteria

1. Detect and block `ChannelName` usage in executor configuration.
  -  return human-readable error with suggestion how to fix it.
2. Update documentation to describe our RBAC policy for actions.

### Later

Wait for user feedback and if requested add support for the `ChannelName` RBAC is used for executor under actions. Maybe it would be worth to get the channel name directly from triggered source. However, let's wait for real use-cases.

## Reason

Improve UX.

PR #1065, opened 2023-05-15
Block actions with ChannelName executors

josefkarasek:
## Description

- Block actions with ChannelName executors

## Testing

Define a executor with `ChannelName` RBAC policy:

```yaml
executors:
  k8s-default-tools:
    botkube/helm:
      enabled: true
      config:
        defaultNamespace: "default"
        helmDriver: "secret"
        helmConfigDir: "/tmp/helm/"
      context: &default-plugin-context
        defaultNamespace: "default"
        rbac:
          group:
            type: ChannelName

actions:
  'describe-created-resource':
    enabled: true
    displayName: "Describe created resource"
    command: "kubectl describe {{ .Event.Kind | lower }}{{ if .Event.Namespace }} -n {{ .Event.Namespace }}{{ end }} {{ .Event.Name }}"
    bindings:
      sources:
        - k8s-create-events
      executors:
        - k8s-default-tools
```

Such action config should be blocked by config validator.

## Related issue(s)

https://github.com/kubeshop/botkube/issues/1045

### DIFF
--- a/pkg/config/validator.go
+++ b/pkg/config/validator.go
@@ -370,10 +370,11 @@ func validateActionExecutors(sl validator.StructLevel, executors map[string]Exec
 			if !plugin.Enabled {
 				continue
 			}
-			if plugin.Context.RBAC != nil {
-				if plugin.Context.RBAC.Group.Type == ChannelNamePolicySubjectType {
-					sl.ReportError(bindings, pluginKey, executor, invalidActionRBACTag, "")
-				}
+			if plugin.Context.RBAC == nil {
+				continue
+			}
+			if plugin.Context.RBAC.Group.Type == ChannelNamePolicySubjectType {
+				sl.ReportError(bindings, pluginKey, executor, invalidActionRBACTag, "")
 			}
 		}
 	}
--- a/pkg/config/validator.go
+++ b/pkg/config/validator.go
@@ -125,7 +125,7 @@ func registerBindingsValidator(validate *validator.Validate, trans ut.Translator
 		conflictingPluginVersionTag: "{0}{1}",
 		invalidPluginDefinitionTag:  "{0}{1}",
 		invalidPluginRBACTag:        "Binding is referencing plugins of same kind with different RBAC. '{0}' and '{1}' bindings must be identical when used together.",
-		invalidActionRBACTag:        "Plugin {0} has 'ChannelName' RBAC policy. This is not supported for actions.",
+		invalidActionRBACTag:        "Plugin {0} has 'ChannelName' RBAC policy. This is not supported for actions. See https://docs.botkube.io/configuration/action#rbac",
 	})
 }
 
--- a/pkg/config/validator.go
+++ b/pkg/config/validator.go
@@ -23,6 +23,7 @@ const (
 	invalidPluginDefinitionTag  = "invalid_plugin_definition"
 	invalidAliasCommandTag      = "invalid_alias_command"
 	invalidPluginRBACTag        = "invalid_plugin_rbac"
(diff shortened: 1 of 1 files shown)


pkg/config/validator.go
```diff
@@ -360,6 +363,22 @@ func validatePluginRBAC[P pluginProvider](sl validator.StructLevel, pluginConfig
...
+		execConf := executors[executor]
+		for pluginKey, plugin := range execConf.Plugins {
+			if !plugin.Enabled {
+				continue
+			}
+			if plugin.Context.RBAC != nil {
+				if plugin.Context.RBAC.Group.Type == ChannelNamePolicySubjectType {
+					sl.ReportError(bindings, pluginKey, executor, invalidActionRBACTag, "")
+				}
+			}
```
mszostok, line 377: we can easily get rid of nested `if's`
```suggestion
			if plugin.Context.RBAC == nil {
				continue
			}
			if plugin.Context.RBAC.Group.Type == ChannelNamePolicySubjectType {
				sl.ReportError(bindings, pluginKey, executor, invalidActionRBACTag, "")
			}
```

BTW, maybe this should be a part of `validateExecutorBindings`? Just a suggestion but not pushing into it 👍

mszostok (commented):
Overall LGTM but I will leave final testing and approving for selected reviewer 👍

pkg/config/validator.go
josefkarasek: `validateExecutorBindings` is also used by `botBindingsStructValidator`, so I opted for a brand new func.

josefkarasek:
Example faulty actions:
```
2023/05/15 14:53:51 while merging app configuration: found critical validation errors: 4 errors occurred:
	* Key: 'Config.Actions[describe-created-resource].Bindings.k8s-default-tools' Plugin botkube/kubectl has 'ChannelName' RBAC policy. This is not supported for actions.
	* Key: 'Config.Actions[describe-created-resource].Bindings.k8s-default-tools' Plugin botkube/helm has 'ChannelName' RBAC policy. This is not supported for actions.
	* Key: 'Config.Actions[show-logs-on-error].Bindings.k8s-default-tools' Plugin botkube/kubectl has 'ChannelName' RBAC policy. This is not supported for actions.
	* Key: 'Config.Actions[show-logs-on-error].Bindings.k8s-default-tools' Plugin botkube/helm has 'ChannelName' RBAC policy. This is not supported for actions.
```

huseyinbabal:
As a general comment, so user just saw that message after they use actions with executors with channel RBAC, then what to do next? If they duplicate that without channel RBAC to use with actions, it will be a security concern for them. If we will introduce a support for actions then better to state that maybe? Any valuable link to our existing actions page would be valuable, of course we need to have a scenario for this.

pkg/config/validator.go
```diff
@@ -360,6 +363,23 @@ func validatePluginRBAC[P pluginProvider](sl validator.StructLevel, pluginConfig
...
+	for _, executor := range bindings {
+		execConf := executors[executor]
+		for pluginKey, plugin := range execConf.Plugins {
+			if !plugin.Enabled {
+				continue
+			}
+			if plugin.Context.RBAC == nil {
+				continue
+			}
+			if plugin.Context.RBAC.Group.Type == ChannelNamePolicySubjectType {
```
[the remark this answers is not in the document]
josefkarasek: `Group` is a struct, but `RBAC` is a pointer.

https://github.com/kubeshop/botkube/blob/bf1d111dc3df67c078e843f330a3dae21d3a2446/pkg/config/config.go#L216-L227

huseyinbabal (approved):

Issue #1027, opened 2023-03-28
Botkube killed when plugin download takes too long


1d9akash:
## Description

Tried To Setup Botkube In Minikube. Faced Some Issues:
Setting Up Bot In Slack Went Smoothly,
Faced Issue In Setting Up Botkube In Minikube Cluster, The Bot Was Deployed Successully:
![image](https://user-images.githubusercontent.com/127375633/228222942-aa748b70-e6d6-4f4b-8645-d8b64efe8d56.png)
But When I Tried To Ping The Bot In Slack, It Didnt Worked.
Thanks To The Community Members For Solving The Issue

## Expected behavior
I Thought That There Might Be An Issue In Passing The Environment Variables & Might Be Issue With Helm

## Actual behavior

The Issue Is With The Helm Charts,
After Trying Various Solutions To Troubleshoot
Setting Up This `--set 'executors.helm.botkube/helm.enabled'=false` in Installation Procedure
Resolved My Issue.

Once Again Thanks To The Team For Solving The Issue.

pkosiec:
See the full thread here: https://botkube.slack.com/archives/C01CR1KS55K/p1679989389837669

The reason is that the Botkube Pod were killed as the Helm plugin download took too much time.

PR #1072, opened 2023-05-24
Calculate and check sha256 of plugins

josefkarasek:
## Description

- Introduce temporary dir structure `.downloading` for plugins while they're being downloaded
- Add sha256 checksum of Botkube plugins to Plugin Index
- Check checksum when downloading plugins

## Testing

0. Rebuild your plugins `make build-plugins-single` and host them locally `PLUGIN_SERVER_HOST=http://localhost go run test/helpers/plugin_server.go`

1. In another terminal check for plugin cache dir structure. While Botkube download plugins, you should see temp dir `.downloading`. This dir should disappear when plugin download succeeds.

```
watch -n 1 tree /tmp/plugins
```

2. Run botkube

```
rm -rf /tmp/plugins ; go run cmd/botkube/main.go
```

3. Mess up checksum of some plugin in plugin-index.yaml. Botkube now fails to start.

```
rm -rf /tmp/plugins ; go run cmd/botkube/main.go
```

## Related issue(s)

https://github.com/kubeshop/botkube/issues/1027


test/fake/plugin_server.go
```diff
@@ -35,14 +35,11 @@ func NewPluginServer(cfg PluginConfig) (string, func() error) {
 
 	basePath := fmt.Sprintf("%s:%d", cfg.Server.Host, cfg.Server.Port)
 	builder := plugin.NewIndexBuilder(loggerx.NewNoop())
-
+	idx, err := builder.Build(cfg.BinariesDirectory, basePath+"/static", ".*")
```
mszostok, line 38: I put that inside the handler as I wanted to rebuild it each time botkube calls the endpoint. I'm using:
```
env PLUGIN_SERVER_HOST=http://host.k3d.internal go run test/helpers/plugin_server.go
```

as a long living process. So I don't need to restart that if I rebuild the binary, as it will fail on checksum. Can we revert that?

mszostok (approved):
Tested e2e works well 👍 I have only one comment related to the fake plugin server 🙂

PR #1077, opened 2023-05-30
custom sa token path

huseyinbabal:
## Description

- Default token path for botkube deployment is changed. Now we use uuid in the token path

## Testing

- `gh ckec
- Create an instance in Botkube Cloud
- Run following

### DIFF
--- a/helm/botkube/templates/deployment.yaml
+++ b/helm/botkube/templates/deployment.yaml
@@ -1,3 +1,5 @@
+{{- $uuid := include "botkube.uuid" . -}}
+
 apiVersion: apps/v1
 kind: Deployment
 metadata:
@@ -74,6 +76,9 @@ spec:
           volumeMounts:
             - name: cache-volume
               mountPath: /tmp
+            - name: {{ $uuid }}
+              mountPath: /var/run/{{ $uuid }}/secrets/kubernetes.io/serviceaccount
+              readOnly: true
           {{- if not (include "botkube.remoteConfigEnabled" $) }}
             - name: config-volume
               mountPath: "/config"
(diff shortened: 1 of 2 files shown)


helm/botkube/templates/deployment.yaml
```diff
@@ -1,3 +1,5 @@
+{{- $uuid := include "botkube.uuid" . -}}
+
 apiVersion: apps/v1
 kind: Deployment
```
pkosiec, line 4: Unfortunately it won't work for now - we're missing `automountServiceAccountToken: false` property on Pod: https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/

Also, we'll need to adjust the code in `main.go` around

```go
        // Prepare K8s clients and mapper
	kubeConfig, err := clientcmd.BuildConfigFromFlags("", conf.Settings.Kubeconfig)
```
As underneath it uses `InClusterConfig()` which has fixed paths to files. So probably we need to pass them via the env:
image

pkosiec (commented):
I like the path randomizaton approach 👍 However, it won't work correctly for now as we still have the automounted token from SA. Please apply the changes and let me know about it 🙂

helm/botkube/templates/deployment.yaml
huseyinbabal: I only verified it on resource level as below, I didn't know there is such an effect, on it

Screenshot 2023-05-30 at 14 07 14
Screenshot 2023-05-30 at 14 07 05

pkosiec: Yep, on the resource level is totally fine, but we need to ensure this mounted kubeconfig is used by Botkube Core, and that the automounted token is no longer there so that no plugin will fall back to it 👍 Thanks!

huseyinbabal: @pkosiec I handled the custom sa token path + root ca

pkosiec (approved):
LGTM 🚀

pkosiec (approved):
LGTM 🚀

PR #1091, opened 2023-06-14
Send the cluster name for all interactive commands

mszostok:
## Description

-Send the cluster name for all interactive commands

The current implementation is quite hacky. The main problem is that only buttons are easily to handle as we know the whole command up front and we can simply add `--cluster-name` flag.

For all others, like selects, inputs, etc. the command is constructed either via https://github.com/kubeshop/botkube/blob/428e62405e9002a0b3b02ad9697e254521cbbb7c/pkg/bot/socketslack.go#L517-L544

or by using state https://github.com/kubeshop/botkube/blob/428e62405e9002a0b3b02ad9697e254521cbbb7c/internal/executor/kubectl/builder/kubectl.go#L393. 

This introduces additional complexity that we need to handle when we are injecting the `--cluster-name` flag as otherwise we will break the parsing on the plugin side.

### Notes

I tried to use a new field, use ids, and slack message metadata, none of them worked for me 😞 So I ended up with this kinda ugly solution.

### Future

Because of the current architecture, I cannot normalize it in a more straightforward way. IMO we should decouple slack commands and slack state. I would see sth like:

- clean cmd that is always passed to the plugin
- optional `metadata` which is `map[string]string` that holds the data that we extract from the Slack state. This way our plugins won't parse the initial slack state and as a result we can hold more details in the state that will be only handled in the core code.

## Testing

- Install and play with the interactive kubectl or help message.

### DIFF
--- a/cmd/botkube/main.go
+++ b/cmd/botkube/main.go
@@ -239,7 +239,7 @@ func run(ctx context.Context) error {
 		}
 
 		if commGroupCfg.CloudSlack.Enabled {
-			sb, err := bot.NewCloudSlack(commGroupLogger.WithField(botLogFieldKey, "CloudSlack"), commGroupName, commGroupCfg.CloudSlack, executorFactory, reporter)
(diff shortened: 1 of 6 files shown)


josefkarasek (approved):
lgtm :+1:

PR #1096, opened 2023-06-16
Fix ServiceAccount token path in Config Watcher

huseyinbabal:
## Description

- Config watcher docker image is updated and now helm chart uses up to date version
- It contains fix for config watcher service token problem


pkosiec (approved):
Works flawlessly 👌 Thanks for spotting and fixing this issue!

PR #1118, opened 2023-07-03
Streamline the help message

### LINKED ISSUE #1046: Make the Help message more streamlined
This is a proposal for a more streamlined Help message design.
**Reason:** we already have a lot going on and as we keep adding more and more features to document this message is only going to grow.

Notes:
* I think there is no need for `show config` to be part of this message, people can refer to the docs if they want to edit the full config file
* Let's refer to docs for features/plugins instead of trying to cram commands and help about each thing.
* We can either link to our web docs or send an additional help message when the user presses, for example, "kubectl help"
* Leave notification options and the ping as they are because they are good to have at people's disposal initially and they serve as examples that people can actually type commands @ the bot.

image

pkosiec:
## Description

- Streamline the help message

I did it _almost_ like Lacho suggested:
- Added "multi-cluster" section for Cloud Slack as these flags are really hidden
- description + 2 buttons in the same line is not possible by Slack API
Compare it with the task: #1046

## Testing

You can configure Botkube with some comm platforms and see the results.  You can modify `sendHelp` function to always show help:

```go
 
 // sendHelp sends the help message to all interactive bots.
 func sendHelp(ctx context.Context, s *storage.Help, clusterName string, executors []string, notifiers map[string]bot.Bot) error {
-	alreadySentHelp, err := s.GetSentHelpDetails(ctx)
+	_, err := s.GetSentHelpDetails(ctx) // TODO: Revert
 	if err != nil {
 		return fmt.Errorf("while getting the help data: %w", err)
 	}
 
 	var sent []string
 
 	for key, notifier := range notifiers {
-		if alreadySentHelp[key] {
-			continue
-		}
+		// TODO: revert
+		//if alreadySentHelp[key] {
+		//	continue
+		//}

```

But you can also check the screenshots below:

### Socket Slack

image

### Cloud Slack

image

### Mattermost

image

### Discord

image

## Related issue(s)

Resolves #1046


josefkarasek (approved):
Very nice! :+1:

PR #1122, opened 2023-07-05
Fix event update logic for Kubernetes source

### LINKED ISSUE #1081: Update events every 30 minutes
## Description

Using `Botkube v1.0.1` inside `Kubernetes v1.26.5` (Digitalocean managed).
Using `socketSlack`.
Using unmodified sources `k8s-all-events` and `k8s-recommendation-events`.
Using unmodified `k8s-default-tools` with only `botkube/kubectl` enabled.

I observe a strange behavior.. after the start of Botkube, **every 30 minutes** I get Slack notifications about all the **daemonsets, statefulsets, deployments** and **jobs** inside the cluster as if they are all being updated. Those notifications never contain any diffs. If I do `kubectl get event -A` right after I get those Slack notifications, there is nothing there.. when I do regular deploys of my stuff I can see events using kubectl and I get Slack notifications via Botkube.

## Expected behavior

I would like to get notifications only when something is actually happening.

## Actual behavior

Getting

:bulb: apps/v1/statefulsets updated
:bulb: apps/v1/daemonsets updated
:bulb: batch/v1/jobs updated
:bulb: apps/v1/deployments updated

for all resources of these types inside the cluster every 30 minutes when nothing is actually being updated... at least not that I know of or can see via `kubectl get event -A`.

## Steps to reproduce

Not really sure, but it's happening on both of my clusters.

##
Any idea how tho troubleshoot this would be great, for I'm kind of stuck with `k8s-err-events` for now.

pkosiec:
## Description

- Fix event update logic for Kubernetes source
   - when `updateSetting.fields` are provided, ensure there's a diff between these fields.
   - when there's no `updateSetting.fields` config, route all update events.
   - if `includeDiff` is true, print diff as a part of the message.
- Fix diff:
    - ensure it doesn't stop on first error
    - ensure we diff between empty property values (as something might be added/removed)
- ~Revert breaking changes in messages API~ extracted to #1124 

This bug was related to changes in our event dispatching system when extracting Kubernetes events to a separate plugin. An event was qualified to dispatch even if it didn't contain changes for the `updateSetting.Fields`.

This regression was introduced in 1.0.0.

## Testing

Run Botkube from this PR with the following config:

```yaml
communications:
  "default-group":
    socketSlack:
      enabled: true
      appToken: "xapp-"
      botToken: "xoxb-"
      channels:
        "default":
          name: botkube-demo
          bindings:
            executors:
              - k8s-default-tools
            sources:
              - k8s-all-events
plugins:
  cacheDir: /tmp
  repositories:
    botkube:
      url: http://localhost:3000/botkube.yaml
```

```
go run test/helpers/plugin_server.go
make build-plugins-single
export KUBECONFIG="..."
export BOTKUBE_SETTINGS_LOG_LEVEL=debug                     
export BOTKUBE_SETTINGS_KUBECONFIG=$KUBECONFIG
export BOTKUBE_CONFIG_PATHS="$(pwd)/helm/botkube/values.yaml,$(pwd)/comm_config.yaml"
export LOG_LEVEL_SOURCE_BOTKUBE_KUBERNETES="debug"
go run ./cmd/botkube/main.go
```

Apply sample resource, e.g. STS:
`k apply sts.yaml`

```yaml
apiVersion: apps/v1
kind: StatefulSet
metadata:
  name: web
spec:
  selector:
    matchLabels:
      app: nginx # has to match .spec.template.metadata.labels
  serviceName: "nginx"
  replicas: 3 # by default is 1
  minReadySeconds: 10 # by default is 0
  template:
    metadata:
      labels:
        app: nginx # has to match .spec.selector.matchLabels
    spec:
      terminationGracePeriodSeconds: 10
      containers:
      - name: nginx
        image: registry.k8s.io/nginx-slim:0.8
        ports:
        - containerPort: 80
          name: web
        volumeMounts:
        - name: www
          mountPath: /usr/share/nginx/html
  volumeClaimTemplates:
  - metadata:
      name: www
    spec:
      accessModes: [ "ReadWriteOnce" ]
      storageClassName: "my-storage-class"
      resources:
        requests:
          storage: 1Gi
```

Edit the resource:

`k edit sts web`

The following fields will trigger an update event:

```
                - status.readyReplicas
```

So you can try multiple different edits to see if they trigger the updates.

For previous Botkube, any update of such STS would trigger an event.

## Related issue(s)

Resolves #1081

### DIFF
--- a/helm/botkube/e2e-test-values.yaml
+++ b/helm/botkube/e2e-test-values.yaml
@@ -121,7 +121,7 @@ sources:
             updateSetting:
               includeDiff: false
               fields:
-                - spec.data
+                - data
   'k8s-annotated-cm-delete':
     displayName: "K8s ConfigMap delete events"
     'botkube/kubernetes':
@@ -199,7 +199,7 @@ sources:
             updateSetting:
               includeDiff: false
               fields:
-                - spec.data
+                - data
 
   'other-plugins':
     displayName: "K8s ConfigMaps changes"
--- a/helm/botkube/e2e-test-values.yaml
+++ b/helm/botkube/e2e-test-values.yaml
@@ -118,6 +118,10 @@ sources:
             - update
         resources:
           - type: v1/configmaps
+            updateSetting:
+              includeDiff: false
+              fields:
+                - spec.data
   'k8s-annotated-cm-delete':
     displayName: "K8s ConfigMap delete events"
     'botkube/kubernetes':
--- a/helm/botkube/e2e-test-values.yaml
+++ b/helm/botkube/e2e-test-values.yaml
@@ -192,6 +192,10 @@ sources:
             event: # overrides top level `event` entry
               types:
                 - update
+            updateSetting:
+              includeDiff: false
+              fields:
+                - spec.data
 
   'other-plugins':
     displayName: "K8s ConfigMaps changes"
--- a/internal/executor/kubectl/builder/kubectl_dropdowns.go
+++ b/internal/executor/kubectl/builder/kubectl_dropdowns.go
@@ -103,7 +103,7 @@ func PreviewSection(cmd string, input api.LabelInput) []api.Section {
 		},
 		{
 			Buttons: api.Buttons{
-				btn.ForCommandWithoutDesc(interactive.RunCommandName, cmd, api.ButtonStylePrimary),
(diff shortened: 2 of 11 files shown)


huseyinbabal (approved):
Works as expected 👍

Issue #1125, opened 2023-07-06
Finalize Botkube CLI migration tool


pkosiec:
## Acceptance Criteria

- Address comments from: https://github.com/kubeshop/botkube/pull/1119
- [ ] Implement UI pages on Botkube Cloud
    - Error state (https://www.figma.com/file/QNcHtrYHsRdjOhTfClMmmE/Botkube-SaaS?type=design&node-id=409-17173&t=jMNOZjtGpWp2u0Tc-0)?
- [ ] Consider using GCS to publish CLI?
- [ ] Make sure this job is ran on Cloud main (dispatch event to trigger the job)
   - Add some assertions
   - Introduce Slack notifications 
- [ ] Update docs -> @mszostok will do this

mszostok:
I know that we also discussed nesting cloud-related commands under `cloud`. So instead of `botkube login`, we should have `botkube cloud login`. The same applies to `migrate`. What do you think?

This way, we can clearly state in the documentation that if you want to work with the cloud, you always need to use `botkube cloud login`.

However, I'm not sure about the installation process. We could simply use `botkube install`, and if someone specifies, for example, an `apiKey`, we will fail if other flags, such as a values file, are provided. Another option is to have:
- `botkube install` to support installing the OS Helm chart while allowing changes to all settings.
- `botkube cloud install` as a simplified version of `botkube install`, where only a few dedicated flags are exposed, as the rest will be taken from the cloud.
  - Since we will be logged in, we could also enhance the installation process with data from the cloud (e.g., validate that the deployment status was changed to `connected`, check that you are installing a deployment that belongs to your account/org, etc.).

PR #1128, opened 2023-07-07
Improve migration command in Botkube CLI

pkosiec:
## Description

- Improve migration command in Botkube CLI

## Related issue(s)

#1125


cmd/cli/cmd/migrate.go
```diff
@@ -96,28 +116,47 @@ func NewMigrate() *cobra.Command {
...
 	flags.StringVar(&opts.Token, "token", "", "Botkube Cloud authentication token")
 	flags.StringVar(&opts.InstanceName, "instance-name", "", "Botkube Cloud Instance name that will be created")
-	flags.StringVar(&opts.CloudAPIURL, "cloud-api-url", "https://api.botkube.io", "Botkube Cloud API URL")
+	flags.StringVar(&opts.CloudAPIURL, "cloud-api-url", "https://api.botkube.io/graphql", "Botkube Cloud API URL")
 	flags.StringVar(&opts.CloudDashboardURL, "cloud-dashboard-url", "https://app.botkube.io", "Botkube Cloud URL")
 	flags.StringVarP(&opts.Label, "label", "l", "app=botkube", "Label of Botkube pod")
 	flags.StringVarP(&opts.Namespace, "namespace", "n", "botkube", "Namespace of Botkube pod")
 	flags.BoolVarP(&opts.SkipConnect, "skip-connect", "q", false, "Skips connecting to Botkube Cloud after migration")
+	flags.BoolVar(&opts.SkipOpenBrowser, "skip-open-browser", false, "Skips opening web browser after migration")
+	flags.BoolVar(&opts.AutoUpgrade, "auto-upgrade", false, "Automatically upgrades Botkube instance without additional prompt")
```
mszostok, line 146: **suggestion**

WDYT about having sth more generic like `--no-input`/`--no-interactivty`/`--no-confirm`?
So basically a user will have one flag to use when they want to disable all prompts, e.g. on CI.

https://clig.dev/#interactivity

cmd/cli/cmd/migrate.go
```diff
@@ -96,28 +116,47 @@ func NewMigrate() *cobra.Command {
...
+	flags.StringVar(&opts.ConfigExporter.Registry, "cfg-exporter-image-registry", "ghcr.io", "Config Exporter job image registry")
+	flags.StringVar(&opts.ConfigExporter.Repository, "cfg-exporter-image-repo", "kubeshop/botkube-config-exporter", "Config Exporter job image repository")
+	flags.StringVar(&opts.ConfigExporter.Tag, "cfg-exporter-image-tag", DefaultImageTag, "Config Exporter job image tag")
+	flags.DurationVar(&opts.ConfigExporter.PollPeriod, "cfg-exporter-poll-period", 1*time.Second, "Config Exporter job poll period")
+	flags.DurationVar(&opts.ConfigExporter.Timeout, "cfg-exporter-timeout", 1*time.Minute, "Config Exporter job timeout")
+	flags.BoolVarP(&opts.Debug, "debug", "d", false, "Turn on debug logging")
```
mszostok, line 152: On my PR I introduce a global `verbose` flag, so it will be easy to check across the whole code-base whether we should print more info. It should also make it consistent in case we reuse some components between commands, as the one global value will be used.

Additionally, I introduced an option to specify verbose level, the same what we had in the previous project. It can be useful when we run some GraphQL calls, and with `trace` we can print them too, but not always when `-v` is specified as it could be too verbose.

josefkarasek (approved):
lgtm :+1:

.github/workflows/branch-build.yml
```diff
@@ -4,13 +4,14 @@ on:
   push:
     branches:
       - main
+      - migration-improvements
```
pkosiec, line 7:
```suggestion
```

Issue #1133, opened 2023-07-09
Add `install` command to Botkube CLI


mszostok:
## Overview

A user needs to know whether the Botkube installation process is successful or not, and be informed of any errors. The current helm install process will time out if there are problems and the user needs to look at the pod logs to troubleshoot the issue. There are some common pitfalls with Botkube configuration that cause the installation to fail (e.g. bot not invited to channel, wrong channel name, etc.) and resolving those issues is cumbersome.

## Acceptance Criteria

- Add `install` command
      - use Helm Go SDK underneath to install the chart
      - Use Kubernetes API and stream Pod logs
          - print them to user only errors/warnings/info
          - ensure pretty print (logs are properly colored)
      - make sure it can be run with debug/verbose
      - support dry-run
      - support Botkube version and overrides
      - Respect KUBECONFIG env variable
- Document this command on docs.botkube.io

## New scope added

- Add `uninstall` command: https://github.com/kubeshop/botkube/pull/1141

PR #1134, opened 2023-07-09
Add option to install Botkube via CLI, add `--verbose` flag

mszostok:
## Description

- Add option to install Botkube via CLI
      - use Helm Go SDK underneath to install the chart
      - make sure it can be run with debug/verbose
      - support dry-run
      - support Botkube version and overrides
      - Respect KUBECONFIG env variable
- Add `--verbose` flag
  - it supports also modes that can be used to print more details. For example, for GraphQL call, we can print requests/responses when `--verbose=trace` is used. It should be useful to print it only if needed, as printing that all the time could be too messy.

## Testing

1. Build the CLI
2. Create k8s cluster
3. Run:

```bash
export SLACK_CHANNEL_NAME=random
export SLACK_APP_TOKEN="xapp-1-"

botkube install --verbose \
--set communications.default-group.socketSlack.enabled=true \
--set communications.default-group.socketSlack.channels.default.name=${SLACK_CHANNEL_NAME} \
--set communications.default-group.socketSlack.appToken=${SLACK_API_APP_TOKEN} \
--set communications.default-group.socketSlack.botToken=${SLACK_API_BOT_TOKEN} \
--set settings.clusterName=labs \
--set 'executors.k8s-default-tools.botkube/kubectl.enabled'=true \
--set 'executors.k8s-default-tools.botkube/helm.enabled'=true
```

here are described all `install` flags: https://github.com/kubeshop/botkube/blob/db6a62fcce7dee3856827fa05a32971e837c400f/cmd/cli/docs/botkube_install.md

https://github.com/kubeshop/botkube/assets/17568639/c61ff2f3-3d6f-4a8f-9ac4-23be107c9eca

## Notes

**In the next PR I will add streaming Botkube Pod logs. This PR is already huge and I'm still working on logs.**

## Related issue(s)

- https://github.com/kubeshop/botkube/issues/1133


internal/kubex/config.go
```diff
@@ -0,0 +1,89 @@
...
+
+	// 3. In-cluster config if running in cluster
+	// 4. $HOME/.kube/config if exists
+	// 5. user.HomeDir/.kube/config if exists
+	//
+	// NOTE: For default config file locations, upstream only checks
+	// $HOME for the user's home directory, but we can also try
+	// os/user.HomeDir when $HOME is unset.
+	//
+	// TODO(jlanford): could this be done upstream?
```
huseyinbabal, line 63: `jlanford`? :)

internal/cli/install/install.go
```diff
@@ -0,0 +1,153 @@
...
+	renderer := style.NewGoTemplateRender(style.DefaultConfig(successInstallGoTpl))
+
+	props := map[string]string{
+		"DocsURL": "https://docs.botkube.io/cli/getting-started/#first-use",
+		"Version": version,
+	}
+
+	out, err := renderer.Render(props, cli.IsSmartTerminal(w))
+	if err != nil {
+		return err
```
huseyinbabal, line 93: we can add err context

huseyinbabal (approved):
Lot's of improvements, thank you 🚀

internal/kubex/config.go
mszostok: as mentioned in the function comment, I copied that from controller-runtime and I wanted to keep is as close as possible to original. But I can remove that comment 👍

PR #1148, opened 2023-07-14
Add timestamp annotation on update

mszostok:
## Description

- Add timestamp annotation on update


pkosiec (approved):
With an upgrade which nothing changes works flawlessly 👍 But I noticed sometimes we don't show the logs when we do actual upgrade (which recreates the pod during actual upgrade). See the video for details - looks like a race condition issue.


https://github.com/kubeshop/botkube/assets/7155799/639aae23-0f50-4844-82b8-3df1d32c2b3b


I don't think this is something very urgent,  but maybe easy to fix? if not, we can do it in a follow-up.

PR #1153, opened 2023-07-18
chore: pkg imported more than once

testwill:
## Description
package "github.com/charmbracelet/log" is being imported more than once 

- ...


pkosiec (approved):
Good catch, thanks!

PR #1154, opened 2023-07-18
add human friendly error messages

huseyinbabal:
## Description

- Mapped error messages in bot packages
- There is no need to add error mapping for mattermost since they already send readable error descriptions, we use them.

Here are the screenshots for testing results for different scenarios for Slack, Discord, and Mattermost
### Slack
Screenshot 2023-07-21 at 13 36 39

### Discord
Screenshot 2023-07-21 at 13 29 34

## Testing
You can create an instance on Botkube Cloud Dev, and provide wrong creds, wrong channel, or not invited channel for each platform. You need to see human-friendly messages as shown above. 

You need to run botkube deployment via `go run ...` on this branch after exporting 
```bash
export CONFIG_PROVIDER_API_KEY=...
export CONFIG_PROVIDER_ENDPOINT=https://api-dev.botkube.io/graphql
export CONFIG_PROVIDER_IDENTIFIER=...
```

## Related issue(s)
https://github.com/kubeshop/botkube-cloud/issues/520

### DIFF
--- a/cmd/botkube-agent/main.go
+++ b/cmd/botkube-agent/main.go
@@ -5,7 +5,6 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
-	"strings"
 	"time"
 
 	"github.com/google/go-github/v44/github"
@@ -20,7 +19,7 @@ import (
 	"k8s.io/client-go/discovery/cached/memory"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
-	k8sStrings "k8s.io/utils/strings"
+	"k8s.io/utils/strings"
 	"sigs.k8s.io/controller-runtime/pkg/manager/signals"
 
 	"github.com/kubeshop/botkube/internal/analytics"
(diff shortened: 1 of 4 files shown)


pkosiec (commented):
I see some problems with the approach:
- we're losing context - we don't know which channel is invalid, or which channel the bot should join. These errors are user friendly, but unfortunately not helpful.
- it will be hard to maintain, as some of the errors are wrapped even 2 times.
- we still log less user friendly messages.

While the idea of error mapping of course totally makes sense, I'd suggest doing it in different way:
- map the errors in corresponding Bots (like bot/socketslack.go,  bot/discord.go)
    - avoid string comparison if possible: if discord/mattermost/slack packages export some errors, use them with `errors.Is(...)`. Of course if that's not possible, then string comparison is our last resort.
    - add context like channel name and communication group name.
- these errors will be still logged, wrapped and reported as they were in main. So user checking logs will be able to understand the issue, with all the details.

What do you think? Thanks!

huseyinbabal:
> 

Hi @pkosiec here is what I did;

### Slack
They don't have error model for our cases, so I couldn't use them, used error string instead

### Discord
I have used their  error model and re-mapped error with additional context

### Mattermost
We don't need to do anything for mattermost since they already have error description, and they are human-readable.

pkosiec (approved):
👌 Looks good to me! Didn't test it but based on the screenshots I see that it works great.

Issue #1126, opened 2023-07-06
Add acknowledge indicator for processing Slack messages


mszostok:
## Overview

With the current release, we will introduce some commands that may take longer (e.g. flux install/chat gpt reponse/etc).
We should user know that we received the request and we are processing it.

## Acceptance Criteria

- [x] For each message that Botkube starts processing, add “:eyes:” as “ack” and replace with “:heavy_check_mark: ” once the message is sent
- [x] Cover only Socket and Cloud slack platforms
- [x] Update documentation for Socket Slack about new scope (reactions:write)
- [x] Update Cloud Slack app and add `reactions:write`
- [x] **[UX]** Add a note to the UI for already existing cloud slack deployments to “re-install” Bot in their workspace to get the token refreshed with a new scope.

- [x] Mattermost/Discord/Teams are out of scope - double check @brampling 

## Reason

We don’t have an option to stream executor response yet, so the user doesn’t know whether the command is processed or not.

## UX Design

https://www.figma.com/file/QNcHtrYHsRdjOhTfClMmmE/Botkube-SaaS?type=design&node-id=2278-27459&mode=design&t=mhZm0mx5aaSsPdVk-4

## Demo

https://github.com/kubeshop/botkube/assets/17568639/0992c7bf-30cf-4cb7-9403-8fc43aee4207

mszostok:
I spent ~0.5MD more to establish designs and implement them in a new way. We need to pay attention to have designs before estimations :)

PR #1156, opened 2023-07-19
Add reactions to Slack messages

mszostok:
## Description

- [x] Add support both for Socket and Cloud Slack platforms
- [x] Add reactions both for typed messages add interactive events (button clicks)
  - [x] works in thread
- [x] we don't do that for messages  that are visible only to a user. Basically, Slack doesn't support that. However, this is OK for us as otherwise it will look wired e.g. for the `kubectl` command builder

## Testing

Install Socket Slack and run:
-  `@Botkube kubectl get pods -A`
- `@Botkube kubectl`
- `@Botkube help` and click some buttons.

## Related issue(s)

- https://github.com/kubeshop/botkube/issues/1126


huseyinbabal:
@mszostok this works on socket slack, but couldn't make it work for cloud slack. I added `reactions:write` to `BotkubeDev` user already and reinstalled botkube playground channel. Do I need to reconnect to cloud workspace?

pkg/bot/slack_msg_tracker.go
```diff
@@ -0,0 +1,80 @@
...
+import (
+	"fmt"
+
+	"github.com/sirupsen/logrus"
+	"github.com/slack-go/slack"
+)
+
+const (
+	msgReceivedEmoji  = "eyes"
+	msgProcessedEmoji = "heavy_check_mark"
```
huseyinbabal, line 12: I would use following to show completed process
Screenshot 2023-07-20 at 10 59 35
If you are using dark mode, marked icon is almost invisible as follows
Screenshot 2023-07-20 at 11 00 36

pkg/bot/slack_socket.go
```diff
@@ -216,6 +205,7 @@ func (b *SocketSlack) Start(ctx context.Context) error {
 						UserName:        userName,
 						CommandOrigin:   cmdOrigin,
 						State:           state,
+						EventTimeStamp:  callback.Message.Timestamp, // FIXME
```
huseyinbabal, line 208: do we need to fix this in this PR?

mszostok: the comment is  already outdated, I will remove it 👍

pkg/bot/slack_msg_tracker.go
mszostok: changed 👍 I also double-checked with Donna  and she also prefers "✅ "

huseyinbabal (approved):

PR #1161, opened 2023-07-21
fix doctor plugin schema

madebyrogal:
## Description

- change doctor plugin schema 04 to 07
- change plugin server port 3000 -> 3010 cos 3000 is usually used by react and conflicts with local UI (this is a proposal)

### DIFF
--- a/test/helpers/plugin_server.go
+++ b/test/helpers/plugin_server.go
@@ -4,6 +4,7 @@ import (
 	"log"
 	"os"
 	"path/filepath"
+	"strconv"
 
 	"github.com/kubeshop/botkube/test/fake"
 )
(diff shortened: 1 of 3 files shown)


test/helpers/plugin_server.go
```diff
@@ -22,7 +22,7 @@ func main() {
 		BinariesDirectory: binDir,
 		Server: fake.PluginServer{
 			Host: host,
-			Port: 3000,
+			Port: 3010,
```
mszostok, line 25: you can also make it configurable in the same why as we have for `PLUGIN_SERVER_HOST` 👍

mszostok (approved):
Please update helm doc: https://github.com/kubeshop/botkube/blame/main/cmd/executor/helm/README.md#L48 with the port change 👍 Thanks!

PR #1164, opened 2023-07-25
Handle server closing grpc stream

josefkarasek:
## Description

- ...


pkg/api/cloudslack/error.go
```diff
@@ -0,0 +1,35 @@
+package cloudslack
+
+type CloudSlackErrorType string
+
+const (
+	CloudSlackErrorBadRequest CloudSlackErrorType = "BadRequest"
```
pkosiec, line 6: Can we rename this error? Bad request sounds a bit too generic to assume this is always "quota exceeded" case. So WDYT about renaming this to `QuotaExceeded`, both on client and server?

pkosiec (commented):
Works well! 👍 Overall LGTM with a single comment here.

pkosiec (approved):
LGTM, but please wait for e2e testing before merge 👍

PR #1186, opened 2023-08-11
Add flux under default executors

mszostok:
## Description

- Add flux under default executors


josefkarasek (approved):
lgtm

Issue #1172, opened 2023-08-02
Improve `botkube migrate` flow


mszostok:
## Overview

The `botkube migrate` flow uses an outdated approach for installation and is missing some UX enhancements. 

<details open><summary>Issues</summary>
<p>

- **[bug]** When I don’t have the `k8s-default-tools` executor group bound or at least one executor enabled, I got:
    ```json
    {"level":"fatal","msg":"while running application: while merging app configuration: found critical validation errors: 2 errors occurred:\n\t* Key: 'Config.Actions[show-logs-on-error].Bindings.k8s-default-tools' 'k8s-default-tools' binding not defined in Config.Executors\n\t* Key: 'Config.Actions[describe-created-resource].Bindings.k8s-default-tools' 'k8s-default-tools' binding not defined in Config.Executors","time":"2023-08-02T14:31:01+02:00"}
    ```
      so I cannot migrate even though my OS setup is working fine. It’s probably because of our default automations that we added, they most probably require having that but during migration we create deployment without it.

    - Solution: filter out invalid bindings for Action same as we do for platforms: https://github.com/kubeshop/botkube-cloud/pull/534/files
- **[UX]** The `botkube migrate` uses the helm install command instead of `botkube install` one, so all features are missing (logs streaming, cluster info, etc.)

- **[bug]** In docs we do not specify that exported `KUBECONFIG` is required
    - **[UX]** IMO we should also match the botkube install experience and OOTB get current user K8s context
- **[UX]** We should show a cluster info against which the migration is running
- **[UX]** The Botkube CLI help should specify that `botkube install` is used for both installing and upgrading Botkube in a cluster.

</p>
</details> 

## Acceptance Criteria

- allow migration when  `k8s-default-tools` are disabled/not bound
- use `install` internal command to perform Botkube installation
- use `kubex.LoadRestConfigWithMetaInformation()` and register `--kubeconfig` flag, to get OOTB experience with finding current k8s context
- print cluster context name to tell the user against which cluster migration is executed
- For `v1.2.0` migration document, add a note that `KUBECONFIG` env variable is required (for new docs, it won't be required)
- migrator side (generate unique configurationName) 
   **Problem**
    When creating an instance, CLI migrate sends configurationName which is not unique. For example, when migrating k8s-default-tools plugin configuration group (that contains both kubectl and helm), we send configurations for kubectl and helm with the same configurationNames (k8s-default-tools).
[...]

PR #1198, opened 2023-08-17
Improve migrate flow

josefkarasek:
## Description

- allow migration when  `k8s-default-tools` are disabled/not bound
- use `install` internal command to perform Botkube installation
- use `kubex.LoadRestConfigWithMetaInformation()` and register `--kubeconfig` flag, to get OOTB experience with finding current k8s context
- print cluster context name to tell the user against which cluster migration is executed
- generate unique `configurationName`
- Botkube CLI help specifies that `botkube install` is used for both installation and upgrading an existing Botkube instance.

## Testing

Create Botkube instance in your local env
Login and migrate
```
helm upgrade botkube --install --version v1.2.0 --namespace botkube --create-namespace --wait   --set communications.default-group.socketSlack.enabled=true   --set communications.default-group.socketSlack.channels.default.name=${SLACK_CHANNEL_NAME}   --set communications.default-group.socketSlack.appToken=${SLACK_API_APP_TOKEN}   --set communications.default-group.socketSlack.botToken=${SLACK_API_BOT_TOKEN}   --set executors.k8s-default-tools.botkube/kubectl.enabled=true   --set settings.clusterName=${CLUSTER_NAME}   --set settings.log.level=debug   --set analytics.disable=true   botkube/botkube

go run cmd/cli/main.go login --cloud-dashboard-url=http://localhost:3000
go run cmd/cli/main.go migrate -w --cloud-api-url=http://localhost:8080/graphql --cloud-dashboard-url=http://localhost:3000
```

For testing that Actions can be migrated even when a bound plugin is not defined - rename a plugin in one of the pre-defined actions and migrate. The Action should be omitted and not included in the migration.

## Related issue(s)

https://github.com/kubeshop/botkube/issues/1172


internal/cli/install/helm/install.go
```diff
@@ -77,7 +76,7 @@ func (c *Helm) Install(ctx context.Context, status *printer.StatusPrinter, opts
 			}
 
 			if !upgrade {
-				return nil, errors.New("upgrade aborted")
+				return nil, nil
```
josefkarasek, line 79: User chooses not to run `helm upgrade` - this is not an error situation.

huseyinbabal:
@josefkarasek I saw only one glitch about flow, even we use `-w` flag in migrate command, it does  not wait for botkube become ready. For example it says, botkube upgraded successfully, but actually it is still in Terminating state, and becomes available afterwards, just FYI...

josefkarasek:
> @josefkarasek I saw only one glitch about flow, even we use `-w` flag in migrate command, it does not wait for botkube become ready. For example it says, botkube upgraded successfully, but actually it is still in Terminating state, and becomes available afterwards, just FYI...

I could not reproduce this. With Huseyin we agreed we'll monitor this in the immediate future.

PR #1201, opened 2023-08-19
Better human readable format for err

SubhajitDutta07:
## Description
i just added a better format for that error to be read by the users. 

- ...

### DIFF
--- a/pkg/notifier/bot.go
+++ b/pkg/notifier/bot.go
@@ -43,7 +43,7 @@ func SendPlaintextMessage(ctx context.Context, notifiers []Bot, msg string) erro
 			},
 		})
 		if err != nil {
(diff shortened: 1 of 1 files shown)


pkosiec:
Hi @SubhajitDutta07,
Thanks for opening the PR, but unfortunately we cannot accept this contribution. The error you modified uses the standard convention we use for wrapping Go errors across the whole codebase. This would bring inconsistency. Also, additional, unnecessary whitespace characters would make it worse when printing it on output. 

Before opening next PRs, please get familiar with our contribution rules: https://docs.botkube.io/community/contribute/#making-a-change

Creating an issue or discussing a future contribution on Slack will be much easier for us and will make it easier to get a PR merged. If you encountered a place where such error is printed in a bad way, then let us know about this in an issue.

Cheers!

PR #1202, opened 2023-08-21
Add support for archived plugins

mszostok:
## Description

- Add support for archived plugins

## Testing
Integration test are already using archives. However, you can test that also locally by running botkube with local plugins in the same way as you did before.

## Related issue(s)

- https://github.com/kubeshop/botkube-docs/issues/235

### DIFF
--- a/.github/workflows/pr-build.yaml
+++ b/.github/workflows/pr-build.yaml
@@ -237,6 +237,7 @@ jobs:
         env:
           # we hardcode plugins version, so it's predictable in e2e tests
           GORELEASER_CURRENT_TAG: "v0.0.0-latest"
+          OUTPUT_MODE: "archive"
         run: |
           make build-plugins
 
--- a/.github/workflows/pr-build.yaml
+++ b/.github/workflows/pr-build.yaml
@@ -252,3 +252,7 @@ jobs:
         run: |
           KUBECONFIG=$(k3d kubeconfig write ${{ matrix.integration }}-test-cluster) \
             make test-integration-${{ matrix.integration }}
+
+      - name: Dump cluster
+        if: ${{ failure() }}
+        uses: ./.github/actions/dump-cluster
--- a/hack/target/build-plugins/main.go
+++ b/hack/target/build-plugins/main.go
@@ -1,0 +1,73 @@
+package main
+
+import (
+	"flag"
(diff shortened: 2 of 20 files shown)


hack/target/build-plugins/main.go
```diff
@@ -0,0 +1,76 @@
...
+	"strings"
+
+	"github.com/mattn/go-shellwords"
+
+	"github.com/kubeshop/botkube/internal/loggerx"
+)
+
+func main() {
+	pluginTargets := flag.String("plugin-targets", getEnv("PLUGIN_TARGETS", ""), "Comma separated list of specific targets to build. If empty, all targets are built.")
+	outputMode := flag.String("output-mode", getEnv("OUTPUT_MODE", "archive"), "Output format. Allowed values: binary or archive.")
```
mszostok, line 18: in the next PR I can revert that and bring the default value to `binary`. However now the `OUTPUT_MODE` set in the workflow file is ignored so I need to do it in the code instead.

.github/workflows/pr-build.yaml
```diff
@@ -251,3 +252,7 @@ jobs:
         run: |
           KUBECONFIG=$(k3d kubeconfig write ${{ matrix.integration }}-test-cluster) \
             make test-integration-${{ matrix.integration }}
+
+      - name: Dump cluster
```
mszostok, line 256: I can add the same in other places, WDYT?

pkosiec (approved):
LGTM, but please see the comments before merge 👍

.github/workflows/pr-build.yaml
```diff
@@ -237,6 +237,7 @@ jobs:
         env:
           # we hardcode plugins version, so it's predictable in e2e tests
           GORELEASER_CURRENT_TAG: "v0.0.0-latest"
+          OUTPUT_MODE: "archive"
```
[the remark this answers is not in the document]
mszostok: my idea was to change the default as I wrote here https://github.com/kubeshop/botkube/pull/1202#discussion_r1301527837

however, even if we will not change the default it highlights that the archive is used, so I left it for now to make it more explicit, hope you don't mind 👍

pkosiec (approved):
👌

Issue #878, opened 2022-12-06
 Watch and respawn started plugins that exited


mszostok:
## Overview

Watch and respawn started plugin sources and executors that exited. Currently, for sources, the `Dispatcher` service only get a client and starts the streaming process.

https://github.com/kubeshop/botkube/blob/22da62e1a67a85e0fb2e4af9a6e9106871179d14/internal/source/dispatcher.go#L42-L62

However, each plugin is a separate binary that starts as a sub-process. As a result, if there is e.g. panic in a given plugin, it closes this sub-process, and we will not receive new events from this source until the Botkube process will be restarted. Also, executors should be covered too.

We should detect such situations and add a retry mechanism:
- It can be as simple as just starting the binary once again.
- We can also think about sth similar to what K8s does when the Pod is crashing. 

- related issue on go-plugin repo: https://github.com/hashicorp/go-plugin/issues/31. We can take a look on how other projects are handling such situation.

## Acceptance Criteria

- detect crashing plugin and restart them using simple retry mechanism
- document implemented approach 
- e2e test coverage
 - test executor: update echo plugin: add code to fail with a specific command
 - consider testing source as well: update config map watcher to fail on a given annotated cfg map, test that it worked after deleting it
- have a retry threshold and after X retries, we can give up with a given source/executor restarting and send a message to Slack that such source was deactivated because of constant crashing.
   - send message via Bot that a given plugin is crashing and related events may be lost
     - send the message only to the bind channels
   - make it configurable for user
     - default strategy: keep running Botkube anyway, but ofc send message
     - strategy: if one plugin is failing, exit Botkube if user prefers that

## Reason

Make the source plugin dispatcher more reliable to avoid the situation when our users stop receiving source events because of some random plugin crashes.

PR #1204, opened 2023-08-23
Restart crashed plugins

josefkarasek:
## Description

- Use built-in hashicorp `Ping()` API call to monitor health of source plugins
- Release resources of crashed plugins
- Restart crashed plugins
- Define configurable strategies to approach restart policies 

## Testing

Add source cm-watcher and executor echo to your comm platform.

Add to your config:
```
plugins:
    agentRestartPolicy:
      # -- Restart policy type. Allowed values: "restartAgent", "deactivatePlugin".
      type: "deactivatePlugin"
      # -- Number of restarts before policy takes into effect.
      threshold: 5
```

Recompile plugins and start botkube locally

```
make build-plugins-single gen-plugins-index
go run cmd/botkube-agent/main.go
```

Watch botkube pod logs.

### Executor testing

`@Botkube echo @panic` will cause the echo plugin to panic and exit. Wait a few seconds and it will be restarted.
Check again with `@Botkube echo hello`.

### Source testing

Create cm with annotation `die: "true"`.

```
kubectl apply -f - <<EOF
apiVersion: v1
kind: ConfigMap
metadata:
  annotations:
    die: "true"
  name: watcher
  namespace: botkube
EOF
```

When this cm exists, cm-watcher plugin will continue to crash. Remove the cm. The plugin should restart.
Create the cm without the annotation - plugin should send message to specified channel.

## Related issue(s)

https://github.com/kubeshop/botkube/issues/878

### DIFF
--- a/internal/plugin/health_monitor.go
+++ b/internal/plugin/health_monitor.go
@@ -74,7 +74,7 @@ func (m *HealthMonitor) monitorSourcePluginHealth(ctx context.Context) {
 			}
 
 			m.sourcesStore.EnabledPlugins.Insert(repoPluginPair, p)
-			m.schedulerChan <- fmt.Sprintf("%s/%s/%s", plugin.group, plugin.repo, plugin.name)
+			m.schedulerChan <- repoPluginPair
 		}
 	}
 }
--- a/internal/plugin/health_monitor.go
+++ b/internal/plugin/health_monitor.go
@@ -52,7 +52,7 @@ func (m *HealthMonitor) monitorSourcePluginHealth(ctx context.Context) {
 		case <-ctx.Done():
 			return
 		case plugin := <-m.sourceSupervisorChan:
-			m.log.Infof("Restarting source plugin %q...", plugin.name)
+			m.log.Infof("Restarting source plugin %q, attempt %d/%d...", plugin.name, m.pluginRestartStats[plugin.name]+1, m.policy.Threshold)
 			if source, ok := m.sourcesStore.EnabledPlugins.Get(plugin.name); ok && source.Cleanup != nil {
 				m.log.Debugf("Releasing resources of source plugin %q...", plugin.name)
 				source.Cleanup()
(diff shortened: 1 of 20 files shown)


internal/plugin/health_monitor.go
```diff
@@ -0,0 +1,132 @@
...
+			m.executorsStore.EnabledPlugins.Insert(repoPluginPair, p)
+		}
+	}
+}
+
+func (m *HealthMonitor) shouldRestartPlugin(plugin string) bool {
+	restarts := m.pluginRestartStats[plugin]
+	m.pluginRestartStats[plugin]++
+
+	switch m.policy.Type {
```
mszostok, line 120: maybe we can normalize it, to all small letters? so even if I type `restartAgent`  instead of `RestartAgent` it will work.

internal/plugin/health_monitor.go
```diff
@@ -0,0 +1,132 @@
...
+	case config.KeepAgentRunningWhenThresholdReached:
+		return restarts < m.policy.Threshold
+	case config.RestartAgentWhenThresholdReached:
+		if restarts >= m.policy.Threshold {
+			m.log.Fatalf("Plugin %q has been restarted %d times and selected agentRestartPolicy is %q. Exiting...", plugin, restarts, m.policy.Type)
```
mszostok, line 125:
```suggestion
			m.log.Fatalf("Plugin %q has been restarted %d times and selected restartPolicy is %q. Exiting...", plugin, restarts, m.policy.Type)
```

internal/plugin/health_monitor.go
```diff
@@ -0,0 +1,132 @@
...
+}
+
+func (m *HealthMonitor) monitorExecutorPluginHealth(ctx context.Context) {
+	m.log.Info("Starting executor plugin supervisor...")
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case plugin := <-m.executorSupervisorChan:
+			m.log.Infof("Restarting executor plugin %q...", plugin.name)
```
mszostok, line 89: would be nice to print the "status" with the number of retries and max retries like `(attempt no 2 of max 10) `

mszostok (approved):
Very impressive! 🚀  I like the implementation, and I left only minor comments.

I see such todos:
- add option to print plugin status or add it to `list executors/sources` as a new column
- add e2e tests cases
- update documentation

let me know if we should take over those items 👍 

P.S. in the PR desc you have `agentRestartPolicy` but it should be `restartPolicy` and also for the current impl types should start with upper case.

mszostok:
Code merged in #1236

PR #1214, opened 2023-08-25
Add fallback strategy to get MM team

mszostok:
## Description

- Add fallback strategy to get MM team

## Testing

When you're running Botkube against the MM docker image `mattermost/mattermost-preview` it does not work as they don't have the search team endpoint implemented. So this change allows using the GetTeam as a fallback strategy.

You can test it by following those instructions: https://github.com/mattermost/mattermost-bot-sample-golang/tree/master

and later doing the same but with a helm chart:
```
$ helm repo add mattermost https://helm.mattermost.com
$ helm install mattermost/mattermost-team-edition \
  --set mysql.mysqlUser=sampleUser \
  --set mysql.mysqlPassword=samplePassword \
```


pkg/bot/mattermost.go
```diff
@@ -528,6 +523,29 @@ func getBotUserID(ctx context.Context, client *model.Client4, teamID, botName st
...
+		botTeam, _, err := client.GetTeamByName(ctx, name, "")
+		if err != nil {
+			return nil, fmt.Errorf("while getting team by name: %w", err)
+		}
+		botTeams = append(botTeams, botTeam)
+	}
+	if len(botTeams) == 0 {
+		return nil, fmt.Errorf("team %q not found", name)
+	}
+
```
pkosiec, line 545: Should we handle a case when there are multiple `botTeams`? Maybe return an error that we found multiple (printing theirs name + displayName)?

pkosiec (approved):
LGTM
Disclaimer: I did just the code review.

Issue #1167, opened 2023-07-31
Filter out NodeSysCtlChange events by default


huseyinbabal:
Especially in GKE clusters, kubernetes plugin sends notification about following event. 
```
 v1/nodes error
Kind: Node
Name: gke-botkube-cloud-co-botkube-cloud-co-a821e340-dgd9
Reason: NodeSysctlChange
Cluster: botkube-cloud-prod
Messages
• {"unmanaged": {"fs.aio-nr": "128", "net.netfilter.nf_conntrack_buckets": "32768"}}
```
This is a false positive message and we can filter out this by default. There is also a [question in StackOverflow](https://stackoverflow.com/questions/64467772/gcp-gke-cluster-node-throws-unmanaged-fs-aio-nr-2272-net-ipv4-tcp-fa) about this

## AC:
- Ignore such error as a part of default `vaules.yaml` config

PR #1218, opened 2023-08-30
Ignore error-like GKE node maintenance events

huseyinbabal:
## Description

- Exluded error-like GKE node maintenance events.

## Related issue(s)

https://github.com/kubeshop/botkube/issues/1167


mszostok (approved):
LGTM, if that will not help, we can be even more aggressive and  ignore all `NodeSysctlChange`

PR #1226, opened 2023-09-01
Fix printing steps when file redirect is used

mszostok:
## Description

- Fix printing steps when file redirect is used

## Testing

`botkube config get > /tmp/config.yaml`


pkosiec (approved):
LGTM, works well 👌

PR #1220, opened 2023-08-30
Introduce interactive ArgoCD source plugin

pkosiec:
## Description

Changes proposed in this pull request:

- Introduce ArgoCD source plugin
   - Configure ArgoCD notifications on plugin startup
   - Receive interactive notifications from ArgoCD
   - Define triggers and templates based on ArgoCD notification catalog

## Questions

- Cleanup? For now it is manual. Should we introduce a new method for plugins and run it on SIGINT (context cancel)? I'm afraid it can be difficult to fit in time with modifying ConfigMaps and removing annotation for apps for all configurations 🤔 

## Follow-up PRs

- 1. Improvements 
    - [ ] Enhanced suport for other communication platforms
    - [x] Introduce JSON schemas for configuration and request (done in this PR)
- 2. Documentation

## Video

https://github.com/kubeshop/botkube/assets/7155799/cba259c5-39a8-4ba9-95f4-c1a3b69dc219

(sorry for a long delay when typing admin password - I was looking for it 😅)

BTW just two `kubectl` commands are supported:
image

## Testing

Checkout PR:
```
gh pr checkout 1220
```

Build all plugins:
```
rm -rf /tmp/botkube;  rm /tmp/botkube.yaml; rm -rf /tmp/plugins
PLUGIN_DOWNLOAD_URL_BASE_PATH="http://localhost:64538/plugin-dist" make gen-plugins-index
npx serve --listen 64538
```

Prepare `comm_config.yaml`:
```yaml
communications:
  "default-group":
    socketSlack:
      enabled: true
      appToken: "xapp-"
      botToken: "xoxb-"
      channels:
        "default":
          name: "botkube-demo"
          bindings:
            executors:
              - k8s-default-tools
            sources:
              - argocd

plugins:
  repositories:
    botkube:
       url: http://localhost:64538/plugins-index.yaml

  incomingWebhook:
    enabled: true
    inClusterBaseURL: http://host.k3d.internal:2115

sources:
  'argocd':
    botkube/argocd:
      enabled: true
      context:
        rbac:
          group:
            type: Static
            static:
              values: [*argocd-group]
      config:
        log:
          level: debug
        defaultSubscriptions:
          applications:
            - name: guestbook # your app name
              namespace: argocd

        argoCD:
          uiBaseUrl: http://localhost:8080 # if argo is not exported
```
For full options check out `values.yaml`.

Apply these RBAC resources:

<details><summary>Details</summary>
<p>

```yaml
---
[...]

Issue #1175, opened 2023-08-07
Enhanced monitoring of Argo deployments with Botkube and GitHub integration


mszostok:
## Overview

The goal is to integrate with Argo CD. Besides forwarding the original Argo CD notifications, we want to enhance monitoring of Argo deployments by integrating with Botkube and possibly GitHub. 

## Phases

```[tasklist]
- [ ] #1238
- [ ] #1239
```

## Acceptance Criteria

- Create Argo source plugin
   -  Allow to forward standard Argo CD events that are available OOTB by Argo itself.
- Install Argo plugin on our dev cluster, ping dev team to play around with it
- Get feedback from Kubefirst and Testkube team
- Implement enhancements after dog-fooding

## Reason

- Users can manage platform secrets and notification in a single place instead of doing that per project, e.g., Kubernetes, Prometheus, Argo, etc.
- User has a more straight forward way to enable Argo notification
- User gain more visibility into Argo deployment process thanks to combining GitHub and Argo events and getting them directly on configured communication platform

mszostok:
Two stages:

1. Create Argo CD Botkube plugin:
    - [ ] how to integrate with Argo:
         - webhooks? 
         - custom trigger?
         - custom polling for Argo events
    - [ ] simplify subscribing for events, instead of require creating configmaps, secrets, templates, we should sth like:
      ```yaml
       on:
         - type: "app-synced":
             enabled: true
             apps: ['foo', 'bar']
             namespace:
               include: ["foo"]
      ```
    - [ ] create documentation

2. Dog-fooding:
   - use Botkube Argo CD plugin for our deployments to see more insights about the status on our configured channels
   - scenario:
      - we update the App version
      - we get notification that app is being deployed
      - if case of error we are notified about it also. For example, 
        - `Error: secret "botkube-api-key" not found`


3. Interactivity, fine-tune configuration:


**Prerequisite: Once implemented, play inside the team and see which actions and configuration options (e.g. ignore some spammy events) would be valuable to add for given events.**

based on our dog fooding;
- add some pre-defined buttons that are useful. For example:
      - `logs`
      - `describe app`
      - `force re-sync`
- useful filters to ignore spammy/not relevant events
   Notes:
   - https://argoproj.github.io/argo-events/sensors/filters/intro/ 
   - https://github.com/argoproj/argo-events/tree/master/examples/sensors

- work on the message format to make it more user-friendly

pkosiec:
[...]

PR #1228, opened 2023-09-06
Improve non-interactive messages for ArgoCD source

pkosiec:
## Description

- Improve non-interactive messages for ArgoCD source
- Simplify config in `values.yaml`
- Fix emoji rendering for Teams in cards

## Screenshots

FYI they are slightly outdated - I've replaced ✅ icon for "on-app-deployed" to 🚀 , as the same icon was used for "on-successful-sync" trigger.

### Mattermost
image

### Discord

image

### Teams

image

### Slack (no change here)

image

## Testing

Follow the same instruction as in #1220, but also configure all other platforms.

## Related issue(s)

#1175


helm/botkube/values.yaml
```diff
@@ -564,12 +564,15 @@ sources:
             type: Static
             static:
               values: [*argocd-group]
+      # -- Config contains configuration for ArgoCD source plugin.
+      # This section lists only basic options, and uses default triggers and templates which are based on ArgoCD Notification Catalog ones.
```
huseyinbabal, line 568: What about giving the catalog link here? I meant this URL 👉 https://argocd-notifications.readthedocs.io/en/latest/catalog/

internal/source/argocd/default-config.yaml
```diff
@@ -0,0 +1,364 @@
+log:
+  level: "info"
+
+interactivity:
```
huseyinbabal, line 4: Maybe it is better add a field description, "Not all the platforms support this. Only Slack for now". Or something relevant.

huseyinbabal (approved):
🚀 with minor comments.

Issue #1188, opened 2023-08-15
Support openai base URL support for doctor plugin


huangyingting:
## Overview
Currently doctor only requires API Key, however, there are some cases require a base URL, for example, 
- 3rd llama based models offers openai API support, 
- Azure OpenAI service
- OpenAI proxy
go-gpt3 supports overriding baseURL in client option, so would be great have this support in doctor plugin
https://pkg.go.dev/github.com/PullRequestInc/go-gpt3#WithBaseURL

## Acceptance Criteria
doctor plugin add support to configure openai base URL

## Reason

Currently doctor only requires API Key, however, there are some cases require a base URL, for example, 
- 3rd llama based models offers openai API support, 
- Azure OpenAI service
- OpenAI proxy
go-gpt3 supports overriding baseURL in client option, so would be great have this support in doctor plugin
https://pkg.go.dev/github.com/PullRequestInc/go-gpt3#WithBaseURL

PR #1232, opened 2023-09-07
Support custom OpenAI API params for Doctor plugin

pkosiec:
## Description

- Support custom OpenAI API endpoint for Doctor plugin

## Related issue(s)

Resolves https://github.com/kubeshop/botkube/issues/1188

### DIFF
--- a/internal/executor/doctor/executor.go
+++ b/internal/executor/doctor/executor.go
@@ -26,8 +26,11 @@ var (
 )
 
 type Config struct {
-	APIBaseURL string `yaml:"apiBaseUrl"`
-	APIKey     string `yaml:"apiKey"`
+	APIBaseURL     string `yaml:"apiBaseUrl"`
+	APIKey         string `yaml:"apiKey"`
+	DefaultEngine  string `yaml:"defaultEngine"`
+	OrganizationID string `yaml:"organizationID"`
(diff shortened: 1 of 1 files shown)


internal/executor/doctor/executor.go
```diff
@@ -153,11 +160,16 @@ func (d *Executor) Help(context.Context) (api.Message, error) {
...
 	defer d.l.Unlock()
-	if cfg.ApiKey == "" {
+	if cfg.APIKey == "" {
 		return nil, fmt.Errorf("OpenAPI API Key cannot be empty. You generate it here: https://platform.openai.com/account/api-keys")
 	}
+
+	var opts []gpt3.ClientOption
+	if cfg.APIBaseURL != "" {
+		opts = append(opts, gpt3.WithBaseURL(cfg.APIBaseURL))
+	}
```
mszostok, line 170: I didn't test that but as mentioned, it looks like it should work. However, to be on the safe side please add 2 more options:
- `WithOrg`
- `WithDefaultEngine`

because I'm afraid that if the different base URL will be used, the engine name may be also different or some org id param will be required.

mszostok (approved):

pkosiec:
Merging as the plugin is not tested as a part integration tests, and all other pipelines are green 👍

PR #1236, opened 2023-09-07
Restart crashed plugins

mszostok:
## Description

- Restart crashed plugins based on  https://github.com/kubeshop/botkube/pull/1204
  - [x] change the `runningProcesses` to support plugin name in the context of a given configuration
  - [x] fix logging statements to make them more readable
  - [x] apply suggestions from https://github.com/kubeshop/botkube/pull/1204
  - [ ] add e2e tests
  - [x] IMPORTANT: fixed key names used for string enabled plugins in map.
  - [x] `INFO[2023-09-07T17:05:44+02:00] Starting server on %q...:2115                 component="Incoming Webhook Server`

- still todo:
  - add option to print plugin status or add it to list executors/sources as a new column
  - add e2e tests cases
  - update documentation

## Testing

- [ ] add e2e tests

Co-authored-by: Josef Karasek <josef@kubeshop.io>

### DIFF
--- a/helm/botkube/values.yaml
+++ b/helm/botkube/values.yaml
@@ -1187,6 +1187,13 @@ plugins:
     enabled: true
     port: 2115
     targetPort: 2115
+    # -- Botkube Restart Policy on plugin failure.
+  restartPolicy:
+    # -- Restart policy type. Allowed values: "RestartAgent", "DeactivatePlugin".
+    type: "DeactivatePlugin"
+    # -- Number of restarts before policy takes into effect.
+    threshold: 10
+  healthCheckInterval: 10s
 
 # -- Configuration for synchronizing Botkube configuration.
 config:
--- a/internal/source/scheduler.go
+++ b/internal/source/scheduler.go
@@ -3,12 +3,17 @@ package source
 import (
 	"context"
 	"fmt"
+	"sync"
 
 	"github.com/sirupsen/logrus"
 	"gopkg.in/yaml.v3"
 
 	"github.com/kubeshop/botkube/pkg/api/source"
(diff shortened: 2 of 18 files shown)


helm/botkube/values.yaml
```diff
@@ -1187,6 +1187,13 @@ plugins:
     enabled: true
     port: 2115
     targetPort: 2115
+    # -- Botkube Restart Policy on plugin failure.
+  restartPolicy:
```
huseyinbabal, line 1191: So, to better understand: It will do the heatlhcheck each 10s and if it restarts 5 times, the plugin will be disabled with this setup, am I correct?

internal/source/scheduler.go
```diff
@@ -44,89 +49,174 @@ type StartedSource struct {
...
+		schedulerChan:        schedulerChan,
 	}
+	go s.monitorHealth(ctx)
+	return s
 }
 
 // Start starts all sources and dispatch received events.
 func (d *Scheduler) Start(ctx context.Context) error {
+	if err := d.generateConfigs(ctx); err != nil {
+		return err
```
huseyinbabal, line 112: to differentiate errors in this function, better to add error context.

huseyinbabal (approved):
Left minor comment, thanks

helm/botkube/values.yaml
mszostok: exactly, however I think that we can bump the disable threshold to ~10 because for now it's for the whole life of the plugin, until the botkube pod won't be restarted. I would like to change it to but probably in the follow-up PRs.

Issue #1238, opened 2023-09-08
Implement ArgoCD source plugin


pkosiec:
## Description

- [x] how to integrate with Argo:
         - webhooks? 
         - custom trigger?
         - custom polling for Argo events
- [x] simplify subscribing for events, instead of require creating configmaps, secrets, templates, we should sth like:
      ```yaml
       on:
         - type: "app-synced":
             enabled: true
             apps: ['foo', 'bar']
             namespace:
               include: ["foo"]
      ```
- [x] simple, but usable implementation (e.g. we can postpone some fancy stuff for later)
- [ ] #1250

pkosiec:
Implementation insights were described here: https://github.com/kubeshop/botkube/issues/1175

PR #1252, opened 2023-09-12
Fix Helm chart overrides for additional RBAC groups

pkosiec:
## Description

- Fix Helm chart overrides for additional RBAC groups
- Add group for Flux executor

Unfortunately, the array structure of RBAC groups doesn't allow properly override single properties without providing an override file.

In theory it should be possible according to Helm 3 documentation ([link](https://helm.sh/docs/intro/using_helm/#the-format-and-limitations-of---set)), but there's an undocumented caveat that you need to provide default values.yaml to keep the previous array content. See https://github.com/helm/helm/issues/11734

So, in order to make it easier for the users, this PR changes the structure to map which doesn't have such issue. As we'll release 1.4.1 soon, this should be included in the patch release.

## Testing

Check out this PR and see output for:

```
helm template --debug --dry-run --set="rbac.groups.argocd.create=true" ./helm/botkube
helm template--debug --dry-run --set="rbac.create=true" --set='rbac.staticGroupName=foo' ./helm/botkube
helm template --debug --dry-run --set="rbac.groups.argocd.create=true" --set="rbac.groups.flux.create=true" ./helm/botkube
helm template --debug --dry-run --set="rbac.groups.argocd.create=true" --set="rbac.groups.flux.create=true" --set="rbac.create=true" --set='rbac.staticGroupName=foo' ./helm/botkube
```
## Related issue(s)

#1238  / #1217


mszostok (approved):
good job 👍

PR #1256, opened 2023-09-13
Show plugin activation status

josefkarasek:
## Description

- `@Botkube list sources|executors` now prints status of each plugin with regards to plugin restarts.

image

## Testing

Enable `echo` executor:

```
executors:
  echo:
    botkube/echo:
      enabled: true
```

Enable `cm` source:

```
sources:
  cm:
    displayName: "Kubernetes Config Map Watcher"
    botkube/cm-watcher:
      context: *default-plugin-context
      enabled: true
      config:
        configMap:
          name: watcher
          namespace: botkube
          event: ADDED
```

Restart cm and echo plugins with:

```
@Botkube echo @panic
```

```
apiVersion: v1
kind: ConfigMap
metadata:
  annotations:
    die: "true"
  name: watcher
  namespace: botkube
```

Check results with:

```
@Botkube list src
@Botkube list exec
```

## Related issue(s)

https://github.com/kubeshop/botkube/issues/878


internal/plugin/health_stats.go
```diff
@@ -0,0 +1,80 @@
...
+func (h *HealthStats) GetStats(plugin string) (status string, restarts int, threshold int, timestamp string) {
+	h.RLock()
+	defer h.RUnlock()
+	if _, ok := h.pluginStats[plugin]; !ok {
+		status = pluginRunning
+		threshold = h.globalRestartThreshold
+		return
+	}
+
+	status = pluginRunning
```
huseyinbabal, line 72: Maybe move it before line 65 and remove it here? seems obsolete to me

huseyinbabal (approved):
Works as expected, left one minor comment

PR #1257, opened 2023-09-13
add report deployment connection init (botVer, k8sVer)

madebyrogal:
## Description

- add report connection init (send botVer & k8sVer)

## Testing

- run agent (connected with cloud deployment)
- check if it report new status connecting
- check if it reports botVer & k8sVer (deployment_status)

## Related issue(s)

#587


cmd/botkube-agent/main.go
```diff
@@ -198,7 +206,6 @@ func run(ctx context.Context) (err error) {
 	})
 
 	cmdGuard := command.NewCommandGuard(logger.WithField(componentLogFieldKey, "Command Guard"), discoveryCli)
-	botkubeVersion, err := findVersions(k8sCli)
 	if err != nil {
 		return reportFatalError("while fetching versions", err)
 	}
```
mszostok, line 211: please remove it.

internal/status/gql_reporter.go
```diff
@@ -36,15 +36,42 @@ type GraphQLStatusReporter struct {
...
+
+	err := r.withRetry(ctx, logger, func() error {
+		var mutation struct {
+			Success bool `graphql:"reportDeploymentConnectionInit(id: $id, botkubeVersion: $botkubeVersion, k8sVer: $k8sVer)"`
+		}
+		variables := map[string]interface{}{
+			"id":             graphql.ID(r.gql.DeploymentID()),
+			"botkubeVersion": version.Info().Version,
+			"k8sVer":         k8sVer,
+		}
```
mszostok, line 63: **nit:** this is static, so it can be extracted from the retry func.

internal/status/gql_reporter.go
```diff
@@ -36,15 +36,42 @@ type GraphQLStatusReporter struct {
...
+		err := r.gql.Client().Mutate(ctx, &mutation, variables)
+		return err
```
mszostok, line 65: **nit:** if you have that success indicator, you can also test it.
```suggestion
		err := r.gql.Client().Mutate(ctx, &mutation, variables)
		if err != nil {
			return err
		}
		if !mutation.Success {
			return errors.New("failed to report startup")
		}
		return nil
```

internal/status/reporter.go
```diff
@@ -3,23 +3,38 @@ package status
...
+	ReportDeploymentConnectionInit(ctx context.Context, k8sVer string) error
 	ReportDeploymentStartup(ctx context.Context) error
 	ReportDeploymentShutdown(ctx context.Context) error
 	ReportDeploymentFailure(ctx context.Context, errMsg string) error
 	SetResourceVersion(resourceVersion int)
+	SetLogger(logger logrus.FieldLogger)
 }
 
-func GetReporter(remoteCfgEnabled bool, logger logrus.FieldLogger, gql GraphQLClient, resVerClient ResVerClient, cfgVersion int) StatusReporter {
+func GetReporter(remoteCfgEnabled bool, gql GraphQLClient, resVerClient ResVerClient, logger *logrus.FieldLogger) StatusReporter {
```
mszostok, line 21: the `logrus.FieldLogger` it already an interface definition, so you should use a pointer. With Go is also better to avoid `if else` statements, and just return early, so I would propose sth like:
```go
func GetReporter(remoteCfgEnabled bool, gql GraphQLClient, resVerClient ResVerClient, log logrus.FieldLogger) StatusReporter {
	if remoteCfgEnabled {
		log  = withDefaultLogger(log)
		return newGraphQLStatusReporter(
			log.WithField("component", "GraphQLStatusReporter"),
			gql,
			resVerClient,
		)
	}

	return newNoopStatusReporter()
}

func withDefaultLogger(log logrus.FieldLogger) logrus.FieldLogger {
	if log != nil {
		return log
	}
	return loggerx.New(config.Logger{
		Level:         "info",
	})
}
```

or 
```go
func GetReporter(remoteCfgEnabled bool, gql GraphQLClient, resVerClient ResVerClient, log logrus.FieldLogger) StatusReporter {
	if remoteCfgEnabled {
		if log == nil {
			log = loggerx.New(config.Logger{
				Level: "info",
			})
		}

		return newGraphQLStatusReporter(
			log.WithField("component", "GraphQLStatusReporter"),
			gql,
			resVerClient,
		)
	}

	return newNoopStatusReporter()
}
```

up to you.

internal/status/gql_reporter.go
madebyrogal: TBH I don't see relation between static and retry. And what do you mean that "this is static"?

internal/status/reporter.go
madebyrogal: Got you really nice

internal/status/gql_reporter.go
mszostok: By static, I mean that they don't change between each retry execution. By having it in the retry function, you are going to always recreate it and GC will need to clean the previously allocated variables.

Because of that, you can just have it outside the retry and use closure function. 

However, it not that big deal, so that's the reason I used also "**nit**" 🙂

madebyrogal: Make sens. Fixed

PR #1264, opened 2023-09-18
Fix panic for Cloud Slack reconnect and JSON schema for K8s configuration

pkosiec:
## Description

- Include default node message exclusion in JSON schema for K8s plugin
- Fix panic for Cloud Slack reconnect


pkg/bot/slack_cloud_test.go
```diff
@@ -81,24 +81,24 @@ func TestWithRetriesFunc(t *testing.T) {
 
 	t.Run("Respected canceled context", func(t *testing.T) {
```
mszostok, line 82: the test name should change.  BTW shouldn't we add it just as a new test?

internal/source/kubernetes/source.go
```diff
@@ -29,6 +29,13 @@ import (
 
 var _ source.Source = (*Source)(nil)
 
+var (
+	// configJSONSchema JSON schema with duplications,
+	// as some UI components (e.g. https://github.com/rjsf-team/react-jsonschema-form) don't support nested defaults for definitions.
+	//go:embed config-jsonschema.json
+	configJSONSchema string
```
mszostok, line 36: From what I've checked, only the `".*nf_conntrack_buckets.*"` was added. But it was added in a valid place, so what do you mean by:
> JSON schema with duplications

pkg/bot/slack_cloud.go
```diff
@@ -161,6 +160,7 @@ func withRetries(ctx context.Context, log logrus.FieldLogger, maxRetries int, fn
 }
 
 func (b *CloudSlack) start(ctx context.Context) error {
+	b.messages = make(chan *pb.ConnectResponse, platformMessageChannelSize)
 	defer b.shutdown()
```
mszostok, line 164: WDYT about scoping this? 
```suggestion
	messages := make(chan *pb.ConnectResponse, platformMessageChannelSize)
	defer b.shutdown(messages)
```

This way it will be easier to track. I tested, and it can be easily shared only with `startMessageProcessor` and `shutdown` just via input arg. For me it will be a bit cleaner 🙂

pkg/bot/slack_cloud.go
```diff
@@ -111,7 +111,6 @@ func NewCloudSlack(log logrus.FieldLogger,
 		clusterName:      clusterName,
 		realNamesForID:   map[string]string{},
 		msgStatusTracker: NewSlackMessageStatusTracker(log, client),
-		messages:         make(chan *pb.ConnectResponse, platformMessageChannelSize),
 		messageWorkers:   pool.New().WithMaxGoroutines(platformMessageWorkersCount),
```
mszostok, line 114: tested e2e, unfortunately also got panic:
```
panic: send on closed channel

goroutine 171 [running]:
github.com/sourcegraph/conc/pool.(*Pool).Go(0xc00098c280, 0xc0009ef680)
	/Users/mszostok/workspace/go/pkg/mod/github.com/sourcegraph/conc@v0.3.0/pool/pool.go:63 +0xe7
github.com/kubeshop/botkube/pkg/bot.(*CloudSlack).startMessageProcessor(0xc000a92480, {0x39791b8?, 0xc0009ec7d0})
	/Users/mszostok/workspace/go/src/github.com/kubeshop/botkube/pkg/bot/slack_cloud.go:228 +0x130
created by github.com/kubeshop/botkube/pkg/bot.(*CloudSlack).start in goroutine 99
	/Users/mszostok/workspace/go/src/github.com/kubeshop/botkube/pkg/bot/slack_cloud.go:203 +0x61e
```

we shutdown worker service, we need to also recreated it.

internal/source/kubernetes/source.go
pkosiec: I moved this comment from the previous place:

image


Basically, in the JSON schema there are some duplications because the JSON schema form component doesn't support too deeply nested defaults. That's why we couldn't define some nested objects once and refer to them via `$ref`.

pkg/bot/slack_cloud.go
pkosiec: Thank you, fixed here: https://github.com/kubeshop/botkube/pull/1264/commits/1d4becd23492bfd28c6fc67c0bb44ad17e851ca7

pkg/bot/slack_cloud_test.go
pkosiec: Actually I wanted to test it in unit test but it became too hard (as essentially I would need to test the whole `start` that connects to grpc service) so I abandoned the idea but it seems I forgot to revert all changes in this file. Fixed that.

mszostok (approved):
tested e2e, LGTM 👍

PR #1268, opened 2023-09-19
Document Doctor configuration options in Helm values.yaml, fix custom engine override

pkosiec:
## Description

- Document Doctor configuration options in Helm values.yaml
- Fix custom engine override
- Add logging for Doctor
- Fix logging configuration for Flux

## Screenshots

image

Screenshot 2023-09-20 at 10 26 54


internal/executor/doctor/executor.go
```diff
@@ -185,13 +184,16 @@ func (d *Executor) getGptClient(cfg *Config) (gpt3.Client, error) {
 		return nil, fmt.Errorf("OpenAPI API Key cannot be empty. You generate it here: https://platform.openai.com/account/api-keys")
 	}
 
-	var opts []gpt3.ClientOption
-	if cfg.APIBaseURL != "" {
-		opts = append(opts, gpt3.WithBaseURL(cfg.APIBaseURL))
+	defaultEngine := gpt3.TextDavinci003Engine
```
mszostok, line 187: Maybe we should do the same for other properties? Because described defaults are assumptions based on the underlying implementation. While bumping deps, it may change so OS and cloud will work differently as OS won't have those taken from JSON Schema.

mszostok (approved):

internal/executor/doctor/config-jsonschema.json
```diff
@@ -0,0 +1,89 @@
...
+      "title": "Organization ID",
+      "default": ""
+    },
+    "userAgent": {
+      "description": "User agent to use for requests",
+      "type": "string",
+      "title": "User Agent",
+      "default": "go-gpt3"
+    },
+    "log": {
```
mszostok, line 36: personally, I'm not sure if that should be exposed to the users 🤔 I use it only during development, and sometimes during debugging, so that was the reason why I didn't include it in other plugins  (flux, gh-events) to make them feel like sth easy to configure 😄

internal/executor/doctor/executor.go
```diff
@@ -178,29 +158,53 @@ func (d *Executor) Help(context.Context) (api.Message, error) {
...
+		gpt3.WithUserAgent(userAgent),
+		gpt3.WithOrg(orgID),
 	}
+
+	log.WithFields(logrus.Fields{
+		"baseURL":       baseURL,
+		"defaultEngine": defaultEngine,
+		"userAgent":     userAgent,
+		"orgID":         orgID,
+	}).Debugf("Creating GPT3 client with API key starting with %q...", stringsutil.ShortenString(cfg.APIKey, printAPIKeyCharCount))
```
mszostok, line 205:
```suggestion
		"apiKeyPrefix":         stringsutil.ShortenString(cfg.APIKey, printAPIKeyCharCount),
	}).Debugf("Creating GPT3 client...")
```

would be better to keep the same style, otherwise, you could do:
`Creating GPT3 client with API key starting with %q for endpoint %q using %q engine in the context of %q org id` 😄

mszostok (approved):
LGTM 👍

internal/executor/doctor/executor.go
pkosiec: Yes and no, but ok 😄

internal/executor/doctor/config-jsonschema.json
pkosiec: This is something what we do in most of the plugins, and this is the only way to configure this when using cloud. Yes, we should have consistency here, but I'd rather expose this option to users.

PR #1271, opened 2023-09-21
Add segment analytics to CLI

josefkarasek:
## Description

- Add segment analytics to CLI

## Testing

Build CLI using goreleaser - provide segment API key in envar `CLI_ANALYTICS_API_KEY` 

Test that individual commands report usage to segment

image

Test that `--opt-out-analytics` skips reporting analytics.

## Related issue(s)

https://github.com/kubeshop/botkube-cloud/issues/559


internal/cli/config.go
```diff
@@ -18,7 +18,8 @@ var (
 
 // Config is botkube cli config
 type Config struct {
-	Token string `json:"token"`
+	Token    string `json:"token"`
+	Identity string `json:"identity"`
```
pkosiec, line 22: **[TO DISCUSS]**
I think it would be great to add the analytics property to config so that people won't need to disable telemetry every time. What do you think?

and maybe have a command like `botkube config set analytics=false` or something like this? 🤔 it could work in generic way, so someone could also execute `botkube config set token=...` manually 🤔 

Let's ask @mszostok what he thinks about this approach 👍

internal/cli/analytics/reporter.go
```diff
@@ -0,0 +1,29 @@
...
+
+// Reporter defines behavior for reporting analytics.
+type Reporter interface {
+	ReportCommand(cmd string) error
+	ReportError(err error, cmd string) error
+	Close()
+}
+
+// NewReporter creates a new Reporter instance.
+func NewReporter() Reporter {
```
pkosiec, line 20: I think it would be better to return here noop reporter in case the analytics are disabled. This would be consistent without approach with Botkube Agent. In that way we won't report anything even if someone by accident executes a `Report` method in the code. WDYT?

Also, I'd suggest renaming this to `GetReporter()` or something like this as this is a conditional constructor and it's not clear from the usage perspective.

internal/cli/login/login.go
```diff
@@ -29,7 +29,10 @@ func Run(ctx context.Context, w io.Writer, opts Options) error {
 		return err
 	}
 
-	c := cli.Config{Token: t.Token}
+	c := &cli.Config{}
+	_ = c.Read()
```
pkosiec, line 33: why this error is ignored?

internal/cli/analytics/segment_reporter.go
```diff
@@ -0,0 +1,119 @@
...
+	})
+	if err != nil {
+		return errors.Wrap(err, "failed to report error")
+	}
+	return nil
+}
+
+// Close closes the reporter.
+func (r *SegmentReporter) Close() {
+	_ = r.client.Close()
```
pkosiec, line 78: Why this error is not propagated? 🤔 I think we should propagate it to the caller

internal/cli/analytics/segment_reporter.go
```diff
@@ -0,0 +1,119 @@
...
+	err := r.client.Enqueue(segment.Identify{
+		AnonymousId: machineID,
+		Traits:      newProperties(machineID, isLoggedIn),
+	})
+	if err != nil {
+		return isLoggedIn, errors.Wrap(err, "failed to report identity")
+	}
+
+	c.Identity = machineID
+	if err := c.Save(); err != nil {
```
pkosiec, line 101: I don't understand it - why do we save the machine id and read the config, if the machine id is always got in the reporting? 

I think we should get rid of save/load of the machine ID and execute `machineid.ID()` every time, as you do right now. But, please move `machineid.ID()` to `reportAndSaveIdentity`, to not expect that each`Report...()` will call it every time. If that's always done, let's do it here 👍

internal/cli/analytics/segment_reporter.go
```diff
@@ -0,0 +1,119 @@
...
+func (r *SegmentReporter) reportAndSaveIdentity(machineID string) (bool, error) {
+	c := &cli.Config{}
+	if err := c.Read(); err != nil {
+		return false, errors.Wrap(err, "failed to read config")
+	}
+
+	isLoggedIn := c.Token != ""
+	if c.Identity == machineID {
+		return isLoggedIn, nil
+	}
```
pkosiec, line 90: We don't need this check (ergo: we don't need to save the machine id in config). We can call `Identify` multiple times, so we don't need to store the machineID. Yes, `isLoggedIn` is good metadata and we should include it, but this is something which we done based on `c.Token`, so.. 🤔

internal/cli/analytics/segment_reporter.go
```diff
@@ -0,0 +1,119 @@
...
+func newProperties(id string, cloudLogin bool) map[string]interface{} {
+	v := defaultCliVersion
+	if vals := version.Get(); vals != nil {
+		v = vals.Version
+	}
+	return map[string]interface{}{
+		"OS":          runtime.GOOS,
+		"version":     v,
+		"machine_id":  id,
+		"cloud_login": cloudLogin,
```
pkosiec, line 117: Maybe something like this? 🤔 
```suggestion
		"cloud_logged_in": cloudLogin,
```

In theory we could get the user email based on the access token, but this is something we can add on demand later.

pkosiec (commented):
Here are a few comments based on code review. Didn't test it, but the screenshots looks good 👍

internal/cli/analytics/segment_reporter.go
josefkarasek: I'm trying to avoid crashing/spitting out errors on telemetry errors - at some level this err is always surpressed

cmd/cli/cmd/telemetry/disable.go
```diff
@@ -0,0 +1,37 @@
...
+)
+
+// NewEnable returns a new cobra.Command for disabling telemetry.
+func NewDisable() *cobra.Command {
+	enable := &cobra.Command{
+		Use:   "disable",
+		Short: "Disable Botkube telemetry",
+		Example: heredoc.WithCLIName(`
+			# The Botkube CLI tool collects anonymous usage analytics.
+			# This data is only available to the Botkube authors and helps us improve the tool.
```
pkosiec, line 17: WDYT about changing that to:

```
To improve the user experience, Botkube collects anonymized data. It does not collect any identifying information, and all analytics are used only as aggregated collection of data to improve Botkube and adjust its roadmap. 

Read our privacy policy at https://docs.botkube.io/privacy
```

cmd/cli/cmd/telemetry/enable.go
```diff
@@ -0,0 +1,37 @@
...
+)
+
+// NewEnable returns a new cobra.Command for enabling telemetry.
+func NewEnable() *cobra.Command {
+	enable := &cobra.Command{
+		Use:   "enable",
+		Short: "Enable Botkube telemetry",
+		Example: heredoc.WithCLIName(`
+			# The Botkube CLI tool collects anonymous usage analytics.
+			# This data is only available to the Botkube authors and helps us improve the tool.
```
pkosiec, line 17: Please update this description as well, based on my previous comment

cmd/cli/cmd/telemetry/telemetry.go
```diff
@@ -0,0 +1,17 @@
+package telemetry
+
+import "github.com/spf13/cobra"
+
+// NewCmd returns a new cobra.Command subcommand for telemetry-related operations.
+func NewCmd() *cobra.Command {
+	root := &cobra.Command{
+		Use:   "telemetry",
+		Short: "This command consists of subcommands to disable or enable telemetry",
```
pkosiec, line 9:
```suggestion
		Short: "Configure collection of anonymous analytics",
```

internal/cli/analytics/reporter.go
```diff
@@ -0,0 +1,39 @@
...
+
+// Reporter defines behavior for reporting analytics.
+type Reporter interface {
+	ReportCommand(cmd string) error
+	ReportError(err error, cmd string) error
+	Close()
+}
+
+// NewReporter creates a new Reporter instance.
+func GetReporter() Reporter {
```
pkosiec, line 22: Can we print just a single line which reporter is used when the global verebose flag (`-v/--verbose`) is used? That would be helpful in debugging scenario.

internal/cli/config.go
```diff
@@ -18,19 +18,46 @@ var (
...
 }
 
-// NewConfig creates new Config from local data
-func NewConfig() (*Config, error) {
+// TelemetrySwitch controls telemetry
+type TelemetrySwitch string
+
+const (
+	// TelemetryEnabled is a value for enabled telemetry
+	TelemetryEnabled TelemetrySwitch = "enabled"
```
pkosiec, line 30: This value is not used at all. Does it make sense to use enum instead of `bool`? I know there could be multiple levels of telemetry at some point, but for now it looks like an overkill to me 🤔

pkosiec (commented):
Overall it looks good to me, just a few small comments 👍

internal/cli/config.go
josefkarasek: I forgot to include it in telemetry/enable.go. It is used now.

.github/workflows/branch-build.yml
```diff
@@ -48,6 +48,7 @@ jobs:
         run: make release-snapshot
         env:
           ANALYTICS_API_KEY: ${{ secrets.ANALYTICS_API_KEY }}
+          CLI_ANALYTICS_API_KEY: ${{ secrets.CLI_ANALYTICS_API_KEY }}
```
pkosiec, line 51: This is not needed, right? As `release-snapshot` doesn't build CLI?

internal/cli/config.go
pkosiec: ok 👍

internal/cli/analytics/segment_reporter.go
```diff
@@ -0,0 +1,99 @@
...
+	_ = r.client.Close()
+}
+
+func newProperties(id string, cloudLogin bool) map[string]interface{} {
+	v := defaultCliVersion
+	if vals := version.Get(); vals != nil {
+		v = vals.Version
+	}
+	return map[string]interface{}{
+		"OS":              runtime.GOOS,
```
pkosiec, line 94: can we also report architecture (GOARCH)? I've just realized we don't have a clue if someone runs e.g. ARM or AMD64, and that could be helpful with debugging such issues:
image

pkosiec (approved):
LGTM but please apply my comments before merge

internal/cli/analytics/segment_reporter.go
josefkarasek: image

Issue #890, opened 2022-12-15
Return the JSON status object from the `healthz` endpoint


mszostok:
## Overview
_(Discussed on the dev-team meeting on Dec 15)_

In the https://github.com/kubeshop/botkube/issues/881 we introduce a dedicated Botkube `heatlhz` endpoint that returns 200 (OK). 

As a part of this task, we should return JSON with information about Botkube core health:
- and enabled plugins (after https://github.com/kubeshop/botkube/issues/878)
- and communications platform statuses (if possible)

- Ensure the bot + plugin statuses can be used as a part of `reportDeploymentHeartbeat` GraphQL query

## Reason

Provide visibility about Botkube and related components' health.

PR #1276, opened 2023-09-27
Healthz improve endpoint

madebyrogal:
## Description

- add more details to healthz endpoint

## Testing

- make bot connected to all available platforms
- run bot-agent and trigger /healthz endpoint

## Related issue(s)

#890

### DIFF
--- a/cmd/botkube-agent/main.go
+++ b/cmd/botkube-agent/main.go
@@ -195,7 +195,7 @@ func run(ctx context.Context) (err error) {
 	defer pluginManager.Shutdown()
 
 	// Health endpoint
-	healthChecker := health.NewChecker(conf, pluginHealthStats)
+	healthChecker := health.NewChecker(ctx, conf, pluginHealthStats)
 	healthSrv := healthChecker.NewServer(logger.WithField(componentLogFieldKey, "Health server"), conf.Settings.HealthPort)
 	errGroup.Go(func() error {
 		defer analytics.ReportPanicIfOccurs(logger, reporter)
@@ -317,6 +317,7 @@ func run(ctx context.Context) (err error) {
 			sinkNotifiers = append(sinkNotifiers, wh)
 		}
 	}
+	healthChecker.SetBots(bots)
 
 	if conf.ConfigWatcher.Enabled {
 		restarter := reloader.NewRestarter(
--- a/cmd/botkube-agent/main.go
+++ b/cmd/botkube-agent/main.go
@@ -195,7 +195,7 @@ func run(ctx context.Context) (err error) {
 	defer pluginManager.Shutdown()
 
 	// Health endpoint
-	healthChecker := health.NewChecker()
+	healthChecker := health.NewChecker(conf, pluginHealthStats)
 	healthSrv := healthChecker.NewServer(logger.WithField(componentLogFieldKey, "Health server"), conf.Settings.HealthPort)
 	errGroup.Go(func() error {
 		defer analytics.ReportPanicIfOccurs(logger, reporter)
--- a/cmd/botkube-agent/main.go
+++ b/cmd/botkube-agent/main.go
@@ -28,6 +28,7 @@ import (
 	intconfig "github.com/kubeshop/botkube/internal/config"
 	"github.com/kubeshop/botkube/internal/config/reloader"
 	"github.com/kubeshop/botkube/internal/config/remote"
+	"github.com/kubeshop/botkube/internal/health"
 	"github.com/kubeshop/botkube/internal/heartbeat"
 	"github.com/kubeshop/botkube/internal/httpx"
 	"github.com/kubeshop/botkube/internal/insights"
@@ -54,7 +55,6 @@ const (
 	botLogFieldKey            = "bot"
 	sinkLogFieldKey           = "sink"
 	commGroupFieldKey         = "commGroup"
-	healthEndpointName        = "/healthz"
 	printAPIKeyCharCount      = 3
 	reportHeartbeatInterval   = 10
 	reportHeartbeatMaxRetries = 30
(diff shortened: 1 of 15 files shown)


internal/health/health.go
```diff
@@ -0,0 +1,186 @@
...
+func (h *Checker) SetBots(bots map[string]bot.Bot) {
+	h.bots = bots
+}
+
+// SetSinks sets platform sink instances.
+func (h *Checker) SetSinks(sinks []notifier.Sink) {
+	h.sinks = sinks
+}
+
+func (h *Checker) getSourcePluginsStatuses(plugins map[string]pluginStatuses) {
```
mszostok, line 134: **nit** I would rename it to `collect` instead of `get`. It doesn't return the result, so collect, emphasis more that it will be collected into a given map. In general, it's not a best practice to mutate method param, but I'm fine with it here 👍

internal/health/health.go
```diff
@@ -0,0 +1,186 @@
...
+}
+
+type platformStatuses map[string]bot.Status
+
+// Status defines bot agent status.
+type Status struct {
+	Botkube   botStatus
+	Plugins   map[string]pluginStatuses
+	Platforms platformStatuses
+}
```
mszostok, line 57: you should specify the `json` tags, otherwise it uses the fields names, which looks odd:
`{"Botkube":{"Status":"Unhealthy"},"Plugins":{},"Platforms":{}}`

You should:
1. Move the `bot.Status` object into this package. Use it bot for sink and bots. Add JSON tags.
2. Add JSON tags to `Status` with `omittempty`:
    ```go
    // Status defines bot agent status.
    type Status struct {
	    Botkube   botStatus                 `json:"botkube"`
	    Plugins   map[string]pluginStatuses `json:"plugins,omitempty"`
	    Platforms platformStatuses          `json:"platforms,omitempty"`
    }
    
    ```

internal/health/health.go
```diff
@@ -0,0 +1,186 @@
+package health
```
mszostok, line 1: it would be good to have some minimal tests, e.g. for not ready it can be sth like:
```go
func TestServeHTTP(t *testing.T) {
	// given
	checker := &Checker{
		config: &config.Config{},
	}
	expectedStatus := checker.GetStatus()

	req, err := http.NewRequest("GET", "/", nil)
	require.NoError(t, err)
	rr := httptest.NewRecorder()

	// when
	checker.ServeHTTP(rr, req)

	// then
	assert.Equal(t, http.StatusServiceUnavailable, rr.Code)
	assert.Equal(t, "application/json", rr.Header().Get("Content-Type"))

	var resp Status
	err = json.Unmarshal(rr.Body.Bytes(), &resp)
	require.NoError(t, err)

	assert.Equal(t, BotkubeStatusUnhealthy, resp.Botkube.Status)
	assert.Equal(t, resp.Botkube.Status, expectedStatus.Botkube.Status)
}

```

pkg/bot/slack_cloud.go
```diff
@@ -108,21 +112,27 @@ func NewCloudSlack(log logrus.FieldLogger,
 		clusterName:      clusterName,
 		realNamesForID:   map[string]string{},
 		msgStatusTracker: NewSlackMessageStatusTracker(log, client),
+		status:           StatusUnknown,
+		maxRetries:       maxRetries,
+		failuresNo:       0,
+		failureReason:    "",
```
mszostok, line 118: All reasons work good. However, I got the `Unknown` status until I run at least one command from Slack. The reason is that you have `setFailureReason` in `handleStreamMessage` and the `b.start` is long-running function, exits only on error so the `setFailureReason` from `retry.Do` doesn't work as expected. In general, you should fix it so when started and there are no errors, the status is `Healthy`.

mszostok (commented):
The overall code LGTM 👍 but there are a few things that need to be fixed before merged 👍

internal/health/health.go
```diff
@@ -0,0 +1,186 @@
...
+		Botkube: botStatus{
+			Status: h.getBotkubeStatus(),
+		},
+		Plugins:   pluginsStats,
+		Platforms: h.getPlatformsStatus(),
+	}, nil
+}
+
+// ServeHTTP serves status on health endpoint.
+func (h *Checker) ServeHTTP(resp http.ResponseWriter, _ *http.Request) {
```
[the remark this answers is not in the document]
madebyrogal: Re-implemented it with your guidance. Works fine also if app is not ready returns json with available statuses.
On 500 returns string error

pkg/bot/slack_cloud.go
madebyrogal: fixed

cmd/botkube-agent/main.go
```diff
@@ -192,8 +192,8 @@ func run(ctx context.Context) (err error) {
 	defer pluginManager.Shutdown()
 
 	// Health endpoint
-	healthChecker := healthChecker{applicationStarted: false}
-	healthSrv := newHealthServer(logger.WithField(componentLogFieldKey, "Health server"), conf.Settings.HealthPort, &healthChecker)
+	healthChecker := health.NewChecker(ctx, conf, pluginHealthStats)
+	healthSrv := healthChecker.NewServer(logger.WithField(componentLogFieldKey, "Health server"), conf.Settings.HealthPort)
```
mszostok, line 196: let's move starting the health server to line 189:
```go
	// Health endpoint
	healthChecker := health.NewChecker(ctx, conf, pluginHealthStats)
	healthSrv := healthChecker.NewServer(logger.WithField(componentLogFieldKey, "Health server"), conf.Settings.HealthPort)
	errGroup.Go(func() error {
		defer analytics.ReportPanicIfOccurs(logger, reporter)
		return healthSrv.Serve(ctx)
	})
```

this way it will be available before the `	err = pluginManager.Start(ctx)` which may take some time as it downloads all enabled plugins.

mszostok (approved):
Great work! 

Please apply the two final comments before merging :)

PR #1292, opened 2023-10-10
Use main chart in tests

huseyinbabal:
## Description

- Main charts are published periodically. Now we use that charts in tests


josefkarasek (approved):
lgtm

PR #1294, opened 2023-10-10
Add thread mate plugin

mszostok:
## Description

- Add thread mate plugin
  - randomize initial next idx value
  - opt to define ack message emoji
  - opt to bind for channel messages
  - persist processed activity

## Testing

1. Build plugin: `PLUGIN_TARGETS="thread-mate" make build-plugins-single`
2. Create config:
    <details><summary>Details</summary>
    <p>

    ```yaml
    communications:
      default-group:
        socketSlack:
          enabled: true
          channels:
            general:
              name: backoffice
              bindings:
                executors:
                  - helping-hands-assigner
                sources: []
            default:
              name: helping-hands
              textMessageBindings:
                event: ChannelMessage # currently only ChannelMessage is supported. Other options that we may add in the future: ChannelMessage, ThreadMessage, AllMessages
                messages:
                  - text:
                      include: ".*" # all messages
                    users:
                      exclude: ["U02KKBR5PE1:Mateusz", "etc"]
                    processedEmojiIndicator: "robot_face" # empty - disabled
                    command: "thread-mate pick --bk-cmd-header 'Help assistance' "
                    executors:
                      - helping-hands-assigner
              bindings:
                executors: []
                sources: []
    
          appToken: "xapp-"
          botToken: "xoxb-"
    
    executors:
      helping-hands-assigner:
        botkube/thread-mate:
          enabled: true
          context:
            rbac:
              group:
                type: Static
                static:
                  values: [ "system:masters" ]
          config:
            pick:
              # userCooldownTime represents the time that must elapse for a given user's message before the "pick" method can be used again.
              userCooldownTime: 3m
              #messageTemplate: |
                # ...
            roundRobin:
              assignees: ["U02KKBR5PE1:Mateusz", "etc"]
              groupName: "botkube"
    
    aliases:
      act-all:
        command: thread-mate get activity
        displayName: "Get thread activity"
      act:
        command: thread-mate get activity --assignee-ids
        displayName: "Get thread activity"
    
    plugins:
      cacheDir: "/tmp/plugins"
      repositories:
        botkube:
          url: http://host.k3d.internal:3010/botkube.yaml
    
    settings:
      log:
        level: "debug"
        formatter: text
      clusterName: "labs"
      upgradeNotifier: false
    
    analytics:
      disable: true
    ```
    
    </p>
    </details> 

3. Start plugin server: `env PLUGIN_SERVER_HOST=http://host.k3d.internal go run test/helpers/plugin_server.go`
4. Run Botkube: `env LOG_LEVEL_EXECUTOR_BOTKUBE_THREAD_MATE=debug BOTKUBE_CONFIG_PATHS="./helm/botkube/values.yaml,/tmp/mate-cfg.yaml" go run ./cmd/botkube-agent/main.go`
5.  Test:
    - round-robin rotation
    - getting DMs
    - use custom emoji
    - cooldown window for using `pick`
        - if a user sends e.g. 3 messages on main thread within X minutes, only the first one is used to select a support person. 
    - `@Botkube thread-mate get activity`:
        - takeover button
        - mark as resolved
        - export data as markdown and csv
        - search options
    - define different message template
    - Botkube should ignore messages from the support person
    - Thread messages should be ignored
    - restart Botkube, activity should be preserved

## Related issue(s)

- https://github.com/kubeshop/botkube-cloud/issues/631

## Notes
Make sure that your socket slack app is subscribed to channel events
Screenshot 2023-10-11 at 22 55 24


cmd/executor/thread-mate/main.go
```diff
@@ -0,0 +1,157 @@
...
+		}, nil
+	case cmd.Get != nil && cmd.Get.Activity != nil:
+		return executor.ExecuteOutput{
+			Message: svc.GetActivity(cmd.Get.Activity, in.Context.Message),
+		}, nil
+	case cmd.Resolve != nil:
+		return executor.ExecuteOutput{
+			Message: svc.Resolve(cmd.Resolve, in.Context.Message),
+		}, nil
+	case cmd.Takeover != nil:
```
pkosiec, line 119: BTW later, in a follow-up, we could post a message in the thread that the "ticket" has been taken over by someone. Anyway, not needed for now 👍

cmd/executor/thread-mate/main.go
```diff
@@ -0,0 +1,157 @@
...
+		return executor.ExecuteOutput{
+			Message: svc.Takeover(cmd.Takeover, in.Context.Message),
+		}, nil
+	case cmd.Export != nil:
+		return executor.ExecuteOutput{
+			Message: svc.Export(cmd.Export),
+		}, nil
+	default:
+		return executor.ExecuteOutput{
+			Message: api.NewPlaintextMessage("Command not supported", false),
```
pkosiec, line 129: Can we tread no command equally with `help`? So that `@Botkube thread-mate` is effectively `@Botkube thread-mate help`. Thanks!

internal/executor/thread-mate/cm_sync.go
```diff
@@ -0,0 +1,83 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: namespace,
+		},
+		Data: map[string]string{
+			dataFieldName: data,
+		},
+	}
+
+	ctx := context.TODO()
```
pkosiec, line 40: can we pass it from above? If that will be too difficult, please at least use `context.Background()`

internal/executor/thread-mate/cm_sync.go
```diff
@@ -0,0 +1,83 @@
...
+// Get retrieves data from a ConfigMap in the specified namespace.
+func (a *ConfigMapDumper) Get(namespace, name string) (string, error) {
+	cm := &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: namespace,
+		},
+	}
+
+	ctx := context.TODO()
```
pkosiec, line 76: Same as above

internal/executor/thread-mate/system_data.go
```diff
@@ -0,0 +1,44 @@
...
+	SystemData struct {
+		sync.RWMutex
+		dirty      bool
+		roundRobin RoundRobin
+	}
+	RoundRobin struct {
+		next uint32
+		// TODO: rotationExclusion holds the assignee IDs as a key.
+		// At a later stage, we can add for each entry the endDate, so it will be automatically enabled again.
+		//rotationExclusion map[string]struct{}
```
pkosiec, line 18: Is it still needed?

internal/executor/thread-mate/service.go
```diff
@@ -0,0 +1,586 @@
...
+		assignees = append(assignees, Assignee{ID: id, DisplayName: displayName})
+	}
+
+	return &ThreadMate{
+		log:       loggerx.New(cfg.Logger),
+		cfg:       cfg,
+		assignees: assignees,
+		systemData: SystemData{
+			roundRobin: RoundRobin{
+				//nolint:gosec // //nolint:gosec // false positive
```
pkosiec, line 66:
```suggestion
				//nolint:gosec // false positive
```

internal/executor/thread-mate/service.go
```diff
@@ -0,0 +1,586 @@
...
+		btnBuilder:            api.NewMessageButtonBuilder(),
+	}
+}
+
+// Start starts the ThreadMate component.
+func (t *ThreadMate) Start() {
+	t.ongoingThreads = t.tryToLoadOldThreads(ongoingCMName)
+	t.resolvedThreads = t.tryToLoadOldThreads(resolvedCMName)
+
+	go func() {
```
pkosiec, line 82: shouldn't it take context and finish gracefully on `ctx.Done()`?

internal/executor/thread-mate/service.go
```diff
@@ -0,0 +1,586 @@
...
+		item = strings.TrimSpace(item)
+		item = extractIDFromMention(item)
+		if item == "" {
+			continue
+		}
+		assignees = append(assignees, item)
+	}
+
+	var ongoing []api.Section
+	if cmd.Type == "" || strings.EqualFold(cmd.Type, "ongoing") {
```
pkosiec, line 220: WDYT about using enum here?

pkosiec (commented):
This is just fantastic 🚀 A few small comments/questions, but overall the functionality works even better than I thought it will be!

internal/executor/x/mathx/int.go
```diff
@@ -18,8 +18,8 @@ func DecreaseWithMin(in, min int) int {
 	return in
 }
 
-// Max returns the largest of a or b.
-func Max(a, b int) int {
+// Min returns the smallest of a or b.
+func Min(a, b int) int {
```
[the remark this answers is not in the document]
mszostok: had the same problem 😄 maybe a "typo"? but dunno 😄

pkg/execute/plugin_executor.go
```diff
@@ -124,7 +136,8 @@ func (e *PluginExecutor) Execute(ctx context.Context, bindings []string, slackSt
 	}
 
 	out := interactive.CoreMessage{
-		Message: resp.Message,
+		Message:  resp.Message,
+		Messages: resp.Messages,
```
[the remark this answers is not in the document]
mszostok: I treat the `Messeges` property as the experimental one as you can read in the comment, it's only SocketSlack feature, meaning I don't want to remove it and also don't want to update all our plugins:
```go
		// Messages holds a collection of messages that should be dispatched to the user in the context of a given command execution.
		// To avoid spamming, you can specify max 15 messages.
		// Limitations:
		//   - It's available only for SocketSlack. In the future, it may be adopted across other platforms.
		//   - Message filtering is not available.
		Messages []api.Message
```

internal/executor/thread-mate/mentions.go
```diff
@@ -0,0 +1,16 @@
...
+
+func extractIDFromMention(in string) string {
+	in = strings.TrimPrefix(in, "<@")
+	in = strings.TrimSuffix(in, ">")
+	return in
+}
+
+func asMention(in string) string {
+	return fmt.Sprintf("<@%s>", in)
+}
```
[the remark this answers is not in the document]
mszostok: kinda, basically the executor plugin doesn't know what for which platform it was enabled, however in template I already use:
```
 Thanks for reaching out! Today, {{ .Assignee.ID | toMention }} will assist you in getting your Botkube up and running :botkube-intensifies:
```
so it will be just a manner of moving renderer into Botkube core. I will add a dedicated comment about it.

cmd/executor/thread-mate/main.go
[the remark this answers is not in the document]
mszostok: discussed f2f, in general you need to explicitly add it to the exclude list.

cmd/executor/thread-mate/main.go
mszostok: It would require a new feature on the socket slack side to support sth like:
```go
api.Message{
	ThreadResponseURL: "<>"
}
```
but I like the idea and it's totally doable 👍 sth to consider as the follow-up improvement 👍

internal/executor/thread-mate/service.go
mszostok: we don't have context, I would need to do some signal channel and bind it to context. The one from the `Execute` method cannot be used as this needs to be a long-running operation, not per execute.

I think that we can do it later

pkg/api/message.go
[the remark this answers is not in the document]
mszostok: The code is in the grpc_adapter, I think that we should use a generic `gopkg.in/yaml.v3` instead of `sigs.k8s.io/yaml`. However, `gopkg` works only on YAML tags. But is it a problem to have both tags? I used `gomodifytags` which make it just 1 liner to put all necessary tags in place.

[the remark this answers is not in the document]
pkosiec: > But is it a problem to have both tags?

It's ok for now, but at some point I think it would be good to unify the approach. Anyway, fine for now 👍

pkosiec (approved):
LGTM - again, well done!

PR #1295, opened 2023-10-11
Include full plugin name in reported command

pkosiec:
## Description

- Include full plugin name in reported command
- Report communication group anonymous ID
- Report Cloud Slack enabled

## Testing

No need, unit tests properly test it.

But, you can change the NoopReporter to:
```go
func (n NoopReporter) ReportBotEnabled(p config.CommPlatformIntegration, i int) error {
	fmt.Println(">>", p, i)
	return nil
}

// ReportSinkEnabled reports an enabled sink.
func (n NoopReporter) ReportSinkEnabled(p config.CommPlatformIntegration, i int) error {
	fmt.Println(">>", p, i)
	return nil
}

// ReportCommand reports a new executed command. The command should be anonymized before using this method.
func (n NoopReporter) ReportCommand(in ReportCommandInput) error {
	formatx.StructDumper().Dump("ReportCommandInput", in)
	return nil
}

```
And execute a few commands:

image

image

image

### DIFF
--- a/cmd/botkube-agent/main.go
+++ b/cmd/botkube-agent/main.go
@@ -44,6 +44,7 @@ import (
 	"github.com/kubeshop/botkube/pkg/config"
 	"github.com/kubeshop/botkube/pkg/controller"
 	"github.com/kubeshop/botkube/pkg/execute"
+	"github.com/kubeshop/botkube/pkg/maputil"
 	"github.com/kubeshop/botkube/pkg/multierror"
 	"github.com/kubeshop/botkube/pkg/notifier"
 	"github.com/kubeshop/botkube/pkg/sink"
@@ -239,10 +240,15 @@ func run(ctx context.Context) (err error) {
 	//    For example, if in both communication groups there's a Slack configuration pointing to the same workspace,
 	//	  when user executes `kubectl` command, one Bot instance will execute the command and return response,
 	//	  and the second "Sorry, this channel is not authorized to execute kubectl command" error.
-	commGroupIdx := 1
-	for commGroupName, commGroupCfg := range conf.Communications {
+	commKeys := maputil.SortKeys(conf.Communications)
+	for commGroupIdx, commGroupName := range commKeys {
+		commGroupCfg := conf.Communications[commGroupName]
+
 		commGroupLogger := logger.WithField(commGroupFieldKey, commGroupName)
-		commGroupMeta := bot.CommGroupMetadata{Name: commGroupName, Index: commGroupIdx}
+		commGroupMeta := bot.CommGroupMetadata{
+			Name:  commGroupName,
+			Index: commGroupIdx + 1,
+		}
 
 		scheduleBotNotifier := func(in bot.Bot) {
 			bots[fmt.Sprintf("%s-%s", commGroupName, in.IntegrationName())] = in
@@ -318,8 +324,6 @@ func run(ctx context.Context) (err error) {
 
 			sinkNotifiers = append(sinkNotifiers, wh)
 		}
-
-		commGroupIdx++
 	}
 	healthChecker.SetNotifiers(getHealthNotifiers(bots, sinkNotifiers))
 
--- a/cmd/botkube-agent/main.go
+++ b/cmd/botkube-agent/main.go
@@ -239,8 +239,10 @@ func run(ctx context.Context) (err error) {
 	//    For example, if in both communication groups there's a Slack configuration pointing to the same workspace,
 	//	  when user executes `kubectl` command, one Bot instance will execute the command and return response,
 	//	  and the second "Sorry, this channel is not authorized to execute kubectl command" error.
+	commGroupIdx := 1
 	for commGroupName, commGroupCfg := range conf.Communications {
 		commGroupLogger := logger.WithField(commGroupFieldKey, commGroupName)
+		commGroupMeta := bot.CommGroupMetadata{Name: commGroupName, Index: commGroupIdx}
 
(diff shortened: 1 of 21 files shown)


mszostok (approved):
LGTM 👍 if you will have some time IMO with 5 params it would be already beneficial to use struct as input, but it's just nice to have, we can do it later.

internal/analytics/reporter.go
```diff
@@ -42,3 +42,11 @@ type ReportEvent struct {
 	PluginName            string
 	AnonymizedEventFields map[string]any
 }
+
+type ReportCommand struct {
```
mszostok, line 46: **nit** 
```suggestion
type ReportCommandInput struct {
```
while imported externally we have `reporter.ReportCommand` that has two meaning, and also it won't allow us to add `ReportCommandOutput` later if needed.

cmd/botkube-agent/main.go
```diff
@@ -239,8 +239,10 @@ func run(ctx context.Context) (err error) {
 	//    For example, if in both communication groups there's a Slack configuration pointing to the same workspace,
 	//	  when user executes `kubectl` command, one Bot instance will execute the command and return response,
 	//	  and the second "Sorry, this channel is not authorized to execute kubectl command" error.
+	commGroupIdx := 1
```
mszostok, line 242: Maybe it would be good to document that, similar as we have here: https://github.com/kubeshop/botkube-cloud/blob/main/docs/global-infrastructure.md#segment-analytics

For me, this index is a bit misleading, so it would be good to have that documented.


```json
[
	{
		"type": "track",
		"messageId": "0",
		"anonymousId": "cluster-id",
		"event": "Integration enabled",
		"timestamp": "2009-11-17T20:34:58.651387237Z",
		"properties": {
			"index": 1,
			"platform": "slack",
			"type": "bot"
		}
	},
	{
		"type": "track",
		"messageId": "1",
		"anonymousId": "cluster-id",
		"event": "Integration enabled",
		"timestamp": "2009-11-17T20:34:58.651387237Z",
		"properties": {
			"index": 2,
			"platform": "discord",
			"type": "bot"
		}
	},
	{
		"type": "track",
		"messageId": "2",
		"anonymousId": "cluster-id",
		"event": "Integration enabled",
		"timestamp": "2009-11-17T20:34:58.651387237Z",
		"properties": {
			"index": 1,
			"platform": "teams",
			"type": "bot"
		}
	}
]
```

Here it looks like the Discord is somehow the second instance. While it's just 1 instance but in the second group, so maybe `index` -> `groundId`?

mszostok (approved):

internal/analytics/reporter.go
pkosiec: I know, I wanted to do that but I went with consistency to the `ReportEvent` , as then I would rename it too... but ok, I will rename it 😄

internal/analytics/reporter.go
```diff
@@ -14,20 +14,20 @@ type Reporter interface {
 	// RegisterCurrentIdentity loads the current anonymous identity and registers it.
 	RegisterCurrentIdentity(ctx context.Context, k8sCli kubernetes.Interface, deployID string) error
 
-	// ReportCommand reports a new executed command. The command should be anonymized before using this method.
-	ReportCommand(platform config.CommPlatformIntegration, command string, origin command.Origin, withFilter bool) error
+	// ReportCommandInput reports a new executed command. The command should be anonymized before using this method.
```
mszostok, line 17:
```suggestion
	// ReportCommand reports a new executed command. The command should be anonymized before using this method.
```

pkg/bot/slack_cloud.go
```diff
@@ -211,6 +212,12 @@ func (b *CloudSlack) start(ctx context.Context) error {
 	b.setFailureReason("")
 	go b.startMessageProcessor(ctx, messageWorkers, messages)
 
+	if err := b.reporter.ReportBotEnabled(b.IntegrationName(), b.commGroupMetadata.Index); err != nil {
+		b.setFailureReason(health.FailureReasonConnectionError)
```
mszostok, line 216: It's analytics, I would skip setting failure reason here. We can fix that also in socket slack code.

pkg/bot/slack_cloud.go
```diff
@@ -211,6 +212,12 @@ func (b *CloudSlack) start(ctx context.Context) error {
 	b.setFailureReason("")
 	go b.startMessageProcessor(ctx, messageWorkers, messages)
 
+	if err := b.reporter.ReportBotEnabled(b.IntegrationName(), b.commGroupMetadata.Index); err != nil {
```
mszostok, line 215: As you know, retries can occur quite often. Would it be a problem to report multiple bot-enabled events? Should we use `sync.Once` here? This way, it will trigger only during each Botkube restart, which happens a bit less often.

mszostok (approved):

pkg/bot/slack_cloud.go
pkosiec: I essentially copied the same code we have across all our bots. Should I adjust the code everywhere? 😀

pkg/bot/slack_cloud.go
pkosiec: No, that's not a problem as we already report such events multiple times for the instances that restart. So in the telemetry dashboard we already can't rely on number of such events per instance - instead, we should count such distinct events for platforms.

Anyway, if I'll be already modifying the PR, I can add sync.Once, that's not a problem 👍 it'll reduce the emitted events by tiny bit, so it's still a small improvement.

PR #1300, opened 2023-10-26
Update thread-mate schema, fix setting emoji indicator

mszostok:
## Description

- Update thread-mate schema, fix setting emoji indicator

### DIFF
--- a/internal/executor/thread-mate/config.go
+++ b/internal/executor/thread-mate/config.go
@@ -56,6 +56,11 @@ func (c *Config) Validate() error {
 // MergeConfigs merges the configuration.
 func MergeConfigs(configs []*executor.Config) (Config, error) {
 	defaults := Config{
+		Logger: config.Logger{
+			Level:         "info",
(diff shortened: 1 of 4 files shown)


internal/executor/thread-mate/config.go
```diff
@@ -56,6 +56,11 @@ func (c *Config) Validate() error {
 // MergeConfigs merges the configuration.
 func MergeConfigs(configs []*executor.Config) (Config, error) {
 	defaults := Config{
+		Logger: config.Logger{
+			Level:         "info",
```
josefkarasek, line 60: Doesn't look configurable, but I'll leave it up to you if it needs to be configurable or not.

josefkarasek (approved):
lgtm

internal/executor/thread-mate/config.go
mszostok: It's just a default, other values can be specified in the YAML config :)

PR #1310, opened 2023-11-16
Reduce amount of anonymous analytics events

pkosiec:
## Description

- Reduce amount of anonymous analytics events

## Testing
The easiest way is to reduce the `tickDuration` in code (`Segment Reporter`):
```diff
 // NewSegmentReporter creates a new SegmentReporter instance.
 func NewSegmentReporter(log logrus.FieldLogger, cli segment.Client) *SegmentReporter {
 	return &SegmentReporter{
 		log:          log,
 		cli:          cli,
 		batchedData:  batched.NewData(defaultTimeWindowInHours),
-		tickDuration: defaultTimeWindowInHours * time.Hour,
+		tickDuration: 10 * time.Seconds,
 	}
 }

```

In the same file, you can set fixed APIKey:
```diff
 var (
 	// APIKey contains the API key for external analytics service. It is set during application build.
-	APIKey string
+	APIKey string = "abc..."
 )
```

Run Botkube with debug log (`export BOTKUBE_SETTINGS_LOG_LEVEL=debug`) and observe Segment events in debugger. Emit some sample events.

## Screenshots
image
image


internal/analytics/segment_reporter.go
```diff
@@ -158,12 +181,61 @@ func (r *SegmentReporter) ReportFatalError(err error) error {
...
+			if err != nil {
+				return fmt.Errorf("while reporting heartbeat event: %w", err)
+			}
+
+			return nil
+		case <-ticker.C:
+			err := r.reportHeartbeatEvent()
+			if err != nil {
+				wrappedErr := fmt.Errorf("while reporting heartbeat event: %w", err)
+				r.log.Error(wrappedErr.Error())
```
mszostok, line 204: Personally, I like that approach more when it comes to logging, but it's up to you.
```suggestion
				r.log.WithError(err).Error("Failed to report heartbeat event")
```

mszostok (approved):
I didn't test that with real segment, but code LGTM 👍 

But please double-check with Ole that Segment.io doesn't have any message size limit.

Because we may switch the events quota problems to size quota problem 😄 

If so, we can keep details maybe about N last source events? while counters will still tell about the exact number. This will prevent us from failing to send an event when e.g. agent cluster is misconfigured and someone watches e.g. for all events.

pkosiec:
Merging as the integration test issue for Discord wasn't related to the PR and Slack tests passed 👍

PR #1319, opened 2023-11-23
Additional logs for the grpc conn.

huseyinbabal:
## Description

- Added additional logs for grpc connection
- Set timeout for grpc connection methods since somehow retry mechanism somehow stuck on second retry after 502 server response and it keeps ctx open infinitely

### DIFF
--- a/pkg/bot/slack_cloud.go
+++ b/pkg/bot/slack_cloud.go
@@ -38,10 +38,11 @@ import (
 )
 
 const (
-	retryDelay              = time.Second
-	maxRetries              = 30
-	successIntervalDuration = 3 * time.Minute
-	quotaExceededMsg        = "Quota exceeded detected. Stopping reconnecting to Botkube Cloud gRPC API..."
+	retryDelay               = time.Second
+	maxRetries               = 30
+	successIntervalDuration  = 3 * time.Minute
+	quotaExceededMsg         = "Quota exceeded detected. Stopping reconnecting to Botkube Cloud gRPC API..."
+	cloudSlackConnectTimeout = 10 * time.Second
 )
 
 var _ Bot = &CloudSlack{}
@@ -178,7 +179,7 @@ func (b *CloudSlack) start(ctx context.Context) error {
 
 	conn, err := grpc.Dial(b.cfg.Server.URL, opts...)
 	if err != nil {
-		return err
+		return fmt.Errorf("while creating gRPC connection: %w", err)
 	}
 	defer conn.Close()
 
@@ -186,7 +187,9 @@ func (b *CloudSlack) start(ctx context.Context) error {
 		InstanceId: remoteConfig.Identifier,
 		BotId:      b.botID,
 	}
-	c, err := pb.NewCloudSlackClient(conn).Connect(ctx)
+	ctxTimeout, cancelFn := context.WithTimeout(ctx, cloudSlackConnectTimeout)
+	defer cancelFn()
+	c, err := pb.NewCloudSlackClient(conn).Connect(ctxTimeout)
 	if err != nil {
 		return fmt.Errorf("while initializing gRPC cloud client: %w", err)
 	}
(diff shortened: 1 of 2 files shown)


pkosiec (commented):
Not sure if the timeout isn't too small, but let's. test the PR image on our Botkube environments to see if that helps for a longer period of time 👍 

Also, consider also adding https://grpc.io/docs/guides/deadlines/ to the server-side too if that makes sense 👍

huseyinbabal:
> Not sure if the timeout isn't too small, but let's. test the PR image on our Botkube environments to see if that helps for a longer period of time 👍
> 
> Also, consider also adding https://grpc.io/docs/guides/deadlines/ to the server-side too if that makes sense 👍

huseyinbabal:
As a summary;
- I tested the initial version in dev but timeout just didn't fix it
- The problem caused by not-reseted exponential back-off that causes a stuck even we restart slack processor
- I have added exponential back-off reset mechanism by using failure counter. 
- Now, whenever agent connects router, it resets failureNo and backoff strategy is based on failureNo, which means we have faster retries on processor restarts in agent side.

pkg/bot/slack_cloud.go
```diff
@@ -126,34 +126,32 @@ func (b *CloudSlack) Start(ctx context.Context) error {
 func (b *CloudSlack) withRetries(ctx context.Context, log logrus.FieldLogger, maxRetries int, fn func() error) error {
 	b.failuresNo = 0
 	var lastFailureTimestamp time.Time
+	resettableBackoff := func(n uint, err error, cfg *retry.Config) time.Duration {
+		if !lastFailureTimestamp.IsZero() && time.Since(lastFailureTimestamp) >= successIntervalDuration {
+			// if the last run was long enough, we treat is as success, so we reset failures
+			log.Infof("Resetting failures counter as last failure was more than %s ago", successIntervalDuration)
+			b.failuresNo = 0
```
pkosiec, line 133: we can also restet the failure msg here

pkg/bot/teams_cloud.go
```diff
@@ -211,34 +218,32 @@ func (b *CloudTeams) start(ctx context.Context) error {
...
+			return err
 		},
 		retry.OnRetry(func(_ uint, err error) {
 			log.Warnf("Retrying Cloud Teams startup (attempt no %d/%d): %s", b.failuresNo, maxRetries, err)
 		}),
-		retry.Delay(retryDelay),
+		retry.DelayType(resettableBackoff),
 		retry.Attempts(0), // infinite, we cancel that by our own
 		retry.LastErrorOnly(true),
 		retry.Context(ctx),
```
pkosiec, line 249: should we also add something similar here to

```go
	b.failuresNo = 0 // Reset the failures to start exponential back-off from the beginning
	b.setFailureReason("")
	b.log.Info("Botkube connected to Slack!")
```

?

pkg/bot/slack_cloud.go
pkosiec, line 1: while it wasnt the actual reason, imho it would be good to add the same what we have in cloud teams:

```go
ctxTimeout, cancelFn := context.WithTimeout(ctx, cloudTeamsConnectTimeout)
	defer cancelFn()
	err = svc.Start(ctxTimeout)
	if err != nil {
		return fmt.Errorf("while starting gRPC connector %w", err)
	}
```

pkosiec (commented):
Looks very promising!

pkg/bot/slack_cloud.go
huseyinbabal: :ok

pkg/bot/teams_cloud.go
huseyinbabal: I couldn't get the point here, could you please elaborate a bit more?

pkg/bot/slack_cloud.go
huseyinbabal: Actually, I should remove this from teams too, I was testing slack first. So the problem with ctxTimeout, it cancels the existing streaming connection event after the start is initiated. Assume, I provided 15 secs timeout, connection is created successfully, but during the streaming it just cancelled since it is not only used in connection, it is also used in grpc operation.

pkosiec (approved):
LGTM based on the code review and testing on dev 👍

I think we can merge this PR, run Botkube from latest main image for all our envs and observe behavior 👍

PR #1330, opened 2023-12-11
Add origin extraction for cmd execution, add user display name for interactive command

mszostok:
## Description

- Add user display name for interactive command responses.
- Introduce a feature that allows explicit origin extraction for the execution of commands. Furthermore, for commands that do not have an explicit origin property, a fallback strategy has been implemented to determine the command origin based on activity type.

## Related issue(s)

Fix https://github.com/kubeshop/botkube-cloud/issues/739

### DIFF
--- a/pkg/bot/teams_cloud.go
+++ b/pkg/bot/teams_cloud.go
@@ -312,7 +312,7 @@ func (b *CloudTeams) processMessage(ctx context.Context, act schema.Activity, ch
 		Message: trimmedMsg,
 		User: execute.UserInput{
 			// Note: this is a plain text mention, a native mentions will be provided as a part of:
-			// https://github.com/kubeshop/botkube-cloud/issues/738
+			// https://github.com/kubeshop/botkube/issues/1331
 			Mention:     act.From.Name,
 			DisplayName: act.From.Name,
 		},
@@ -323,7 +323,7 @@ func (b *CloudTeams) processMessage(ctx context.Context, act schema.Activity, ch
 
 // generate a function code tab based on in type will return proper command origin.
 func (b *CloudTeams) mapToCommandOrigin(act schema.Activity) command.Origin {
-	// in the newer control-plane version, the origin is explicitly set
+	// in the newer Botkube Cloud version, the origin is explicitly set
 	c, found := b.extractExplicitOrigin(act)
 	if found {
 		return c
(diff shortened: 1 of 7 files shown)


pkg/bot/teams_cloud.go
```diff
@@ -302,20 +306,72 @@ func (b *CloudTeams) processMessage(ctx context.Context, act schema.Activity, ch
...
 			DisplayName: act.From.Name,
 		},
 		AuditContext: act.ChannelData,
 	})
 	return e.Execute(ctx)
 }
 
+// generate a function code tab based on in type will return proper command origin.
+func (b *CloudTeams) mapToCommandOrigin(act schema.Activity) command.Origin {
+	// in the newer control-plane version, the origin is explicitly set
```
pkosiec, line 326: Maybe this? 🤔 
```suggestion
	// in the newer Botkube Cloud version, the origin is explicitly set
```

pkg/bot/teams_cloud.go
```diff
@@ -302,20 +306,72 @@ func (b *CloudTeams) processMessage(ctx context.Context, act schema.Activity, ch
...
+	}
+
+	origin := data.OriginName
+	if origin == "" {
+		origin = data.Action.Data.OriginName
+	}
+	if command.IsValidOrigin(origin) {
+		return command.Origin(origin), true
+	}
+	return "", false
```
pkosiec, line 372: Do you think this check is needed? 🤔

pkosiec (approved):
LGTM 👍 works well

pkg/bot/teams_cloud.go
```diff
@@ -302,20 +306,72 @@ func (b *CloudTeams) processMessage(ctx context.Context, act schema.Activity, ch
...
-			CommandOrigin:    command.TypedOrigin,
+			CommandOrigin:    b.mapToCommandOrigin(act),
 			DisplayName:      channelDisplayName,
 		},
 		Message: trimmedMsg,
 		User: execute.UserInput{
-			// TODO: we need to add support for mentions on cloud side.
-			//Mention:     "",
+			// Note: this is a plain text mention, a native mentions will be provided as a part of:
+			// https://github.com/kubeshop/botkube-cloud/issues/738
```
pkosiec, line 315: Should we make this issue public? 🤔

pkg/bot/teams_cloud.go
mszostok: this is used in combination with fallback strategy, so I think it would be better to do this check here, so fallback will work properly.

PR #1332, opened 2023-12-14
Support TLS encryption for gRPC connections in Cloud Slack and Teams

pkosiec:
## Description

- Support TLS encryption for gRPC connections in Cloud Slack and Teams

## Testing

Test it E2E with the changes on Cloud.

### DIFF
--- a/pkg/grpcx/credentials.go
+++ b/pkg/grpcx/credentials.go
@@ -14,8 +14,8 @@ import (
 
 // ClientTransportCredentials returns gRPC client transport credentials based on the provided configuration.
 func ClientTransportCredentials(log logrus.FieldLogger, cfg config.GRPCServer) (credentials.TransportCredentials, error) {
-	if cfg.DisableSecurity {
-		log.Warn("Server CA certificate is not provided. Using insecure gRPC connection...")
+	if cfg.DisableTransportSecurity {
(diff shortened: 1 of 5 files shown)


pkg/grpcx/credentials.go
```diff
@@ -0,0 +1,49 @@
...
+	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/credentials/insecure"
+
+	"github.com/kubeshop/botkube/pkg/config"
+)
+
+// ClientTransportCredentials returns gRPC client transport credentials based on the provided configuration.
+func ClientTransportCredentials(log logrus.FieldLogger, cfg config.GRPCServer) (credentials.TransportCredentials, error) {
+	if cfg.DisableSecurity {
+		log.Warn("Server CA certificate is not provided. Using insecure gRPC connection...")
```
mszostok, line 18: It could be misleading; the cert can be provided, but simply flag can disable it. Please update the msg 👍 Also, 'insecure' may be understood as not only encryption is disabled but also auth, so I would update that as well, thanks!

mszostok (approved):
LGTM 🚀

pkosiec:
Merging as integration tests will fail for Slack until we have new Botkube Cloud on Dev 👍

PR #1341, opened 2023-12-22
Forward action output to sinks

mszostok:
## Description

- Forward action output to sinks

## Testing
You can try this tutorial: https://docs.szostok.io/botkube-plugins/sources/incoming-http

### DIFF
--- a/internal/source/dispatcher.go
+++ b/internal/source/dispatcher.go
@@ -259,7 +259,16 @@ func (d *Dispatcher) dispatchMsg(ctx context.Context, event source.Event, dispat
 				defer analytics.ReportPanicIfOccurs(d.log, d.reporter)
 				err := n.SendMessage(ctx, genericMsg, sources)
 				if err != nil {
-					d.log.Errorf("while sending action result message: %s", err.Error())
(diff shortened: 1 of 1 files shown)


internal/source/dispatcher.go
pkosiec, line 1: Currently we don't display actions as a part of sinks on Cloud UI. Please create a separate issue for that, as I don't think it will be trivial to address right now (Why? Actions require also executors, which are hidden as part of sinks; maybe it could be introduced as a part of actions UI overhaul)

pkosiec, line 1: Could you add json struct  tags for the `api.Message`? Right now I can see props `Header`, `Description` etc., and it would be great to have them properly capitalized.

image

BTW AFAIK K8s source event also lacks the proper capitalization for the struct fields.

pkosiec (approved):
Works well 👍

internal/source/dispatcher.go
mszostok: discussed f2f, this will introduce a breaking change between agents and cloud, meaning that we will need to have a conditional unmarshaling based on agent version. For now, we skip fixing it.

PR #1355, opened 2024-01-19
Migrate disabled plugins to Cloud

madebyrogal:
## Description

- migrate disabled plugin
- fix missing displayName during migration
- fix RBAC conversion
- fix installation command abort update existing instalation

## Testing

Test this against Botkube Cloud dev setup (UI and BE).

Check out the PR and build the latest CLI:

```
export IMAGE_TAG=v9.99.9-dev
make release-snapshot-cli
chmod +x ./dist/botkube-cli_darwin_amd64_v1/botkube && cp ./dist/botkube-cli_darwin_amd64_v1/botkube /usr/local/bin/botkube_dev
```

Then, install Botkube in your k3d cluster

```
export ALLOW_KUBECTL=true
export ALLOW_HELM=true
export SLACK_CHANNEL_NAME=botkube-demo \
export SLACK_API_APP_TOKEN=xapp-... \
export SLACK_API_BOT_TOKEN=xoxb-... \
export CLUSTER_NAME=dev
botkube_dev install --version 1.7.0 \
--set communications.default-group.socketSlack.enabled=true \
--set communications.default-group.socketSlack.channels.default.name=${SLACK_CHANNEL_NAME} \
--set communications.default-group.socketSlack.appToken=${SLACK_API_APP_TOKEN} \
--set communications.default-group.socketSlack.botToken=${SLACK_API_BOT_TOKEN} \
--set settings.clusterName=${CLUSTER_NAME} \
--set 'executors.k8s-default-tools.botkube/kubectl.enabled'=${ALLOW_KUBECTL} \
--set 'executors.k8s-default-tools.botkube/helm.enabled'=${ALLOW_HELM}
```
Then you will be asked about updating existing instance (test both options)
- option 1: update existing instance
- option 2: don't update

Each time check if you can see disabled plugins in UI

### DIFF
--- a/internal/cli/install/helm/install.go
+++ b/internal/cli/install/helm/install.go
@@ -74,8 +74,9 @@ func (c *Helm) Install(ctx context.Context, status *printer.StatusPrinter, opts 
 			if err != nil {
 				return nil, fmt.Errorf("while confiriming upgrade: %v", err)
 			}
-
 			if !upgrade {
+				status.Step("Skipped Botkube installation upgrade.")
+				status.End(true)
 				return nil, nil
 			}
 		}
--- a/internal/cli/install/helm/install.go
+++ b/internal/cli/install/helm/install.go
@@ -2,7 +2,6 @@ package helm
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"os"
 	"strings"
@@ -77,7 +76,7 @@ func (c *Helm) Install(ctx context.Context, status *printer.StatusPrinter, opts 
 			}
 
 			if !upgrade {
-				return nil, errors.New("upgrade aborted")
+				return nil, nil
 			}
 		}
 		restartAnnotation := fmt.Sprintf(restartAnnotationFmt, time.Now().Unix())
--- a/internal/cli/install/install.go
+++ b/internal/cli/install/install.go
@@ -97,6 +97,10 @@ func Install(ctx context.Context, w io.Writer, k8sCfg *kubex.ConfigWithMeta, opt
 	if err != nil {
 		return err
 	}
+	if rel == nil {
+		//User answered no on prompt
+		return nil
+	}
 
 	if opts.HelmParams.DryRun {
 		return printSuccessInstallMessage(opts.HelmParams.Version, w)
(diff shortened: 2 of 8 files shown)


pkosiec:
@madebyrogal Please update Migration E2E integration tests to cover the new functionality 🙂 Thanks!

internal/cli/install/helm/install.go
```diff
@@ -77,7 +76,7 @@ func (c *Helm) Install(ctx context.Context, status *printer.StatusPrinter, opts
 			}
 
 			if !upgrade {
-				return nil, errors.New("upgrade aborted")
+				return nil, nil
```
pkosiec, line 79: It'd be good to still print it:
```suggestion
				status.Step("Skipped Botkube installation upgrade.")
				status.End(true)
				return nil, nil
```

internal/cli/install/install.go
```diff
@@ -97,6 +97,10 @@ func Install(ctx context.Context, w io.Writer, k8sCfg *kubex.ConfigWithMeta, opt
 	if err != nil {
 		return err
 	}
+	if rel == nil {
+		//User answered no on prompt
```
pkosiec, line 101: Could you please elaborate here? On which prompt? Please update the comment to clarify which case it covers. Thanks!

pkg/config/config.go
```diff
@@ -347,7 +347,8 @@ const (
 
 // Executors contains executors configuration parameters.
 type Executors struct {
-	Plugins Plugins `yaml:",inline" koanf:",remain"`
+	DisplayName string  `yaml:"displayName"`
```
pkosiec, line 350: Can you add displayName to the `values.yaml` file for the Botkube Helm chart, so that we have the display names by default for all executor config groups? Thanks!

pkosiec (commented):
Didn't test it, just code review. Please update Migration E2E integration tests to cover the new functionality and after that (plus fixing the minor comments) I'll test the PR E2E 🙂 Thanks!

Hint: you can open a new PR based on this one using branch on the upstream repo, to be able to trigger the migration tests there.

pkg/config/config.go
madebyrogal: I've added to specific executor, cos sometimes in the group there is more the one plugin. Please verify if it's ok.

pkosiec (approved):
LGTM, good work!

PR #1381, opened 2024-02-12
Support custom headers during plugin index request

pkosiec:
## Description

- Support custom headers during plugin index request
- Fix returning errors for CLI migration

## Testing

Test it E2E using the other PR with custom plugin index endpoint.


pkg/config/config.go
```diff
@@ -166,9 +166,10 @@ func (p PluginRestartPolicyType) ToLower() string {
 	return strings.ToLower(string(p))
 }
 
-// PluginsRepositories holds the Plugin repository information.
-type PluginsRepositories struct {
-	URL string `yaml:"url"`
+// PluginsRepository holds the Plugin repository information.
+type PluginsRepository struct {
+	URL     string `yaml:"url"`
+	Headers map[string]string
```
madebyrogal, line 172: When checking `helm/botkube/values.yaml` I don't see headers in plugins.repositories
But I'm not sure if this is an issue anyway.
Maybe some comment in values.yaml will be a good option.

madebyrogal (approved):
Just one comment / question
Beside that. Well done!

PR #1382, opened 2024-02-12
Stop building multi arch binaries for cli snapshot

mszostok:
## Description

- Stop building multi arch binaries for CLI snapshot. For some reason, we built multi arch even though it's not needed and just make the e2e test longer.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -64,8 +64,11 @@ container-image-single: pre-build
 release-snapshot:
 	@./hack/goreleaser.sh release_snapshot
 
+build-single-arch-cli:
+	@./hack/goreleaser.sh build_single_arch_cli
+
 release-snapshot-cli:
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -64,8 +64,11 @@ container-image-single: pre-build
 release-snapshot:
 	@./hack/goreleaser.sh release_snapshot
 
+build-single-arch-cli:
+	@./hack/goreleaser.sh build_single_arch_cli
+
 release-snapshot-cli:
```
mszostok, line 70: It needs to stay because of the `pull_request_target` that always uses the job manifest from `main`. I will remove it in my next PR after merging this one.

vaijab (approved):
Makes sense. Great stuff.

PR #1387, opened 2024-02-19
Add teams and slack parent activity id support, remove not used activityId field from gRPC

mszostok:
## Description

-  Add teams and slack parent activity id support, remove not used `activityId` field from gRPC

## Related issue(s)

fix https://github.com/kubeshop/botkube-cloud/issues/890


proto/cloud_teams.proto
```diff
@@ -16,7 +16,6 @@ enum MessageType {
 
 message Message {
   string teamId = 1;
-  string activityId = 2;
```
pkosiec, line 19: It's ok to remove the field as we updated the plugin contract version recently. Is that correct?

pkosiec (approved):
LGTM (just code review)

proto/cloud_teams.proto
mszostok: it's about cloud_teams, so it is not related, it's only about control-plane <-> agent communication. They should work properly as this field was not used anyway 👍

PR #1391, opened 2024-02-22
Remove unused Lifecycle server properties

pkosiec:
## Description

- Remove unused Lifecycle server properties

Just spotted some old properties which aren't used for a while.

## Testing

No need to test it.


pkosiec:
See https://github.com/kubeshop/botkube/pull/1392

Issue #1376, opened 2024-02-09
Helm upgrade with `existingCommunicationsSecretName` fails


pkosiec:
_Originally posted by @voriol in https://github.com/kubeshop/botkube/issues/1274#issuecomment-1933520800_
            
---

When updating my helm chart to the latest version, 1.8.0 I got the following error:

_**Error: template: sre-botkube/charts/botkube/templates/persistent-config.yaml:6:23: executing "sre-botkube/charts/botkube/templates/persistent-config.yaml" at <index $secret.data "comm_config.yaml">: error calling index: index of untyped nil**_

Reviewing the latest changes I see that the line it refers to was changed in this commit

https://github.com/kubeshop/botkube/commit/fdb389c12600e194e503aaa1eb4e2a914e93be7b#diff-e694b58be4de3f2b3d147e65317458bbbc0f5319bd07c703d80dc3b05bb5d1b7

Specifically, on line 6:

`{{- $data := b64dec (index $secret.data "comm_config.yaml") -}}`

This is the secret content:

```
apiVersion: v1
kind: Secret
metadata:
  name: botkube-communications
stringData:
  comm_config.yaml: |
    communications:
      default-group:
        ## Settings for Slack with Socket Mode.
        ...etc
```

I am doing something wrong? 

Thanks

pkosiec:
Hey @voriol,

There might be a bug in the Helm chart when using existing secret. Unfortunately we're focused on different priorities and I think we won't be able to fix this issue in next few weeks. Here are the possible solutions/workarounds I can see:

1. You might try migrating your instance to Botkube Cloud, which doesn't have this issue as it doesn't use the complex ConfigMap merging logic. It's free for a single cluster! 
1. Alternatively, you could switch to use built-in secret with platform details
1. Or, if you have a bit of time and energy, you could contribute and investigate/fix the bug. That would be much appreciated 🙏  Here are a few tips:

You can clone the repo and try to modify the Helm chart in this place:
https://github.com/kubeshop/botkube/blob/main/helm/botkube/templates/persistent-config.yaml#L6

I didn't try it yet, but this could work:
```diff
 {{- if .Values.existingCommunicationsSecretName }}
   {{- $secret := lookup "v1" "Secret" .Release.Namespace .Values.existingCommunicationsSecretName | default dict  }}
-  {{- $data := b64dec (index $secret.data "comm_config.yaml") -}}
+  {{- $secretData := $secret.data | default dict -}}
[...]

PR #1274, opened 2023-09-22
Fix configuration merging when using `existingCommunicationsSecretName`

pkosiec:
## Description

Changes proposed in this pull request:

- Fix configuration merging when using `existingCommunicationsSecretName`
- Do not create unused ConfigMaps when using remote config

## Testing

See the instruction in #1273 and test the values:
```yaml
existingCommunicationsSecretName: "comm-config.yaml"
settings:
  clusterName: "devops.azure.akelius.com"
sources:
  k8s-custom:
    botkube/kubernetes:
      config:
        event:
          types:
            - error
        namespaces:
          include:
            - kube-infrastructure
        resources:
          - type: v1/pods
      context:
        rbac:
          group:
            prefix: ""
            static:
              values:
                - botkube-plugins-default
            type: Static
      enabled: true
    displayName: Kubernetes Errors Kube Infra

resources:
  limits:
    cpu: 100m
    memory: 512Mi
  requests:
    cpu: 10m
    memory: 136Mi
```

Also, test the following scenarios:
- Remote config

- Using default Communications secret - you can simply use
```yaml
communications:
  'default-group':
    socketSlack:
      enabled: true
      channels:
        'default':
          name: 'k8s_devops_azure'
          bindings:
            sources:
              - k8s-custom
      botToken: 'xoxb-'
      appToken: 'xapp-'
settings:
  clusterName: "devops.azure.akelius.com"
sources:
  k8s-custom:
    botkube/kubernetes:
      config:
        event:
          types:
            - error
        namespaces:
          include:
            - kube-infrastructure
        resources:
          - type: v1/pods
      context:
        rbac:
          group:
            prefix: ""
            static:
              values:
                - botkube-plugins-default
            type: Static
      enabled: true
    displayName: Kubernetes Errors Kube Infra

resources:
  limits:
    cpu: 100m
    memory: 512Mi
  requests:
    cpu: 10m
    memory: 136Mi
```

## Related issue(s)
 
Resolves #1273

PR #1408, opened 2024-03-06
Fix Helm chart upgrade when `lookup` is not supported

pkosiec:
## Description

- Fix Helm chart upgrade when `lookup` is not supported

## Testing

I don't think it's needed to test it but if you want to, see the instruction from https://github.com/kubeshop/botkube/pull/1274

## Related issue(s)

See conversation: https://botkube.slack.com/archives/C01CR1KS55K/p1709740107856559
See also: https://github.com/kubeshop/botkube/issues/1376


pauloconnor:
Tested this on our test cluster, and ArgoCD was able to successfully deploy it

PR #1409, opened 2024-03-08
Enhance the help message

mszostok:
## Description

- Fix setting "Active Instance .." header to be visible only on start
- Address helm message enhancements 

## Notes
I have such feedback regarding initial design:

1. Cloud section:
   - add buttons instead of static text
2. AI sections:
   - "Ask a question" - runs "@Botkube ai hi!"
3. Basic commands - all good ✅
4. Notifications:
   - there is no "footer" with btns for all listed options - it's not consistent e.g. with "Basic commands" section
   - Change notification on Botkube Cloud -> Change notification on Cloud - too long to fit into button
     - it will open Instance View as we don't have any other view to link to
5. Run kubectl commands:
    - there is no "footer" with btns for all listed options
6. Multi-cluster flags:
    - the multi-cluster flags are disconnected from the context of setting default cluster per channel - w have that in the "Botkube cloud" section
7. Advanced features section:
    - I don't understand the "[Executors and aliases] - run commands like helm, flux, etc. directly from Slack" what it actually means, where we should link to?
    - I don't think that we should scare our users and call it advanced features. It's not advanced tbh, we just still lack a good UX there, but the overall business functionality is straight forward.

Other:   
- what about helm(?) should we skip it?
- what if we don't have any instance connected and someone runs `botkube help`?
  - in this case we return no instance connected 
     Screenshot 2024-03-08 at 14 07 15

@miklosp  PTAL, below you will find the original message and suggested with small adjustements

#### Slack
<table>
<tr>
<td> Original </td> <td> Suggested </td>
</tr>
<tr>
<td>

Screenshot 2024-03-08 at 11 39 40

</td>
<td>
    
Screenshot 2024-03-08 at 13 03 32

</td>
</tr>
</table>

#### Teams

<table>
<tr>
<td> Original </td> <td> Suggested </td>
</tr>
<tr>
<td>

Screenshot 2024-03-08 at 11 58 47

</td>
<td>
    
Screenshot 2024-03-08 at 13 05 50

</td>
</tr>
</table>

## Testing
Run Botkube with all platforms and ask for help: `@Botkube help`

## Related issue(s)

Resolves https://github.com/kubeshop/botkube-cloud/issues/916


miklosp:
@mszostok LGTM!

PR #1419, opened 2024-03-19
Fix MS Teams integration tests after renderer update

pkosiec:
## Description

- Fix MS Teams integration tests after renderer update


mszostok:
Fix will be handled in https://github.com/kubeshop/botkube/pull/1429

PR #1433, opened 2024-04-18
Add single quote escaping while parsing input command

mszostok:
## Description

- Add single quote escaping while parsing input command. It's more like a workaround.  Ultimately, we should find a better and more generic way for extracting parameters. Additionally, we should delegate command tokenizing to the plugin. However, this would require a bigger refactor, so I deduced to not go into that direction right now.

## Related issue(s)

- https://github.com/kubeshop/botkube-cloud/issues/994


madebyrogal (approved):
Tested in this PR https://github.com/kubeshop/botkube-cloud-plugins/pull/21
Works like a harm!

PR #1441, opened 2024-05-09
fix: close temporary kubeconfig file

testwill:
## Description

- ...


mszostok (approved):
Hi @testwill 

Good catch! Thanks for fixing it! Your fix will be included in the v1.11.0 release in a few days. 🙂

PR #1450, opened 2024-05-27
Introduce italic description for Button Builder API

pkosiec:
## Description

- Introduce italic description for Button Builder API

The new style is used for AI plugin footer formatting.

## Testing

No need to test it.

## Related issue(s)

https://github.com/kubeshop/botkube-cloud/issues/1060


vaijab (approved):
other platforms don't support italic? Or is this for the open source slack integration?

pkosiec:
@vaijab We support advanced formatting like that (description + inline button for Slack and Teams only. 
Teams formatter is closed-source, and we use the italic desc + inline button only for the closed-source AI plugin. 
image

So yeah, the new helper function is used only for AI plugin for now 👍

PR #1455, opened 2024-06-12
Send ack when the new cfg 'resourceVersion' was loaded without restart

mszostok:
## Description

- Send ack when the new cfg 'resourceVersion' was loaded without restart

## Testing

1. Create new Instance
2. Deploy Agent v1.12
3. Update plugin without making any changes
4. See that it stuck in 'Updating' phase
5. Redeploy Agent with the PR image by adding:
    ```sh
    --set settings.log.level="debug
    --set image.repository=kubeshop/pr/botkube
    ```
7. Again, to the plugin update without making any changes
8. See that Agent transition to 'Connected' within 1min

To see the logs you need to have `kubectl set env deployment/botkube BOTKUBE_SETTINGS_LOG_LEVEL="debug"`

https://github.com/kubeshop/botkube/assets/17568639/80e7b6f6-3c30-4cfd-a9b9-e8d5d9df8a60

## Related issue(s)

Fix https://github.com/kubeshop/botkube-cloud/issues/1127

### DIFF
--- a/cmd/botkube-agent/main.go
+++ b/cmd/botkube-agent/main.go
@@ -353,13 +353,14 @@ func run(ctx context.Context) (err error) {
 		cfgReloader, err := reloader.Get(
 			remoteCfgEnabled,
 			logger.WithField(componentLogFieldKey, "Config Reloader"),
+			statusReporter,
 			deployClient,
 			dynamicCli,
 			restarter,
 			analyticsReporter,
 			*conf,
 			cfgVersion,
-			cfgManager,
+			cfgManager, statusReporter,
 		)
 		if err != nil {
 			return reportFatalError("while creating config reloader", err)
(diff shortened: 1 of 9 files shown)


cmd/botkube-agent/main.go
```diff
@@ -353,13 +353,14 @@ func run(ctx context.Context) (err error) {
...
 			logger.WithField(componentLogFieldKey, "Config Reloader"),
+			statusReporter,
 			deployClient,
 			dynamicCli,
 			restarter,
 			analyticsReporter,
 			*conf,
 			cfgVersion,
-			cfgManager,
+			cfgManager, statusReporter,
```
vaijab, line 363: `statusReporter` is passed in twice. Is this meant to be like that?

vaijab (approved):

cmd/botkube-agent/main.go
mszostok: the first one is obtained to report the status by `AckNewResourceVersion`.

While the last parameter is `resVerHolders ...ResourceVersionHolder` so it takes all the instances that share the `resourceVersion` in order to keep them in sync. So it's the same "Instance" but is consumed with different scopes (interfaces)

As the `statusReporter` is already a parameter, I can also handle that directly in the constructor, like:
```go
if remoteCfgEnabled {
		resVerHolders = append(resVerHolders, statusReporter)
```

The first one looks probably more like a Go approach where you define interfaces per usage, instead of having a "fat" one. But honestly, I don't have preferences. I can change it 👍

mszostok:
This introduces instability in the team's tests, so for now, I have put it on hold.

PR #1469, opened 2024-07-04
Hearbeat reporting

madebyrogal:
MOD extend bots and sinks with errorMsg

## Description

- Extend heart beat by adding health status report

## Testing

- This PR needs to be tested with the PR https://github.com/kubeshop/botkube-cloud/pull/1189, where you can find how to test e2e flow

## Related issue(s)

https://github.com/kubeshop/botkube-cloud/issues/142
https://github.com/kubeshop/botkube-cloud/pull/1189


internal/heartbeat/noop_reporter.go
```diff
@@ -8,6 +8,6 @@ var _ HeartbeatReporter = (*NoopHeartbeatReporter)(nil)
 
 type NoopHeartbeatReporter struct{}
 
-func (n NoopHeartbeatReporter) ReportHeartbeat(context.Context, DeploymentHeartbeatInput) error {
+func (n NoopHeartbeatReporter) ReportHeartbeat(context.Context, ReportHeartBeat) error {
```
mszostok, line 11:
```suggestion
func (n NoopHeartbeatReporter) ReportHeartbeat(context.Context, ReportHeartbeat) error {
```

cmd/botkube-agent/main.go
```diff
@@ -61,6 +61,8 @@ const (
 	reportHeartbeatMaxRetries = 30
 )
 
+var healthNotifiers = make(map[string]health.Notifier)
```
mszostok, line 64: 1. it doesn't make sens to have a global var here. Let's more it to the  healtChecker that is responsible for it
2. no need to repeat the interfaces (`HealthNotifierSink`/`HealthNotifierBot`), keep it in a single place (healthChecker)
2. if / else statement is repeated, while it's common for all bots, please see linked code changes
3. FailedSink and FailedBot, referring to the same for the healthChecker - it doesn't care, just have single one
4. You don't need to be worried about integration name, you can use it on 'nil'
    ```go
     func Test(t *testing.T) {
 	    app, err := NewSocketSlackDemo()
 	    assert.Error(t, err)
 	    
 	    fmt.Println(app.IntegrationName())
     }
     
     func NewSocketSlackDemo() (*SocketSlack, error) {
 	    return nil, errors.New("fake")
     }
    ```


Changes in code based on feedback above: https://github.com/kubeshop/botkube/commit/c472d4d285c8208b099bac588bbcad780ec82fdb

mszostok (approved):
Tested e2e, works as expected 👌 

But let's agree on code changes before merging 🙂

PR #1491, opened 2025-07-08
feat: allow botkube to react to messages from other bots/workflows

### LINKED ISSUE #1360: B

albonicc:
## Description
Changed Botkube behavior so it can be triggered by other bots/workflows in Slack, which is a feature that used to work in the past.

A common use case is a Slack workflow that runs on a given schedule. E.g. to invoke botkube for knowing the current scale of nodes/pods every morning at 10:30 AM.

image

## Testing

For testing this functionality, it is needed to create a local k8s cluster (In my case I used K3D), build and run the code by using the steps mentioned in the contribuiting guidelines, and set up a Slack workflow.

image

## Related issue(s)

Resolves #1360


albonicc:
@pkosiec 👀

nthtrung09it (changes requested):
Bug: `b.botID` is set from `authResp.UserID` (a User ID, `U`-prefixed) at `slack_socket.go:86`, while `ev.BotID` from `AppMentionEvent` is a Bot ID (`B`-prefixed). These are different Slack ID namespaces, so the comparison `ev.BotID == b.botID` will **never match**.

This means the bot will never filter its own messages, which could cause infinite loops if bot replies trigger AppMentionEvent.

**Fix:** Use `ev.User == b.botID` instead of `ev.BotID == b.botID` to correctly compare User IDs.

